PR #103, opened 2019-05-08
Block Posts: Allow to select a post type

igmoweb:
Hi.

This might be too much code for a PR coming from an unknown person :). I was trying to find some block plugin that could list posts with different layouts but that could also accept different types of post types. I came across with this plugin and it works great (nice work!) but I'm still missing that Post Type selector that could retrieve posts from another post type so here I am to try to solve this.

I tried also to create my own plugin but then I thought to create a PR against this one so it can be maintained by someone more focused on it.

The PR adds several changes to the Blog Post block:

- It fixes a nasty error that was totally stopping Gutenberg to work when the block is added for the first time in a post. This was due to `latestPosts` `null` value while Gutenberg fetches posts in the background. There are more extra checks now so this case is more unlikely to happen (or impossible).

- It adds a new Post Type selector which displays all the post types registered in WP that are public, have a REST Endpoint and are different from `attachment`.

Captura de pantalla 2019-05-08 a las 16 37 51

- Expands the possibilities of filtering by taxonomies instead of just the category. There's a new `taxQuery` attribute that will handle all taxonomies selected.
Captura de pantalla 2019-05-08 a las 16 38 44

Edit_Post_‹_WordPress_Multisite_—_WordPress

These taxonomies are only those that are hierarchical (So I'm excluding tags), public and have a REST endpoint.

- Instead of a `Display Category` switch, I replaced it for a selector where the user can select what taxonomy to display above every post. Although can be a bit confusing for some users that don't understand what a taxonomy is, I think it made sense to display it this way.

Edit_Post_‹_WordPress_Multisite_—_WordPress

- The plugin is now exposing REST fields for every post type and I replaced `category_list` REST field for a `terms_list` instead, which makes more sense now but I had to rework it as it was returning HTML while I needed to return a list of terms grouped by taxonomy. The HTML is not inserted "Dangerously" now but it could cause some markup differences in some themes. This could be considered to merge this PR I guess.

- The PR takes care of backwards compatibiility: If the user already had the category displaying, this will be still checked.

Thanks!

### DIFF
--- a/src/block/blog-posts/categories-list.js
+++ b/src/block/blog-posts/categories-list.js
@@ -1,0 +1,16 @@
+export default ( props ) => {
+	const { terms = [], separator = ', ' } = props;
+	if ( ! terms.length ) {
+		return null;
+	}
+
+	const parts = terms.map( term => {
+		return <a href={ term.term_link }>{term.name}</a>;
+	} );
+
+	return (
+		<div className="ugb-blog-posts__category-list" >
+			{ parts.reduce(( prev, curr ) => [ prev, separator, curr ] ) }
+		</div>
+	);
+}
--- a/src/block/blog-posts/category-select.js
+++ b/src/block/blog-posts/category-select.js
@@ -1,0 +1,51 @@
+import {
+	TreeSelect,
+} from '@wordpress/components';
+
+import { groupBy } from 'lodash';
+
+/**
+ * Returns terms in a tree form.
+ *
+ * @param {Array} flatTerms  Array of terms in flat format.
+ *
+ * @return {Array} Array of terms in tree format.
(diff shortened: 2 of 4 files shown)


bfintal:
Thanks @igmoweb, this is awesome 💯 

It's quite a large PR though, would you be able to split it into smaller ones? Maybe:

1. latestPosts null fix
2. Post type selector
3. Category/taxonomy dropdown with the terms_list change

I haven't tested your PR, but can you check it with the latest WP 5.2 and the master of Stackable? There were changes in 5.2 that prevented the posts block from working, more specifically on how the categories were queried. Not sure if that would affect your case, but it might be good to check against.

PR #282, opened 2019-10-10
Add support for EditorsKit navigator toolbar

phpbits:
Hi,

I've just recently added Block Navigation toolbar on my EditorsKit plugin. I think this will be of great help on user navigation when both plugins are activated. The toolbar will only be shown on blocks with multiple inner blocks capability.

- [x] Container

Screen Capture on 2019-10-10 at 16-45-11

Let me know your thoughts.

Thanks,
Jeffrey


bfintal:
Awesome! Would you be able to add a PR as well to the [version 2 branch](../tree/feat-three-tabs)?

PR #368, opened 2019-11-19
[block-expand] Limit show less/more link clickable area to text only (backend). Closes #298

### LINKED ISSUE #298: expand show less button clickable area extends to the length of the whole block
Untitled3

jkapunan:


src/block/expand/editor.scss
```diff
@@ -24,3 +24,9 @@
 		font-size: inherit !important;
 	}
 }
+
+.ugb-expand--v2 .ugb-block-content {
```
bfintal: Please use `.ugb-expand`

src/block/expand/editor.scss
```diff
@@ -24,3 +24,9 @@
...
+	a {
```
bfintal: This might affect other links, use `> a`

PR #589, opened 2020-02-04
Google fonts not working in some race hook conditions.

dovy:
Solves an issue where on certain post types, the wp_head hook has already fired. Solve this by hooking to wp_enqueue_scripts which will always run before wp_head, allowing wp_enqueue_style() to always fire properly.


bfintal:
was found to be not an issue

PR #620, opened 2020-02-27
fix (block background): fix inner block content alignment by setting …

### LINKED ISSUE #137: var(--content-width) wont work on IE
**Describe the bug**
On IE, if the Block Background was turned on the list will not be centered even if you choose Align Center. var(--content-width) cant be read on IE.

**To Reproduce**
1. Add a new page
2. Create an Icon List
3. Add a Block Background
4. Publish your page
5. View your page using IE

**Expected behavior**
The alignment should work on IE

**Screenshots**
IE
image

Chrome
image

 - OS: Windows
 - Browser:IE 11

jkapunan:
…a default max-width for ie11. Closes #137

### DIFF
--- a/src/modules/block-background/style.scss
+++ b/src/modules/block-background/style.scss
@@ -11,6 +11,7 @@
 	&.ugb-main-block--inner-center {
 		@media screen and (min-width: $break-large) {
 			.ugb-inner-block {
+				max-width: 580px;
 				max-width: var(--content-width);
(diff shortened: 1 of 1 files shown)


jkapunan:
> What's the reason for using 580px? The usual default content width WP uses is 900px. Do you think it's better to use 580px instead of 900px here?

Changed to 900px.

PR #744, opened 2020-07-08
advanced-toobar for WP 5.5

blankart:
Advanced-toolbar integration for WP5.5


src/components/advanced-toolbar-control/index.js
```diff
@@ -108,18 +110,19 @@ const AdvancedToolbarControl = props => {
...
-						},
-					}
-				} ) }
+				children={
+					controls.map( ( option, index ) => {
+						const controlProps = {
+							...option,
+							onClick: () => props.onChange( option.value !== props.value ? option.value : '' ),
+							isPrimary: props.value === option.value,
+							children: ! option.icon ? option.custom || <span className="ugb-advanced-toolbar-control__text-button">{ option.title }</span> : '',
```
bfintal, line 121: Better to do `null` here since an empty string as a child might still render something

```suggestion
							children: ! option.icon ? option.custom || <span className="ugb-advanced-toolbar-control__text-button">{ option.title }</span> : null,
```

blankart: This is noted. Thank you!

PR #745, opened 2020-07-08
fix($image-upload-placeholder): allow svg uploads

andreiglingeanu:
Check `image.sizes` presence before trying to look if the actual size is present.


bfintal:
Thanks for the PR

Issue #736, opened 2020-07-08
Default spacing, font size, and font style appear to be different


blankart:
Some issues with default padding, font-size, and font-style.

### Accordion Block
All themes have the same issue.
- Title default spacing inconsistent with 5.4
- More font weight in the title
- Larger font
- Different font

Version 5.4
![image](https://user-images.githubusercontent.com/24480990/86877402-6620f300-c119-11ea-89af-91d39665beb2.png)

Version 5.5
![image](https://user-images.githubusercontent.com/24480990/86876946-68368200-c118-11ea-8266-6431cdc5969f.png)

### Blockquote 
All themes have the same issue.
- Different font

Version 5.4
![image](https://user-images.githubusercontent.com/24480990/86879728-2577a880-c11e-11ea-85ae-95992f0c6e85.png)

Version 5.5
![image](https://user-images.githubusercontent.com/24480990/86879794-48a25800-c11e-11ea-98f3-248795b4cd49.png)

### Count up
All themes have the same issue
- Different font

Version 5.4
![image](https://user-images.githubusercontent.com/24480990/86879906-7ab3ba00-c11e-11ea-9e72-6a91c5199d12.png)

Version 5.5
![image](https://user-images.githubusercontent.com/24480990/86879973-9919b580-c11e-11ea-8a04-11ec173b39a8.png)

### Advanced Heading
All themes have the same issue
- DIfferent font

Version 5.4
![image](https://user-images.githubusercontent.com/24480990/86880158-e7c74f80-c11e-11ea-9137-347f0ea65feb.png)

Version 5.5
![image](https://user-images.githubusercontent.com/24480990/86880179-f01f8a80-c11e-11ea-8cd4-efe67c1cda79.png)

### Icon List
All themes have the same issue
- Different font

Version 5.4
![image](https://user-images.githubusercontent.com/24480990/86880327-2bba5480-c11f-11ea-9dc7-ae2b96d31636.png)

Version 5.5
![image](https://user-images.githubusercontent.com/24480990/86880338-2eb54500-c11f-11ea-84d3-3b391d122431.png)

###  Posts
All themes have the same issue
- Different font

Version 5.4
![image](https://user-images.githubusercontent.com/24480990/86881101-4e00a200-c120-11ea-8f3a-3e4eee09eba5.png)

Version 5.5
![image](https://user-images.githubusercontent.com/24480990/86880685-a08d8e80-c11f-11ea-8ac2-9bdb07cc548c.png)

### Button
[...]

PR #749, opened 2020-07-09
Fixed margin-top and margin-bottom in Accordion Preview Mode for WP5.5

blankart:
Fix for margins in Accordion Block. Fixes https://github.com/gambitph/Stackable/issues/736


bfintal:
For testing #775

Issue #741, opened 2020-07-08
h6 Title HTML Tag makes block title capitalized 


blankart:
It also adds a letter spacing. See screenshots below:

Title using h5 HTML tag:
![image](https://user-images.githubusercontent.com/24480990/86887576-59a59600-c12b-11ea-9067-67872ecc8fb3.png)

Title using h6 HTML tag:
![image](https://user-images.githubusercontent.com/24480990/86887584-5c07f000-c12b-11ea-87c5-c96b93550807.png)

PR #750, opened 2020-07-09
Fixed an issue where H6 typography transform is not working as expected

blankart:
Fixes https://github.com/gambitph/Stackable/issues/741


src/components/typography-control/index.js
```diff
@@ -107,7 +107,7 @@ const TypographyControl = props => {
 							{ label: __( 'Uppercase', i18n ), value: 'uppercase' },
 							{ label: __( 'Lowercase', i18n ), value: 'lowercase' },
 							{ label: __( 'Capitalize', i18n ), value: 'capitalize' },
-							{ label: __( 'Normal', i18n ), value: 'normal' },
+							{ label: __( 'Normal', i18n ), value: 'none' },
```
bfintal, line 110: Since there isn't a `normal` value in the first place, let's also change the label to `Normal`
```suggestion
							{ label: __( 'None', i18n ), value: 'none' },
```

PR #753, opened 2020-07-10
Fixed the panel-tabs design for WP 5.5

blankart:


src/components/panel-tabs/editor.scss
```diff
@@ -8,6 +8,14 @@
...
 		box-shadow: none;
 		flex-wrap: nowrap;
+		> button {
+			display: flex !important;
+			transition: all 0.3s ease-in-out;
+			background: #fff;
+			box-shadow: none;
+			border: 0;
+			flex-wrap: nowrap;
+		}
```
bfintal, line 18: 1) This can be cleaned up since `background`, `box-shadow` are repeated.

```css
&,
> button {
	/* put styles here */
}
```

2) I don't think other style rules like `display` don't do anything for this fix

PR #761, opened 2020-07-14
Fixes #757. Fix gap on sides when blocks have block backgrounds turned on

### LINKED ISSUE #757: Align full has gaps on its sides in 5.5
In 5.4:
Screen Shot 2020-07-14 at 12 47 47 PM

In 5.5:
Screen Shot 2020-07-14 at 12 46 18 PM

blankart:
Fixes #757


src/block/icon/editor.scss
```diff
@@ -5,3 +5,8 @@
  */
 
 @import "common";
+
+// Fix inconsistent column-margin in WP5.5
+.ugb-icon__content-wrapper {
+	grid-gap: normal !important;
+}
```
bfintal, line 12: Do we still need this?

src/editor.scss
```diff
@@ -35,16 +35,20 @@ svg.ugb-stackable-icon-gradient {
...
+		h1,
+		h2,
+		h3,
+		h4,
+		h5,
+		h6,
+		p {
+			margin-top: 0;
+			margin-bottom: 0;
+		}
```
bfintal, line 51: I don't think this is the right fix also. I would suggest that you look at how the native blocks are doing it. When you make a native Cover block fullwidth, why doesn't it have any gaps on the sides?

PR #774, opened 2020-07-20
Replaced deprecated IconButton components to Button. Fixes #773

### LINKED ISSUE #773: wp.components.IconButton is deprecated. Please use wp.components.Button instead
There are console warnings about this:

> wp.components.IconButton is deprecated. Please use wp.components.Button instead

IconButton was deprecated WP 5.4, it should be safe to change IconButton components to Button components now

blankart:


blankart:
> @blankart Please do a similar PR in Stackable Premium too

Noted. I already created a pull request in Stackable Premium

PR #780, opened 2020-07-22
Responsive preview mode support for WP 5.5

### LINKED ISSUE #772: Responsive preview for tablet and mobile not working
Toggling the preview to tablet or mobile inconsistent with front end.

image

### LINKED ISSUE #807: Some columns of Stackable Blocks don't get stacked on Mobile Preview Mode
- [ ] The columns of Advanced Columns & Grids don't stack on Mobile Preview Mode

Mobile Preview Mode:
image

Frontend:
image

- [ ] Icon List Block doesn't stack

Mobile Preview Mode:
image

Frontend:
image

blankart:
~Issues need to fix:~
- [ ] ~Some font got bigger on Mobile Preview Mode #806~
- [ ] ~Some columns of Stackable Blocks don't get stacked on Mobile Preview Mode #807~
- [ ] ~Image of Feature Block (Half BG Layout) is longer in Mobile Preview Mode than in Frontend #808~
- [ ] ~Some layouts of Feature is not responsive on Mobile Preview Mode #809~
- [ ] ~Align on Divider Block is not working on Tablet/Mobile Preview Mode #810~

Remaining stuff to do to make this a complete mergeable feature:
- [x] Trigger preview mode change when changing responsive attributes in Stackable controls
- [x] Trigger responsive attribute change in Stackable controls when changing the preview mode (inverse of the one above)
- [x] Split the process into two: 
    - populate the cssObject only when needed,
    - update the mediaText to update the preview
- [x] Changing media query min/max should be reflected
- [x] Developer readme.md describing how preview mode support works
- [x] Fix flickering of style change when modifying styles in tablet/mobile modes
- [x] Write tests

Fixes #772, fixes #807


src/plugins/responsive-preview/index.js
```diff
@@ -0,0 +1,190 @@
...
+
+	// Gets the current width of the visual editor
+	// Cache the width since it's expensive to perform getComputedStyle
+	if ( ! widthsDetected[ mode ] ) {
+		const visualEditorEl = document.querySelector( query )
+		widthsDetected[ mode ] = parseInt( window.getComputedStyle( visualEditorEl ).width, 10 )
+	}
+
+	// Populate the cssObject with media queries and cache values if necessary.
+	cacheCssObject()
```
bfintal, line 164: You may need to throttle this, since goes through a lot of styles for better performance.

src/plugins/responsive-preview/index.js
```diff
@@ -0,0 +1,190 @@
...
+						}
+					} else {
+					// Store the new media query if the value is not found in cached media queries.
+					 mediaIndices[ mediaIndex ] = {
+							mediaText: media.mediaText,
+							min,
+							max,
+							cssText,
+					 }
+					}
```
bfintal, line 101: I'm a bit uncomfortable in comparing a number for the min and max here (even though it's almost unlikely that someone will use 5000px).

bfintal:
The PR is in working condition. I changed a few things to to further separate the util functions:

1. The `cacheCssObject` had a lot of arguments which are used for caching. I moved those inside the function instead since ideally, the function should just be performing the caching internally. The caller of the function doesn't even have to know about it.
2. The `cacheCssObject` had an argument `currentDocument` which was only being used during testing. This is a good move, but I moved this argument to the end and made it have a default value so that you'll only have to use it during tests. When using the function you don't have to put in a `currentDocument`.
3. Renamed the `cacheCssObject` function to `getCssObject` since it really isn't caching the css object, but instead it's getting it.
4. I did the same things for `updateMediaQuery`
5. I noticed that the only purpose of the `cacheCssObject` method is to be immediately passed to `updateMediaQuery`, so moved the call inside. I adjusted the arguments accordingly. This resulted in a more readable function:

```js
updateMediaQuery( [ "css_file.css", "to/change.css" ], mode, width )
```

bfintal:
We need to spin the new TODOs in another PR to prevent this one from getting out of hand. The original intention of this PR is to allow switching, which it already does.

Here's what we need to do:
- [ ] Undo the changes for 3f2d687 from this PR
- [ ] Let's merge this into master when the 3f2d687 has been removed
- [ ] After merging, address issues #806, #807, #808, #809, #810 individually

PR #823, opened 2020-08-12
Icon list improvements

blankart:


src/block/icon-list/edit.js
```diff
@@ -158,34 +171,53 @@ const edit = props => {
...
 
-	return (
-		<BlockContainer.Edit className={ mainClasses } blockProps={ props } render={ () => (
-			<Fragment>
+	const ListItem = range( 1, 7 ).map( index => {
+		const icon = attributes[ `icon${ index }` ]
+		const text = attributes[ `text${ index }` ]
+		return (
+			<li
+				className={ `ugb-icon-list--item${ index }` }
```
bfintal, line 194: Follow [BEM](http://getbem.com/introduction/)

```suggestion
				className={ `ugb-icon-list__item${ index }` }
```

src/block/icon-list/edit.js
```diff
@@ -158,34 +171,53 @@ const edit = props => {
...
+				key={ index }
+			>
+				<SvgIconPlaceholder
+					className={ `ugb-icon-list--icon${ index }` }
```
bfintal, line 198: Follow [BEM](http://getbem.com/introduction/)

```suggestion
					className={ `ugb-icon-list__icon${ index }` }
```

src/block/icon-list/edit.js
```diff
@@ -158,34 +171,53 @@ const edit = props => {
...
+					blockAttributes={ props.attributes }
+					value={ icon }
+					onChange={ value => setAttributes( { [ `icon${ index }` ]: value } ) }
+				/>
 				<RichText
-					tagName="ul"
-					multiline="li"
-					value={ text }
-					onChange={ text => setAttributes( { text } ) }
+					className={ `ugb-icon-list--text${ index }` }
```
bfintal, line 205: Follow [BEM](http://getbem.com/introduction/)

```suggestion
					className={ `ugb-icon-list__text${ index }` }
```

src/block/icon-list/save.js
```diff
@@ -19,29 +20,47 @@ import { Fragment } from '@wordpress/element'
...
 
-	return (
-		<BlockContainer.Save className={ mainClasses } blockProps={ props } render={ () => (
-			<Fragment>
+	const ListItem = range( 1, 7 ).map( index => {
+		const icon = attributes[ `icon${ index }` ]
+		const text = attributes[ `text${ index }` ]
+		return (
+			<li
+				className={ `ugb-icon-list--item${ index }` }
```
bfintal, line 41: Follow [BEM](http://getbem.com/introduction/)

```suggestion
				className={ `ugb-icon-list__item${ index }` }
```

src/block/icon-list/save.js
```diff
@@ -19,29 +20,47 @@ import { Fragment } from '@wordpress/element'
...
+				key={ index }
+			>
+				<SvgIconHelper.Content
+					className={ `ugb-icon-list--icon${ index }` }
```
bfintal, line 45: Follow [BEM](http://getbem.com/introduction/)

```suggestion
					className={ `ugb-icon-list__icon${ index }` }
```

src/block/icon-list/save.js
```diff
@@ -19,29 +20,47 @@ import { Fragment } from '@wordpress/element'
...
+					attrNameTemplate="icon%s"
+					blockAttributes={ props.attributes }
+					value={ icon }
+				/>
 				<RichText.Content
-					tagName="ul"
+					className={ `ugb-icon-list--text${ index }` }
```
bfintal, line 51: Follow [BEM](http://getbem.com/introduction/)

```suggestion
					className={ `ugb-icon-list__text${ index }` }
```

bfintal (changes requested):
A few more issues:

1. If you click on an icon to show the icon popup, then click on the list text to edit, clicking again on the icon won't show the popup (you'll need to click it again)
2. The icon we should use by default should be the icon that's included in FontAwesome (search for "check"). The thin check icon should just be used for migrating old icon list blocks. For users who are creating new icon list blocks, they will never see the old thin check marks again.
3. If you upload a custom SVG it gets re-colored into black if no color is selected. It might be better if colors shouldn't be applied to custom uploaded SVG unless you set a color.

A few requests before we merge this:

1. Suggest moving `updateIconAttribute` inside `with-transform-old-icon-attributes/index.js` since this is only used for that
2. Is `deprecated-style.js` still being used? If it's not then we should remove it
3. Since we now have a lot of deprecated files, maybe it's a good idea to keep them organized to be future-proof. How about the following filestructure:

- icon-list/
  - deprecated.js
  - deprecated/
    - icons.js
    - util.js
    - images/
      - *.svg

bfintal (approved):

PR #903, opened 2020-10-22
Added a fallback array to avoid errors

blankart:


bfintal (changes requested):
Can the `convertBetaStackableColorsToRelease` function be moved out of `domReady` to make things cleaner?

PR #907, opened 2020-11-03
Layout design block adder

### LINKED ISSUE #931: Title not showing in Advanced text
**Describe the bug**
Title is not showing when Side Title 1 and Side Title 2 is chosen from the layout [Advanced text block]

image

### LINKED ISSUE #932: No layout and preset designs setting in Video Popup block
**Describe the bug**
Layout and preset designs setting is missing in Video Popup block

VIDEO POP UP NO LAYOUT

### LINKED ISSUE #946: Switching to current layout reverts back to the Basic layout
1. Add a button block with basic as layout
2. Turn on buttons #2 & #3
3. Switch layout to Full width ✓ Expected behavior
4. Switch layout and click _Full width layout_ again (Reverts back to Basic layout) ✗ Wrong behavior

_Same thing happens to all layouts of the Button block, not just Full width layout_

### LINKED ISSUE #959: In free version, the premium layouts are showing in layout tab
The layout picker when adding a posts block: 

image

The layout tab after selecting a layout in the layout picker:

image

_**The premium layouts should not display in the layout tab for the Free version**_

blankart:
Fixes #931, fixes #932, fixes #946, fixes #959


src/higher-order/with-design-layout-selector/index.js
```diff
@@ -0,0 +1,244 @@
...
+				</div>
+			) }
+			{ !! designs.length && (
+				<div className="ugb-design-layout-selector__design-library">
+					<div className="components-placeholder__instructions" >
+						{ !! layouts.length && __( 'Or pick from our Design Library.', i18n ) }
+						{ ! layouts.length && __( 'Select a design from our library to start with.', i18n ) }
+					</div>
+					<div className="components-placeholder__fieldset ugb-design-layout-selector__design-container">
+						<div className="ugb-design-layout-selector__design-items">
```
bfintal, line 176: Can you use `ul` here instead of a `div`? You'll have to use `li` on the items inside as well.

Please also do this to the designs area

src/higher-order/with-design-layout-selector/index.js
```diff
@@ -0,0 +1,249 @@
...
+	const itemClassNames = classnames( 'ugb-design-layout-selector__item', {
+		[ className ]: className,
+	} )
+	const imgClassNames = classnames( 'ugb-design-layout-selector__image', {
+		active,
+		'is-premium': ! isPro && otherProps.plan === 'premium',
+	} )
+
+	return (
+		<div className={ itemClassNames } { ...otherProps }>
```
bfintal, line 89: Please use `li` for this instead of a div

src/higher-order/with-design-layout-selector/index.js
```diff
@@ -0,0 +1,249 @@
...
+			{ otherProps.plan && ! isPro && otherProps.plan !== 'free' && (
+				<span className="ugb-design-layout-selector__premium">{ otherProps.plan }</span>
+			) }
+			{ src && <img className={ imgClassNames } src={ src } alt={ label } /> }
```
bfintal, line 93: * Please wrap this in a `<button>` but don't include the label.
* Add a hover effect where the button will become a bit gray `#f0f0f0`, you can use `mix-blend-mode: darken` on the image to get rid of the white background. (please do this for the design buttons too)

bfintal (changes requested):
Other requests aside from the inline requests:
- [x] Seems like the basic button here has a darker border:
image

- [x] If there're no layouts, and there are designs, it's not obvious how to start with the basic design. Maybe we should add a "basic" design as the first entry

- [x] Please remove the panel when the picker is visible
image

- [x] When you just added a block, the text to cancel should read "Skip". When you're switching layouts, the text to cancel should read "Cancel"

- [x] When you're switching layouts and the layout picker is shown, unselecting the block should cancel the layout picker screen.

src/higher-order/with-design-layout-selector/index.js
blankart: Update: I reverted the ul and li elements to div. the ul element offsets the li items to the right

bfintal:
Changed the way how the layout picker is shown. The picker is now shown when the `design` attribute is blank, this makes for a less error-prone approach when showing the layout picker. 

We need to do these changes for the blocks (only the Card block has been updated):

1. Separate the schema into a new file `schema.js`
2. Change the schema for the `design` attribute to have a default value of blank `''`
3. Add a new first entry in `deprecated.js` for updating the old blocks:

```js
import schema from './schema'
import save from './save'

// ...further below

const deprecated = [
	// The design attribute should now be blank be default.
	{
		attributes: {
			...schema,
			design: {
				type: 'string',
				default: 'basic',
			},
		},
		save,
		migrate: attributes => {
			return {
				...attributes,
				design: attributes.design || 'basic',
			}
		},
	},
	// ...older deprecations
]
```

Make sure to replace `basic` with what the default `design` is for the block.

4. Update the `stackable.BLOCK.design.filtered-block-attributes` filter in `design.js` to prevent the `design` attribute from going blank. For example, in the card block:

```js
// Ignore these attributes when exporting / applying designs.
addFilter( 'stackable.card.design.filtered-block-attributes', 'stackable/card', ( attributes, blockAttributes = null ) => {
	return {
		...omit( attributes, [
			...( blockAttributes && blockAttributes.image1Id ? [ 'image1Url' ] : [] ),
			'image1Id',
			...( blockAttributes && blockAttributes.image2Id ? [ 'image2Url' ] : [] ),
			'image2Id',
			...( blockAttributes && blockAttributes.image3Id ? [ 'image3Url' ] : [] ),
			'image3Id',
			'button1Url',
			'button1NewTab',
			'button1NoFollow',
			'button2Url',
			'button2NewTab',
			'button2NoFollow',
			'button3Url',
			'button3NewTab',
			'button3NoFollow',
		] ),
		design: attributes.design || 'basic', // Make sure that the design attribute is filled up or our design picker will appear.
	}
} )
```

bfintal (approved):

PR #911, opened 2020-11-06
Added filter attributes in blog posts

blankart:


src/block/blog-posts/design.js
```diff
@@ -14,3 +14,19 @@ addFilter( 'stackable.blog-posts.design.no-text-attributes', 'stackable/blog-pos
 		'text',
 	] )
 } )
+
+// Ignore these attributes when exporting / applying designs.
+addFilter( 'stackable.blog-posts.design.filtered-block-attributes', 'stackable/button', attributes => {
+	return omit( attributes, [
+		'numberOfItems',
```
bfintal, line 21: Maybe numberOfItems should be included since this can vary per design

bfintal (changes requested):

PR #1007, opened 2020-12-18
Fixes html tag for blog-posts

blankart:


bfintal:
The `blockTag` attribute is reserved for changing the block's main html tag.

PR #1020, opened 2021-01-17
Fixes #1019

### LINKED ISSUE #1019: Google font messed if Global Font Family is not set
If Font Family is not set for Global Setting, it enqueues empty string as Google font, and it makes all google font messed.
Should add a line to the [/src/fonts.php line 92](https://github.com/gambitph/Stackable/blob/5bb5cf8552812de385980a06faa233ff1e218578/src/fonts.php#L92)
`foreach ( $google_fonts as &$font ) {
				if (empty($font)) continue;
				$font = str_replace( ' ', '+', $font ) . ':100,100italic,200,200italic,300,300italic,400,400italic,500,500italic,600,600italic,700,700italic,800,800italic,900,900italic';
			}`

rsm0128:
Fix google font empty string enqueue issue. [#1019 ](https://github.com/gambitph/Stackable/issues/1019)

### DIFF
--- a/src/fonts.php
+++ b/src/fonts.php
@@ -89,7 +89,7 @@ if ( ! class_exists( 'Stackable_Google_Fonts' ) ) {
 			}
 
 			foreach ( $google_fonts as &$font ) {
-				if (!empty($font)) {
+				if ( ! empty( $font ) ) {
 					$font = str_replace( ' ', '+', $font ) . ':100,100italic,200,200italic,300,300italic,400,400italic,500,500italic,600,600italic,700,700italic,800,800italic,900,900italic';
 				}
 			}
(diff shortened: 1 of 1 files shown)


src/fonts.php
```diff
@@ -89,7 +89,9 @@ public static function enqueue_google_fonts( $google_fonts, $handle = 'stackable
 			}
 
 			foreach ( $google_fonts as &$font ) {
-				$font = str_replace( ' ', '+', $font ) . ':100,100italic,200,200italic,300,300italic,400,400italic,500,500italic,600,600italic,700,700italic,800,800italic,900,900italic';
+				if (!empty($font)) {
```
bfintal, line 92:
```suggestion
				if ( ! empty( $font ) ) {
```

rsm0128: Thanks for your suggestion and commit :)

PR #1028, opened 2021-01-26
fixes #1027

### LINKED ISSUE #1027: Block Title spacing not updating in frontend in icon block
add an icon block
turn on block title
adjust block title spacing
see bug in front end

blankart:
fixes #1027

### DIFF
--- a/src/block/icon/index.js
+++ b/src/block/icon/index.js
@@ -56,7 +56,9 @@ export const settings = {
 		'advanced-responsive': true,
 		'block-background': true,
 		'block-separators': true,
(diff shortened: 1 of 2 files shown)


bfintal:
Can we change the fix to something specific only to the icon block? If we update something in the module, it will trigger block errors in our other blocks.

blankart:
> Can we change the fix to something specific only to the icon block? If we update something in the module, it will trigger block errors in our other blocks.

It wont affect any existing blocks since overwrite units is set to false by default

PR #1035, opened 2021-01-28
Fixes #1033

### LINKED ISSUE #1033: Auto block recovery should not automatically recover v1 blocks
You can detect v1 blocks if the migration error was caused by a mismatch in css classes, and the original content didn't have a class `ugb-main-block` and the generated one had it.

blankart:
fixes #1033

Note: v2 Separator blocks don't have `ugb-main-block` class


bfintal:
To address the v2 Separator block issue, don't include the check for `ugb-main-block` when the block is `ugb/separator`

PR #1038, opened 2021-02-02
Fixes #1036. Remove % unit

### LINKED ISSUE #1036: Padding top and bottom does not work
Padding Top & Padding Bottom does not work in separator block when unit is set to '%'

1. add a separator block
2. change padding top & bottom at Spacing
3. set unit to '%', change the value
4. see bug

blankart:
Fixes #1036

### DIFF
--- a/src/modules/advanced-block-spacing/index.js
+++ b/src/modules/advanced-block-spacing/index.js
@@ -306,7 +306,7 @@ const inspectorControls = ( blockName, options ) => ( output, props ) => {
 					<WhenResponsiveScreen screen="desktop">
 						<FourRangeControl
 							label={ __( 'Block Margins', i18n ) }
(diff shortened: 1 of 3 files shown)


src/modules/advanced-block-spacing/index.js
```diff
@@ -735,6 +735,8 @@ const advancedBlockSpacing = ( blockName, options = {} ) => {
 		enablePaddingRight: true,
 		enablePaddingBottom: true,
 		enablePaddingLeft: true,
+		overwritePaddingUnits: false,
+		overwriteMarginUnits: false,
```
bfintal, line 739: Instead of doing a default value of false for `overwritePaddingUnits`, we can use `[ 'px', 'em', '%'  ]` as the default value for `overwritePaddingUnits` and then pass it directly to the Control

bfintal (changes requested):

PR #1075, opened 2021-02-26
Clickable Containers

blankart:
### List of Blocks:
- [x] Blockquote
- [x] CTA
- [x] Card
- [x] Column
- [x] Container
- [x] Count up
- [x] Feature Grid
- [x] Feature
- [x] Header
- [x] Notification
- [x] Number Box
- [x] Pricing Box
- [x] Team Member
- [x] Testimonial

### TODOs
- [x] Handle block deprecation
- [x] Test Code
- [ ] Test Blocks


bfintal (changes requested):
Enhance accessibility of the anchor element by adding these attributes:

```aria-hidden="true" tabindex="-1"```

**Review by bfintal — CHANGES_REQUESTED** [review]
If only 1 link is available, don't display "#1" or "1" in the option label

**Review by bfintal — CHANGES_REQUESTED** [review]
The container link doesn't work if there is a background video on the container

**Review by bfintal — CHANGES_REQUESTED** [review]
For the Feature Block, the container link occupies the entire block instead of just the container.

File: src/components/div-background/index.js [hunk 1]
```diff
@@ -21,9 +27,10 @@ const DivBackground = props => {
 	} )
 
 	return (
-		<BlockTag className={ divClasses } { ...propsToPass }>
+		<BlockTag className={ divClasses } { ...omit( propsToPass, 'index' ) }>
 			{ props.children }
 			{ showBackground && showVideoBackground && createVideoBackground( backgroundAttrName, blockProps ) }
```
bfintal, line 32: I placed this back here so that we retain it for v3

src/modules/clickable-container/index.js
```diff
@@ -0,0 +1,227 @@
...
+	const {
+		columns = 1,
+		showClickableContainer = false,
+	} = props.attributes
+
+	return (
+		<Fragment>
+			{ output }
+			<PanelAdvancedSettings
+				title={ __( 'Clickable Containers', i18n ) }
```
bfintal, line 118:
```suggestion
				title={ __( 'Container Link', i18n ) }
```

PR #1086, opened 2021-03-05
Fixes blockquote typography styles.

### LINKED ISSUE #1082: Blockquote block shows a different font weight in front end when set to default
When font weight is set to default, it becomes greater when it is shown in the front end. Expected behavior is for the font weight in the front end and back end to be the same.

1. Create blockquote block
2. By default, the font weight should be set to "Default"
3. Switch to front end and compare font weight with back end

Screenshots:
firefox_nwRobvk95A

blankart:
Fixes #1082


bfintal:
This change will affect all existing Blockquote blocks. Since this is just a minor issue, I think we can set #1082 to won't fix.

PR #1122, opened 2021-03-24
Auto block recovery for reusable blocks

blankart:


bfintal:
@blankart I removed the auto-saving of the reusable block, let's leave the saving to the user. Our main goal is to make sure that the user doesn't encounter an error visually.

PR #1136, opened 2021-04-16
(V3) Advanced Text

blankart:


src/block-components/button/attributes.js
```diff
@@ -0,0 +1,51 @@
...
+
+export const addAttributes = ( attrObject, options = {} ) => {
+	const {
+		selector = '.stk-button__button',
+	} = options
+	attrObject.add( {
+		attributes: {
+			buttonPadding: {
+				stkResponsive: true,
+				type: 'object',
```
bfintal, line 12: you can use `stkUnits: 'px'` here instead of defining your own `buttonPaddingUnit` below

src/block-components/button/attributes.js
```diff
@@ -0,0 +1,51 @@
...
+				stkHover: true,
+				type: 'string',
+				default: '',
+			},
+			buttonBackgroundGradientDirection: {
+				stkHover: true,
+				type: 'number',
+				default: '',
+			},
+			text: {
```
bfintal, line 39: Since this attribute is used primarily for buttons, please use `buttonText` instead of button to prevent conflicts in with other block components

src/block-components/button/edit.js
```diff
@@ -0,0 +1,124 @@
...
+										allowReset={ true }
+										hover="all"
+									/>
+								</Fragment>
+							) }
+						</Fragment>
+					</BaseControl>
+
+					<BorderControls
+						blockEl={ blockEl }
```
bfintal, line 105: Please note that I'm still not sure whether we should keep using `blockEl`. Keep this for now, but we'll decide on a better way to do this in the future

src/block-components/button/edit.js
```diff
@@ -0,0 +1,124 @@
...
+									},
+									{
+										value: 'gradient',
+										title: __( 'Gradient', i18n ),
+									},
+								] }
+								attribute="buttonBackgroundColorType"
+								isSmall={ true }
+								fullwidth={ false }
+								hover="all"
```
bfintal, line 71: added `hover="all"` or `responsive="all"` to color types make things more complicated, don't use hover and responsive for this. You can check https://github.com/gambitph/Stackable/tree/v3/src/block-components/helpers/backgrounds on how it was used there

src/block-components/typography/attributes.js
```diff
@@ -0,0 +1,92 @@
...
+			fontSize: {
+				stkResponsive: true,
+				stkHover: true,
+				type: 'number',
+				default: '',
+				stkUnits: 'px',
+			},
+			lineHeight: {
+				stkResponsive: true,
+				stkHover: true,
```
bfintal, line 19: Don't use hover states for this attribute.

src/block-components/typography/edit.js
```diff
@@ -0,0 +1,265 @@
...
+				title={ __( 'Text', i18n ) }
+				initialOpen={ true }
+				id="text"
+				{ ...{
+					...( withToggle ? {
+						hasToggle: withToggle,
+						checked: getAttribute( 'showText' ),
+						onChange: updateAttributeHandler( 'showText' ),
+					} : {} ),
+				} }
```
bfintal, line 66: I don't think we'll use this. Let's just remove this.

src/block-components/typography/edit.js
```diff
@@ -0,0 +1,265 @@
...
+			>
+				<Fragment>
+					{ enableTextContent && (
+						<TextControl
+							label={ sprintf( __( '%s Content', i18n ), label ) }
+							value={ unescape( getAttribute( 'text' ) ) }
+							onChange={ value => updateAttribute( 'text', escapeHTML( value ) ) }
```
bfintal, line 73: Please add a placeholder `Enter text here`

src/block-components/typography/edit.js
```diff
@@ -0,0 +1,265 @@
...
+						hasToggle: withToggle,
+						checked: getAttribute( 'showText' ),
+						onChange: updateAttributeHandler( 'showText' ),
+					} : {} ),
+				} }
+			>
+				<Fragment>
+					{ enableTextContent && (
+						<TextControl
+							label={ sprintf( __( '%s Content', i18n ), label ) }
```
bfintal, line 71:
```suggestion
							label={ __( 'Content', i18n ) }
```

src/block-components/typography/edit.js
```diff
@@ -0,0 +1,265 @@
...
+							/>
+						</Fragment>
+					) }
+					{ ! disableAlign && (
+						<AlignButtonsControl
+							label={ __( 'Align', i18n ) }
+							attribute="textAlign"
+							responsive="all"
+						/>
+					) }
```
bfintal, line 249: Let's remove this. If there's alignment, the block should just use the `Alignment` block component

src/block-components/typography/edit.js
```diff
@@ -0,0 +1,265 @@
...
+					...( withToggle ? {
+						hasToggle: withToggle,
+						checked: getAttribute( 'showText' ),
+						onChange: updateAttributeHandler( 'showText' ),
+					} : {} ),
+				} }
+			>
+				<Fragment>
+					{ enableTextContent && (
+						<TextControl
```
bfintal, line 70: I changed this to TextControl, but maybe we should have a prop `isMultiline: false` that you can use to make the content control into a TextareaControl

src/block-components/typography/edit.js
```diff
@@ -0,0 +1,265 @@
...
+import { getAttributeName } from '~stackable/util'
+
+export const Edit = props => {
+	const {
+		label,
+		enableTextTag,
+		enableTextContent,
+		attrNameTemplate,
+		disableAlign,
+		withToggle,
```
bfintal, line 42: For the enable props, let's use the prefix `has`. `hasTextTag` `hasTextContent`

src/block-components/typography/style.js
```diff
@@ -0,0 +1,204 @@
...
+						value = `${ clampTabletValue }${ getAttribute( 'fontSizeUnit', device, state ) || 'px' }`
+					} else if ( clampDesktopValue || getAttribute( 'fontSize', device, state ) !== '' ) {
+						value = undefined
+					}
+
+					if ( getAttribute( 'fontSize', device, state ) !== '' ) {
+						value = `${ getAttribute( 'fontSize', device, state ) }${ getAttribute( 'fontSizeUnit', device, state ) || 'px' }`
+					}
+				}
+
```
bfintal, line 62: you can use `clampCallback` to generate the clamped value based on the inherited value

src/block-components/typography/style.js
```diff
@@ -0,0 +1,204 @@
...
+	return [
+		{
+			selector,
+			styleRule: 'fontSize',
+			attrName: 'fontSize',
+			hasUnits: 'px',
+			responsive: 'all',
+			hover: 'all',
+			hoverSelector,
+			valueCallback: ( _value, getAttribute, device, state ) => {
```
bfintal, line 27: You can use `clampCallback` instead to generate the clamped value based on the inherited value

src/block-components/typography/style.js
```diff
@@ -0,0 +1,204 @@
...
+			selector,
+			styleRule: 'fontFamily',
+			attrName: 'fontFamily',
+			valueCallback: value => getFontFamily( value ),
+		},
+		{
+			selector,
+			styleRule: 'letterSpacing',
+			attrName: 'letterSpacing',
+			hasUnits: 'px',
```
bfintal, line 142: `hasUnits` is used alongside `hover="all"` and `stkHover`. Since you're using this to format the CSS rule, use `format: '%spx'` instead.

src/block-components/typography/style.js
```diff
@@ -0,0 +1,204 @@
...
+		selector = '',
+		inherit = true, // If false, desktop styles will only be applied to desktop, etc.
+		inheritMax = 50, // If provided & inherit is true, clamp the inherited value in tablet and mobile to this.
+		inheritMin,
+		hoverSelector = '',
+	} = options
+
+	const styles = getStyles( attributes, getStyleParams( {
+		selector, inherit, inheritMin, inheritMax, hoverSelector,
+	} ) )
```
bfintal, line 194: Pass `options` directly, then deconstruct inside the `getStyleParams` function instead
```suggestion
	const styles = getStyles( attributes, getStyleParams( options ) )
```

src/block-components/typography/edit.js
blankart: I believe this is important to achieve buttons without texts.

src/block-components/button/attributes.js
blankart: The `text` attribute should not be included in the button block-component. I will remove this.

src/block-components/button/edit.js
bfintal: Please remove the `hover="all"` here

bfintal (changes requested):
Findings:
- [x] Button group's bottom margin style isn't generated
- [ ] Add a Text Shadow icon button popup option in the Typography panel
- [x] Button's Gradient Direction on hover doesn't work
- [ ] Button's Custom CSS doesn't work
- [ ] Pressing `enter` on the heading and text blocks should create a new text block

src/util/styles/style-object.js
```diff
@@ -303,15 +303,42 @@ class StyleObject {
...
+		} else {
+			parentHoverSelector = `:where(.stk-hover-parent:hover) .%s ${ selector }`
+			hoverSelector = _hoverSelector || `.stk-block.%s:hover ${ selector }`
+			// This is for the editor, change the selector to make the styles show up right away.
+			if ( blockState === 'hover' ) {
+				hoverSelector = `.%s.stk--is-hovered ${ selector }`
+			}
+			if ( blockState === 'parent-hovered' ) {
+				parentHoverSelector = `.%s.stk--is-hovered ${ selector }`
+			}
```
bfintal, line 341: This looks a bit messy. Can you make a function with the structure:

```js
const parependClass = ( prependString, selector ) => {
    let combinedSelector = ''
    // ...Handle if selector is a string or array
    return combinedSelector
}
```

The function can do the splitting and appending of the class. It should already handle whether the given selector is a string or an array.

It can be called like this:
```js
parentHoverSelector = parependClass( ':where(.stk-hover-parent:hover) .%s', selector )
hoverSelector = _hoverSelector || parependClass( '.stk-block.%s:hover', selector )
```

bfintal (changes requested):

bfintal (changes requested):

src/block-components/icon/edit.js
```diff
@@ -0,0 +1,320 @@
...
+							hover="all"
+						/>
+
+						<ColorPaletteControl
+							label={ getAttribute( 'iconColorType', 'desktop', state ) === 'gradient' ? sprintf( __( 'Icon Color #%s', i18n ), 1 )
+								: __( 'Icon Color', i18n ) }
+							attribute="iconColor1"
+							hover="all"
+						/>
+						{ getAttribute( 'iconColorType', 'desktop', state ) === 'gradient' && (
```
bfintal, line 82: Let's make things simpler, the `iconColorType` should not be responsive and no hover. If the color type is gradient, it should be gradient for all devices and all hover states.

src/block-components/icon/edit.js
```diff
@@ -0,0 +1,320 @@
...
+									placeholder={ 20 }
+								/>
+
+								<ToggleControl
+									label={ __( 'Shape Outline', i18n ) }
+									checked={ getAttribute( 'shapeOutline', 'desktop', state ) }
+									onChange={ updateAttributeHandler( 'shapeOutline', 'desktop', state ) }
+								/>
+
+								{ getAttribute( 'shapeOutline', 'desktop', state ) && (
```
bfintal, line 207: As with color types, let's keep things simple. shape outlines should be not be responsive or hoverable. the shape's color, opacity and other attributes can be changed

src/block-components/icon/font-awesome-icon.js
```diff
@@ -0,0 +1,79 @@
...
+	}, [] )
+
+	const propsToPass = pick( props, [ 'className', 'color', 'fill', 'style' ] )
+
+	// If given an svg, just display it.
+	if ( typeof props.value === 'string' && props.value.match( /^<svg/ ) ) {
+		return <RawHTML { ...propsToPass }>{ prependRender + props.value }</RawHTML>
+	}
+
+	if ( ! faIsAPILoaded() ) {
```
bfintal, line 37: I believe this file was referenced from `src/components/font-awesome-icon/index.js`, put back some of the code comments which are not present here:

```suggestion
	// There's a chance that the Font Awesome library hasn't loaded yet, wait for it.
	if ( ! faIsAPILoaded() ) {
```

src/block-components/icon/font-awesome-icon.js
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	const prefix = props.value ? props.value.replace( /-.*$/, '' ) : props.prefix
+	const iconName = props.value ? props.value.replace( /^.*?-/, '' ) : props.iconName
+
+	// Display the icon.
+	if ( prefix && iconName ) {
+		const iconHTML = faGetSVGIcon( prefix, iconName )
+
+		if ( ! iconHTML ) {
```
bfintal, line 48:
```suggestion
		// Just in case the icon hasn't loaded yet, wait for it.
		if ( ! iconHTML ) {
```

src/block-components/icon/font-awesome-icon.js
```diff
@@ -0,0 +1,79 @@
...
+	const iconHTML = faGetSVGIcon( 'far', 'smile' )
+	return <RawHTML { ...propsToPass }>{ prependRender + iconHTML }</RawHTML>
+}
+
+FontAwesomeIcon.Content = props => {
+	const propsToPass = pick( props, [ 'className', 'color', 'fill', 'style' ] )
+
+	const prependRender = props.prependRender ? renderToString( props.prependRender ) : ''
+
+	if ( typeof props.value === 'string' ) {
```
bfintal, line 66:
```suggestion
	// If given an svg, just display it.
	if ( typeof props.value === 'string' ) {
```

src/block-components/icon/font-awesome-icon.js
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}
+
+	const prefix = props.value ? props.value.replace( /-.*$/, '' ) : props.prefix
+	const iconName = props.value ? props.value.replace( /^.*?-/, '' ) : props.iconName
+
+	const iconHTML = faGetSVGIcon( prefix, iconName )
+	return <RawHTML { ...propsToPass }>{ prependRender + iconHTML }</RawHTML>
+}
+
```
bfintal, line 78: Please re-add the `defaultProps`

src/block-components/icon/style.js
```diff
@@ -0,0 +1,298 @@
...
+		uniqueId = '',
+		wrapperSelector = '',
+		backgroundShapeSelector = '.stk--shape-icon',
+		hoverSelector = '',
+	} = options
+
+	return [
+		// Icon Styles
+		{
+			selector: `${ selector } svg:last-child`,
```
bfintal, line 22: I don't think we need to target `:last-child`
```suggestion
			selector: `${ selector } svg`,
```

src/block/button/style.js
```diff
@@ -0,0 +1,117 @@
...
+			<Icon.Style
+				{ ...{
+					...propsToPass,
+					options: {
+						...propsToPass.options,
+						selector: '.stk-button__button > .stk-button__svg-wrapper > .stk-button__inner-svg',
+						hoverSelector: '.stk-button__button:hover > .stk-button__svg-wrapper > .stk-button__inner-svg',
+						wrapperSelector: '.stk-button__button',
+					},
+				} } />
```
bfintal, line 57: If you adjust `block-components/icon/style.js` and remove `wrapperSelector` this works fine:

```suggestion
			<Icon.Style { ...propsTopass } />
```

src/block-components/button/style.js
```diff
@@ -0,0 +1,237 @@
...
+
+				return `linear-gradient(${ buttonBackgroundGradientDirection !== '' ? buttonBackgroundGradientDirection + 'deg' : '90deg' }, ${ buttonBackgroundColor || buttonBackgroundColor2 }, ${ buttonBackgroundColor2 || buttonBackgroundColor })`
+			},
+			dependencies: [ 'buttonBackgroundGradientDirection', 'buttonBackgroundColor', 'buttonBackgroundColor2', 'buttonBackgroundColorType' ],
+		},
+		{
+			selector: `${ selector }:after`,
+			styleRule: 'background',
+			attrName: 'buttonBackgroundColor',
+			hover: 'all',
```
bfintal, line 88: The hover is incorrectly added to the block instead of the button. The `hoverSelector` should be added:

```suggestion
			hover: 'all',
			hoverSelector: `${ selector }:hover:after`,
```

src/block-components/icon/index.js
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * WordPress dependencies
+ */
+import { useBlockEditContext } from '@wordpress/block-editor'
+import {
+	useMemo, Fragment, useState, useRef, useEffect, useCallback,
+} from '@wordpress/element'
+import { IconSearchPopover } from '~stackable/components'
+
+const SVGIcon = props => {
```
bfintal, line 28: Is it possible to use the existing SVGIcon component in `src/components/svg-icon`?

src/styles/block.scss
```diff
@@ -174,3 +147,64 @@
...
+
+	a.stk-button__button {
+		position: relative;
+		display: flex;
+		justify-content: center;
+		align-items: center;
+		column-gap: 4px;
+		z-index: 1;
+
+		&::after {
```
bfintal, line 197: Can we add `pointer-events: none` to `:after`

src/block-components/button/style.js
blankart: Line 116 is responsible for the button hover. This line only adds the overlay gradient which will appear after when hovered.

src/block-components/icon/style.js
blankart: We need this to avoid resizing the svg with linearGradient tag

src/block-components/link/index.js
```diff
@@ -0,0 +1,91 @@
...
+	} = props
+
+	return (
+		<LinkComponent.Content
+			className={ props.className }
+			href={ attributes.linkUrl || undefined }
+			target={ attributes.linkNewTab ? '_blank' : undefined }
+			rel={ attributes.linkRel || undefined }
+			aria-hidden="true"
+			tabindex="-1"
```
bfintal, line 81: Please remove, this is not good for accessibility, normal buttons should not be hidden

bfintal (changes requested):

src/block/heading/style.js
```diff
@@ -0,0 +1,87 @@
...
+		<Fragment>
+			<Alignment.Style { ...propsToPass } />
+			<BlockDiv.Style { ...propsToPass } />
+			<Advanced.Style { ...propsToPass } />
+			<Typography.Style { ...{
+				...propsToPass,
+				options: {
+					...propsToPass.options,
+					selector: '.stk-heading__text',
+					hoverSelector: '.stk-heading__text:hover',
```
bfintal, line 43: No need for this line. For the most part, we would like the effects to be applied when the block is hovered. There are some exceptions, like the Button Block since it will only look correct if the effect triggers when actually hovering over the button

src/block/heading/edit.js
```diff
@@ -0,0 +1,90 @@
...
+
+	return (
+		<Fragment>
+
+			<InspectorTabs />
+
+			<Alignment.InspectorControls />
+			<BlockDiv.InspectorControls />
+			<Advanced.InspectorControls />
+			<Typography.InspectorControls blockEl={ blockEl } blockElSelector=".stk-heading__text" hasRemoveMargins={ true } />
```
bfintal, line 68: The `blockElSelector` and `blockEl` props seem to be unused

src/block-components/typography/edit.js
bfintal: For icon-only buttons, let's just create a new block for them

PR #1173, opened 2021-05-03
Feat Clickable Containers and Blog Posts Pagination

blankart:


src/block/blog-posts/attributes.php
```diff
@@ -70,8 +70,8 @@ function stackable_blog_posts_attributes() {
 			),
 			'taxonomyFilterType' => array(
 				'type' => 'string',
-    	    	'default' => '__in',
-    		),
+				'default' => '__in',
+  		),
```
bfintal, line 74:
```suggestion
			),
```

PR #1177, opened 2021-05-06
Use get_post_thumbnail for getting post thumbnails

blankart:


src/block/blog-posts/index.php
```diff
@@ -206,13 +206,18 @@ function stackable_render_blog_posts_block( $attributes, $content ) {
...
-						'<figure class="%s"><a href="%s"><img src="%s" alt="%s" width="%s" height="%s"/></a></figure>',
+						'<figure class="%s"><a href="%s">%s</a></figure>',
 						esc_attr( $featured_image_classes ),
 						esc_url( get_permalink( $post_id ) ),
-						esc_url( $featured_image_src[0] ),
-						esc_attr( get_the_title( $post_id ) ),
-						esc_attr( $featured_image_src[1] ),
-						esc_attr( $featured_image_src[2] )
+						get_the_post_thumbnail(
+							$post['ID'],
```
bfintal, line 213: What's the difference if we use this?
```suggestion
							$post_id,
```

PR #1189, opened 2021-05-11
Fixes #1185

### LINKED ISSUE #1185: Video Popup Block CSS issue
https://d.pr/v/iRv14u

happens because of the `all: unset` property

blankart:
Fixes #1185


bfintal:
@blankart we can't remove the `all: unset`

what if we do `top: 0; left: 0; right: 0; bottom: 0`

PR #1216, opened 2021-06-07
Feat Custom Attributes

blankart:


src/modules/advanced-custom-attributes/__test__/index.test.js
```diff
@@ -0,0 +1,37 @@
...
+		} = render( <CustomAttributesControl { ...dummyProps } /> )
+		expect( getByText( 'Custom Attributes' ) ).toBeTruthy()
+
+		const invalidInputs = [
+			'test',
+			'key=value',
+			'key="value" key2=invalid',
+			'style="color: red;"',
+			'dangerouslySetInnerHTML="content"',
+			'ref="sample ref"',
```
bfintal, line 25: Can you also test these incorrectly formatted or odd strings?

```js
'key="value'
'key=\'"value'
'key-_!@#="value"'
'key="value" key="value"'
'key="value"key="value"'
'key="<div>"'
```

src/modules/advanced-custom-attributes/index.js
```diff
@@ -0,0 +1,191 @@
...
+const createAddSaveProps = ( extraProps, blockProps ) => {
+	if ( ! Array.isArray( blockProps.attributes.customAttributes ) || blockProps.attributes.customAttributes.length === 0 ) {
+		return extraProps
+	}
+
+	const customAttributes = {}
+	try {
+		blockProps.attributes.customAttributes.forEach( ( [ key, _value ] ) => {
+			let value = _value
+			const dynamicAttributeMatch = value.match( /%[^\%]*%/g )
```
bfintal, line 38:
```suggestion
			const dynamicAttributeMatch = value.match( /%[^\%]+%/g )
```

src/modules/advanced-custom-attributes/index.js
```diff
@@ -0,0 +1,191 @@
...
+	'customAttributes',
+]
+
+const createAddSaveProps = ( extraProps, blockProps ) => {
+	if ( ! Array.isArray( blockProps.attributes.customAttributes ) || blockProps.attributes.customAttributes.length === 0 ) {
+		return extraProps
+	}
+
+	const customAttributes = {}
+	try {
```
bfintal, line 35: Use the `try` statement inside the `forEach` so that if it encounters an error, it can still proceed.

src/modules/advanced-custom-attributes/index.js
```diff
@@ -0,0 +1,191 @@
...
+const INVALID_BLOCK_ATTRIBUTES = [
+	'customAttributes',
+]
+
+const createAddSaveProps = ( extraProps, blockProps ) => {
+	if ( ! Array.isArray( blockProps.attributes.customAttributes ) || blockProps.attributes.customAttributes.length === 0 ) {
+		return extraProps
+	}
+
+	const customAttributes = {}
```
bfintal, line 34: Instead of defining a blank `customAttributes` then filling it up, it might be better to adjust the values directly.

```js
const customAttributes = Object.fromEntries( blockProps.attributes.customAttributes )
Object.keys( customAttributes ).forEach( key => {
    try {
        const value = customAttributes[ key ]
        // Replace the value if it's dynamic
        const dynamicAttributeMatch = value.match( /%([^\%]+)%/g )
        // ...
        customAttributes[ key ] = newValue
    } catch {}
} )
```

src/modules/advanced-custom-attributes/index.js
```diff
@@ -0,0 +1,191 @@
...
+						}
+					} catch {
+						throw new Error( sprintf( __( "Attribute key '%s' is not allowed.", i18n ), key ) )
+					}
+
+					newCustomAttributes.push( [ key, value ] )
+				} )
+
+				props.onChange( newCustomAttributes )
+				setError( '' )
```
bfintal, line 137: I'd like to recommend something simpler. Please see https://developer.mozilla.org/en-US/docs/Web/API/Element/attributes

You can grab the input of the user, then pass it on as `innerHTML` of a dummy element:

```js
const el = document.createElement( 'div' )
el.innerHTML = `<div ${ stringValue } />`
// Get all the "fixed" attributes:
const attrNodeMap = el.children[0].attributes // This is a NamedNodeMap
const attributes = {}
for (let i = attrNodeMap.length - 1; i >= 0; i--) {
    attributes[ attrNodeMap[ i ].name ] = attrNodeMap[ i ].value
}
```

What you can do is instead of detecting and displaying an error, you can have the above method adjust the value when the input is blurred

src/modules/advanced-custom-attributes/index.js
```diff
@@ -0,0 +1,191 @@
...
+const addInspectorControls = ( output, props ) => {
+	return (
+		<Fragment>
+			{ output }
+			<PanelAdvancedSettings
+				title={ __( 'Custom Attributes', i18n ) }
+				initialOpen={ false }
+			>
+				<p className="components-base-control__help">
+					{ __( 'Add custom HTML attributes to your Stackable blocks.', i18n ) }
```
bfintal, line 169:
```suggestion
					{ __( 'You can type in custom HTML attributes for this block below.', i18n ) }
```

src/modules/advanced-custom-attributes/index.js
```diff
@@ -0,0 +1,191 @@
...
+					&nbsp;
+					<a href="https://docs.wpstackable.com" target="_docs">{ __( 'Learn more about Custom Attributes', i18n ) }</a>
+				</p>
```
bfintal, line 172: Instead of adding this here, add this as the `help` prop of the `CustomAttributesControl`

Also change the text to:
```js
help={ __( 'You can type in custom HTML attributes for this block below.', i18n ) }
```

src/modules/advanced-custom-attributes/index.js
blankart: I intentionally separated the custom attributes to properly omit invalid values from `INVALID_HTML_ATTRIBUTES`.

src/components/custom-attributes-control/index.js
```diff
@@ -0,0 +1,87 @@
...
+const CustomAttributesControl = props => {
+	const [ customAttributes, setCustomAttributes ] = useState(
+		Array.isArray( props.value ) ?
+			props.value.map( attribute => {
+				const [ key, _value ] = attribute
+				const value = `"${ _value }"`
+				return [ key, value ].join( '=' )
+			} ).join( ' ' ) :
+			''
+	)
```
bfintal, line 24: Might need a `useEffect` here to adjust things when the `prop.value` changes externally

src/modules/advanced-custom-attributes/index.js
```diff
@@ -0,0 +1,102 @@
...
+					if (
+						! INVALID_BLOCK_ATTRIBUTES.includes( match ) &&
+						blockProps.attributes.hasOwnProperty( match ) &&
+						! isUndefined( blockProps.attributes[ match ] )
+					) {
+						value = value.replace( _match, striptags( blockProps.attributes[ match ].toString() ) )
+					}
+				} )
+			}
+			customAttributes[ key ] = value
```
bfintal, line 56:
```suggestion
				customAttributes[ key ] = value
			}
```

src/modules/advanced-custom-attributes/index.js
```diff
@@ -0,0 +1,102 @@
...
+	return (
+		<Fragment>
+			{ output }
+			<PanelAdvancedSettings
+				title={ __( 'Custom Attributes', i18n ) }
+				initialOpen={ false }
+			>
+
+				<CustomAttributesControl
+					label={ __( 'Custom Attributes' ) }
```
bfintal, line 84:
```suggestion
					label={ __( 'Custom Attributes', i18n ) }
```

bfintal (changes requested):

PR #1218, opened 2021-06-08
Container padding fix

### LINKED ISSUE #1217: Block paddings for tablet are capping at 30

blankart:
Fixes #1217

### DIFF
--- a/src/components/column-padding-control/index.js
+++ b/src/components/column-padding-control/index.js
@@ -48,7 +48,7 @@ const ColumnPaddingControl = props => {
 			className="ugb--help-tip-advanced-column-paddings"
 			defaultLocked={ true }
 			min={ [ 0, 0, 0 ] }
-			max={ [ 500, 30, 100 ] }
+			max={ [ 500, 100, 100 ] }
 			step={ 1 }
 
 			valueDesktop={ {
--- a/src/components/column-padding-control/index.js
+++ b/src/components/column-padding-control/index.js
@@ -48,7 +48,7 @@ const ColumnPaddingControl = props => {
 			className="ugb--help-tip-advanced-column-paddings"
 			defaultLocked={ true }
 			min={ [ 0, 0, 0 ] }
-			max={ [ 500, 100, 100 ] }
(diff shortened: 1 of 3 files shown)


bfintal (changes requested):
I don't think you need to specify the `sliderMin` and `sliderMax` values here. From what I know the range control should be able to handle things as expected if the `sliderMin` and `sliderMax` values are not provided.

src/components/column-padding-control/index.js
```diff
@@ -48,7 +48,7 @@ const ColumnPaddingControl = props => {
 			className="ugb--help-tip-advanced-column-paddings"
 			defaultLocked={ true }
 			min={ [ 0, 0, 0 ] }
-			max={ [ 500, 30, 100 ] }
+			sliderMax={ [ 500, 100, 100 ] }
```
bfintal, line 51: Aside from this, are there other low tablet values which are clamped lower than mobile? it may be a good time to increase those if there are any.

src/components/spacing-control/index.js
```diff
@@ -20,6 +20,8 @@ const SpacingControl = props => {
 	// Allow different sets of min/max/step values per screen size.
 	const min = Array.isArray( props.min ) ? props.min : [ props.min, props.min, props.min ]
 	const max = Array.isArray( props.max ) ? props.max : [ props.max, props.max, props.max ]
+	const sliderMin = Array.isArray( props.sliderMin ) ? props.sliderMin : [ props.sliderMin, props.sliderMin, props.sliderMin ]
```
bfintal, line 23: I don't think you need to specify the `sliderMin` and `sliderMax` values here. From what I know the range control should be able to handle things as expected if the `sliderMin` and `sliderMax` values are not provided.

blankart: We need to add this in order to pass the corresponding`sliderMin` and`sliderMax` value inside the `FourRangeControl` component

blankart: Additional note: Similar to `min` and `max`, `sliderMin` and `sliderMax` also handle number values `e.g. sliderMax={ 100 }`

PR #1260, opened 2021-06-30
Fixes #1255

### LINKED ISSUE #1255: button alignment not reflecting in frontend
Steps
turn on optimization setting
in the editor, add a columns block
inside one column, add a button block
adjust the text alignment of the button block (in the toolbar) to 'Align text left'
in backend, the button aligns to the left, but in frontend, it's still centered.

blankart:
Fixes #1255

### DIFF
--- a/src/block/button/style.js
+++ b/src/block/button/style.js
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import {
+	appendImportant,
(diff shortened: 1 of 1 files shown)


bfintal:
@blankart can you check whether this is also present in the v3 button block?

PR #1274, opened 2021-07-05
V3 typography block component

blankart:


bfintal (changes requested):
I think this doesn't need to be inside the typography panel. Can you add a General tab at the start of the Style tab, then add the following inside it:
- Columns
- Column gap (px)
- Alignment (find a way to add the alignment controls here instead of in the Block Tab)

blankart:
> I think this doesn't need to be inside the typography panel. Can you add a General tab at the start of the Style tab, then add the following inside it:
> 
> - Columns
> 
> - Column gap (px)
> 
> - Alignment (find a way to add the alignment controls here instead of in the Block Tab)

Are we going to include the options inside the Typography Module? 

It might cause some problems when dealing with multiple General Tabs from other modules in the future

bfintal:
> Are we going to include the options inside the Typography Module?
> 
> It might cause some problems when dealing with multiple General Tabs from other modules in the future

Yea include it in the text block since it won't be re-used in others

blankart:
> > Are we going to include the options inside the Typography Module?
> > It might cause some problems when dealing with multiple General Tabs from other modules in the future
> 
> @blankart Yea include it in the text block since it won't be re-used in others
> 
> * So the columns & column gap should be in the text block, no need for a block component since they will be specific to the text block only.
> * The top and bottom lines should be in the heading block. No need also for a block component since they're only for the heading block

Addressed in https://github.com/gambitph/Stackable/pull/1274/commits/d9255f135b94bb3ffbf22e89a900555d61bef04f

PR #1297, opened 2021-07-08
Fixes #1288

### LINKED ISSUE #1288: block paddings not applying in frontend
turn on optimization setting
1. add a blockquote
2. adjust paddings in advanced tab
3. see bug in frontend.

tablet and mobile does not work too

blankart:
Fixes #1288

### DIFF
--- a/src/block/blockquote/style.scss
+++ b/src/block/blockquote/style.scss
@@ -7,7 +7,7 @@
 @import "common";
 
 // Get rid of possible default blockquote styling.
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
We need the padding here since most `blockquote` tags have this default styling applied by themes. We will need a style generation fix for this.

blankart:
The style generation code already uses `appendImportant` for the paddings. The frontend padding style overrides it.

PR #1312, opened 2021-07-08
[V3] Fixes #1302

### LINKED ISSUE #1302: left, right Block padding of divider not working
turn on optimization setting
add a divider block
In advanced > block spacing, adjust lef & right paddings
see bug in frontend

blankart:


bfintal (changes requested):
We'll need an appendImportant fix here so that existing designs will not get affected.

PR #1313, opened 2021-07-10
[V3] Dynamic Content

blankart:


src/components/image-control2/editor.scss
```diff
@@ -15,14 +15,23 @@
...
-			background: #fafafa;
+		&:not(.disabled) {
+			cursor: pointer;
+			&:hover {
+				background: #fafafa;
+			}
+			&:focus {
+				background: #fafafa;
+				box-shadow: 0 0 0 2px rgba(30, 140, 190, 0.8);
+			}
```
blankart, line 27: the component is not using `disabled` class already. revert the changes

src/components/image-control2/editor.scss
```diff
@@ -15,14 +15,23 @@
...
 		}
-		&:focus {
-			background: #fafafa;
-			box-shadow: 0 0 0 2px rgba(30, 140, 190, 0.8);
+		.dashicons {
+			pointer-events: none;
+			color: #e6e6e6;
+			font-size: 50px;
+			height: 50px;
+			width: 50px;
```
blankart, line 34: check if this is still needed

src/components/dynamic-content-control/index.js
```diff
@@ -0,0 +1,257 @@
...
+			/>
+			{ props.isPopoverOpen && (
+				<Popover
+					position="top right"
+					className="stackable-dynamic-content__popover"
+				>
+					{ ! isPro && (
+						<ProControl
+							title={ __( 'Say Hello to Dynamic Attributes', i18n ) }
+							description={ __( 'Sample Description.', i18n ) }
```
blankart, line 203: update the text

bfintal (changes requested):
Some feedback
1. I'm seeing some React warnings like `Warning: React does not recognize the `isDynamic` prop on a DOM element`
2. When I create a new page, add a title, add a new adv. heading block, then set the attribute to Current Post > Post Title, it doesn't display the title I just inputted. It shows "Auto Draft"
3. As with 2 above, this also happens when you create a new page, assign a featured image, then immediately use the featured image in the a dynamic content.

Regarding points 2 & 3, the current behavior looks like the dynamic fields data at least for the current post are prepared ahead of time, and aren't refreshed

blankart:
> Some feedback
> 
> 1. I'm seeing some React warnings like `Warning: React does not recognize the `isDynamic` prop on a DOM element`
> 2. When I create a new page, add a title, add a new adv. heading block, then set the attribute to Current Post > Post Title, it doesn't display the title I just inputted. It shows "Auto Draft"
> 3. As with 2 above, this also happens when you create a new page, assign a featured image, then immediately use the featured image in the a dynamic content.
> 4. For content text controls, when you use dynamic content alongside static text, e.g. `<span ...>My Title</span> Blahblah`, the input control should display `Post Title Blahblah`. Right now it just displays `Post Title`
> 
> Regarding points 2 & 3, the current behavior looks like the dynamic fields data at least for the current post are prepared ahead of time, and aren't refreshed

I will address the react warnings.

Regarding #2, this is the current behavior of our Dynamic Content. all fetching of data comes from our API, thus returning the currently saved data from the database. The current post title won't be shown unless the page was already published. Same with #3.

Regarding #3, it was already discussed that when a dynamic content is present, only the field name will be shown inside the text input. If there's more than one, the field names will be separated by a comma (`,`)

blankart:
Need to introduce another API for dynamic content resolvers inside the editor

PR #1341, opened 2021-07-15
Github Actions for JS Unit Tests

blankart:


bfintal:
@blankart I think we need to add a corresponding v3 PR

PR #1356, opened 2021-07-19
Fixes auto block recovery for WP <= 5.6

### LINKED ISSUE #1355: Cannot read property `match` of undefined
When you run e2etests with WP version 5.6, this console error shows:

Screen Shot 2021-07-19 at 10 18 13 AM

Screen Shot 2021-07-19 at 10 18 37 AM

Steps to reproduce:
1. Run e2etest with WP 5.6
2. Run accordion desktop advanced spec

blankart:
Fixes #1355


bfintal (changes requested):
There are other occurrences of this, do we need to do those as well?
Also, we need to create another PR for v3 for this.

PR #1367, opened 2021-07-22
[V3] Icon List Block

blankart:


src/block/icon-list/style.scss
```diff
@@ -1,10 +1,10 @@
...
+	ol {
+		list-style-image: none;
 	}
 
 	// Line height creates unnecessary row gaps.
-	ul > li::marker {
+	::marker {
 		line-height: 0;
 	}
 }
```
bfintal: @blankart we can simplify this further, we shouldn't use `!important` also. We need to be more conscious about the styles which will be loaded in the frontend.

PR #1385, opened 2021-07-28
Icon Label Block

blankart:


bfintal (changes requested):
let's not use columns, let's simplify:
```js
const TEMPLATE = [
	[ 'stackable/icon', {} ],
	[ 'stackable/heading', { text: __( 'Icon Label' ), textTag: 'h4' } ],
]
```
* Add option icon width - defaults at 50px
* Add `p` in heading picker

bfintal (changes requested):
If you add a block, it looks too big. It should more or less look like this:
image

PR #1399, opened 2021-07-29
[V3] Fixes

blankart:


src/styles/editor-block.scss
```diff
@@ -43,6 +43,7 @@
 	margin-right: 0;
 	width: 100%; // Some native blocks have auto widths so they won't follow left/center/right text alignment e.g. buttons
 }
+
```
blankart, line 46:
```suggestion
}
```

src/styles/editor-block.scss
```diff
@@ -43,6 +43,7 @@
...
 // Mobile auto-collapses all columns.
```
blankart, line 47:
```suggestion
	margin-right: 0;
	width: 100%; // Some native blocks have auto widths so they won't follow left/center/right text alignment e.g. buttons
}
// Mobile auto-collapses all columns.
```

bfintal (changes requested):
The fix for #1409 is currently to hide the reset button. This feels a bit off since all our other controls have a reset button and it makes the link control hard to clear out.

The behavior should be to still have the reset button, but just clear the field when reset

PR #1413, opened 2021-07-30
Added `ugb/` blocks support for `get_the_excerpt` API

blankart:
https://developer.wordpress.org/reference/hooks/excerpt_allowed_blocks/

### DIFF
--- a/src/blocks.php
+++ b/src/blocks.php
@@ -134,7 +134,6 @@ if ( ! function_exists( 'stackable_add_excerpt_blocks' ) ) {
 			'ugb/team-member',
 			'ugb/testimonial',
 			'ugb/text',
-			'ugb/video-popup'
 		];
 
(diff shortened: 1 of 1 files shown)


src/blocks.php
```diff
@@ -84,3 +84,61 @@ function stackable_register_blocks() {
 	add_action( 'init', 'stackable_register_blocks' );
 }
 
+/**
+ * Allow our blocks to display post excerpts
+ * when calling `get_the_excerpt` function.
+ *
+ * @see https://developer.wordpress.org/reference/functions/excerpt_remove_blocks/
```
blankart, line 91:
```suggestion
 * @see https://developer.wordpress.org/reference/hooks/excerpt_allowed_blocks/
```

bfintal (changes requested):
This is great!

I don't think we need to include these blocks since they won't really show anything in excerpts and might show unwanted html:
- ugb/icon
- ugb/video-popup

PR #1421, opened 2021-08-02
Debounced text fix

blankart:
When adding a dynamic content as format type (e.g. Other Posts), the span gets escaped.


bfintal (changes requested):
Please add some code comments, I don't quite get it how this fixes the issue

PR #1423, opened 2021-08-03
Feat: Spacer Block

blankart:
#### Changes
- No new API introduced


src/block/spacer/style.js
```diff
@@ -0,0 +1,93 @@
...
+
+const getStyleParams = () => {
+	return [
+		{
+			renderIn: 'save',
+			selector: '',
+			styleRule: 'height',
+			attrName: 'height',
+			format: '%spx',
+		},
```
blankart, line 23:
```suggestion
			attrName: 'height',
			format: '%spx',
			responsive: 'all',			
		},
```

bfintal (changes requested):
When you insert the block, the block has 0 height. It should default to `50px`

bfintal (changes requested):
1. The resizer should only be visible when the block is hovered on
2. When the resizer is dragged and you save the post, refreshing the editor will show an error

bfintal (changes requested):
A few more things that we should add:
- [x] Our image block snaps to specific widths (multiples of 50px) when resizing, we should add this here as well
- [x] When resizing, there is no indicator of the current size of the spacer, it might be a good idea to add a tooltip similar to the image block too that's visible when hovering over the spacer block

bfintal (changes requested):
Some comments:
- [ ] The orange color when dragging looks off, make it transparent
- [ ] When you change the height, if you preview in tablet and mobile, the height goes back to default in the editor (works okay in frontend)
- [ ] The popup height shows the reset button when it's already at the default value. Either the reset button shouldn't be there when the value is the default, or the value should show blank when it's the default to prevent the reset button from showing up

PR #1425, opened 2021-08-03
Feat: Icon Button Block

blankart:
#### Changes
- New: Icon Button Block
- New: Added Row and Column Gap options in `stackable/button-group`
- New: Added `.stk-button` class in `button` block component.
- Moved button group related styles from `block/button/style.scss` to `block/button-group/style.scss`


src/block/button/style.scss
```diff
@@ -3,105 +3,43 @@
...
-.stk-block-button-group .has-text-align-left {
-	justify-content: flex-start;
-}
 
 .stk-block-button.is-style-ghost {
-	.stk-block-button__button {
+	.stk-button {
 		&::before {
-			border-width: cssvar(button-ghost-border-width);
+			border-width: 2px;
```
blankart, line 12:
```suggestion
			border-width: cssvar(button-ghost-border-width);
```

bfintal (changes requested):
- [x] Move the icon color to the Button Colors panel
- [x] Move the Icon panel immediately after the Button Colors panel
- [x] Also add "Button Width" below "Min. Button Height"
- [x] Add the Styles panel similar to the Button block

bfintal (changes requested):
- [x] I noticed that there are a lot of repeated styles in the `style.scss` of the button block and the icon button block, these should be shared to lessen the amount of styles we have
- [x] The button inner span previously had the class `stk-button__inner-text` which now became `stk-block-button__inner-text`. We should turn it back since the button text has lost some of its styling. The button class names should be shared between the button block and the icon button block to simplify things.

src/styles/block.scss
```diff
@@ -198,6 +198,122 @@
...
+			.stk--inner-svg {
+				color: cssvar(button-background-color);
+				&,
+				:is(g, path, rect, polygon, ellipse) {
+					fill: cssvar(button-background-color);
+				}
+			}
+		}
+	}
+}
```
bfintal, line 310: Please move all `cssvar` and design system related styles in `block-design-system.scss`

src/block/icon-button/style.scss
```diff
@@ -0,0 +1,18 @@
...
+
+// We have a block group, and that has the bottom margin.
+.stk-block-icon-button {
+	--stk-block-margin-bottom: 0;
+}
+
+.stk-block.stk-block-icon-button {
+	width: auto;
+	margin: 0;
+	text-align: center;
```
bfintal, line 11: these came from the button block, are these styles needed here?

src/block/icon-button/style.scss
```diff
@@ -0,0 +1,18 @@
...
+
+	a.stk-button {
+		// Make the buttons equal in height and width by default.
+		padding: 12px;
+		border-radius: 24px;
```
bfintal, line 16: We can make this `100%` so that the block remains circular when adjusting the icon size
```suggestion
		border-radius: 100%;
```

src/block/button/style.scss
```diff
@@ -1,110 +1,12 @@
...
+.stk-block-button {
+	--stk-block-margin-bottom: 0;
+}
+
 .stk-block.stk-block-button {
 	width: auto;
-}
-.stk-block-button {
 	margin: 0;
 	text-align: center;
```
bfintal, line 11: I checked this, if I remove them they don't seem to do anything. Do we still need these styles?

PR #1429, opened 2021-08-04
Feat: Subtitle Block

blankart:


bfintal (changes requested):
- [x] Simplify CSS selectors
- [x] Add description in block.json
- [x] Remove `subtitle` from the `keywords` in block.json since that's automatically used as a keyword when searching blocks in Gutenberg
- [ ] Add the slash shortcut to subtitle blocks
- [ ] Add the text placeholder "Type / to choose a block"
- [ ] Instead of adding a large help text in the global typography area, add a small "?" icon, then when you hover on it, show the help tooltip similar to how it looks like when you hover over the linking icon

image

PR #1431, opened 2021-08-04
Feat: Link added Sponsored and UGC options

blankart:
### V3 Implementation

Creating a rel `TextControl` would require deprecation of other attributes (e.g.`NoFollow` which is a boolean attribute ). So i added more attributes ( `Sponsored`, `Ugc` ) instead to avoid deprecation.

### DIFF
--- a/src/components/button-controls/index.js
+++ b/src/components/button-controls/index.js
@@ -70,14 +70,14 @@ const ButtonControls = props => {
 			{ props.onChangeUrl && props.onChangeSponsored && (
 				<AdvancedToggleControl
 					label={ __( 'Sponsored', i18n ) }
-					checked={ props.noFollow }
+					checked={ props.sponsored }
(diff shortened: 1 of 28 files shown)


src/components/button-controls/index.js
```diff
@@ -67,6 +67,20 @@ const ButtonControls = props => {
 					onChange={ props.onChangeNoFollow }
 				/>
 			) }
+			{ props.onChangeUrl && props.onChangeSponsored && (
+				<AdvancedToggleControl
+					label={ __( 'Sponsored', i18n ) }
+					checked={ props.noFollow }
```
bfintal, line 73:
```suggestion
					checked={ props.sponsored }
```

bfintal (changes requested):
@blankart  sponsored and ugb doesn't work in the button block

PR #1438, opened 2021-08-06
Added Query Support for Unit Blocks

blankart:


src/block-components/context/index.php
```diff
@@ -0,0 +1,54 @@
...
+		preg_match_all( '/!#STK_CONTEXT_(.*?)!#/', $content, $matches );
+		return $matches[ 1 ];
+	}
+
+	/**
+	 * Utility function to check
+	 * whether the block has !#STK_CONTEXT_.
+	 *
+	 * @param string $content
+	 * $param array $block
```
blankart, line 26:
```suggestion
	 * @param array $block
```

bfintal:
I'll close this. We've decided to go with supporting Query Loop support via Dynamic Content

PR #1459, opened 2021-08-17
Feat: Separators

blankart:


bfintal:
- [ ] When first adding a separator, you can't see it
- [ ] When using the shadow slider, sometimes there is no shadow visible
- [ ] Using the flip horizontally option flips the separator vertically

PR #1467, opened 2021-08-23
Fix: Video URL should only display link dynamic fields

blankart:


bfintal:
What if the field type is text (e.g. in ACF) maybe this is okay to be any

bfintal:
We'll allow normal field types to be used

PR #1471, opened 2021-08-23
Fix: When `props.unit` is provided, use it instead in AdvancedRangeControl

blankart:


src/components/advanced-range-control/index.js
```diff
@@ -30,7 +30,11 @@ const AdvancedRangeControl = props => {
 	const { clientId } = useBlockEditContext()
 	const attributes = useBlockAttributes( clientId )
 
-	const unit = attributes ? attributes[ unitAttrName ] : ''
+	const unit = typeof props.unit === 'string'
+		? ( props.unit || props.units?.[ 0 ] || 'px' )
```
bfintal, line 34: I don't think this `props.units?.[ 0 ]` will ever be used here
```
props.unit || props.units?.[ 0 ] || 'px'
```

bfintal (changes requested):
I don't think `props.units?.[ 0 ]` will ever be used here

PR #1518, opened 2021-09-07
Fix: Fixes block errors when stackable blocks are inside the query loop block

blankart:


bfintal:
This doesn't work correctly. `clearSelectedBlock` is meant to be triggered when nothing is selected. This fix constantly triggers changes where it should only trigger once

PR #1530, opened 2021-09-08
Feat: Pressing backspace should remove the block

blankart:


bfintal (changes requested):
- [ ] You should be able to merge together different blocks together (heading & text, or subtitle and text)
- [ ] Try whether this can also merge with core blocks (e.g. Stackable text w/ native text)

PR #1557, opened 2021-09-10
Fix: Added WP 5.5 compatibility to block adder

blankart:


bfintal (changes requested):
Instead of always implementing our own `createBlocksFromInnerBlocksTemplate`, we should only implement it if the function isn't available - like a polyfill.

PR #1568, opened 2021-09-13
Fix: Added cursor copy when hovering over icons

blankart:


src/block/icon-list/editor.scss
```diff
@@ -3,3 +3,20 @@
 	margin-top: 0;
 	margin-bottom: 0;
 }
+
+.stk-block-icon-list {
```
bfintal, line 7:
```suggestion
// Show a cursor pointer when hovering over the bullet icons
.stk-block-icon-list {
```

PR #1592, opened 2021-09-15
Fix: Separator overflows in tablet and mobile

blankart:


src/deprecated/v2/block/separator/style.scss
```diff
@@ -54,6 +54,7 @@
 		position: relative;
 		height: 100%;
 		width: 100%;
+		overflow: hidden;
```
bfintal, line 57:
```suggestion
		overflow-x: hidden;
```

PR #1599, opened 2021-09-15
Fix: fixes automatic clamping of values in WP 5.5

blankart:


bfintal (changes requested):
In the new refactored code, `setValue( props.value )` will never be called. Is this intended in this change?

blankart:
> In the new refactored code, `setValue( props.value )` will never be called. Is this intended in this change?

Yes

plugin.php
```diff
@@ -6,7 +6,7 @@
  * Author: Gambit Technologies, Inc
  * Author URI: http://gambit.ph
  * Text Domain: stackable-ultimate-gutenberg-blocks
- * Version: 3.0.0-beta3
+ * Version: 3.0.0-beta3-range-control-wp55-fix
```
blankart, line 9:
```suggestion
 * Version: 3.0.0-beta3
```

plugin.php
```diff
@@ -23,7 +23,7 @@
 }
 
 defined( 'STACKABLE_SHOW_PRO_NOTICES' ) || define( 'STACKABLE_SHOW_PRO_NOTICES', true );
-defined( 'STACKABLE_VERSION' ) || define( 'STACKABLE_VERSION', '3.0.0-beta3' );
+defined( 'STACKABLE_VERSION' ) || define( 'STACKABLE_VERSION', '3.0.0-beta3-range-control-wp55-fix' );
```
blankart, line 26:
```suggestion
defined( 'STACKABLE_VERSION' ) || define( 'STACKABLE_VERSION', '3.0.0-beta3' );
```

src/components/advanced-range-control/range-control.js
```diff
@@ -45,14 +45,24 @@ const StackableRangeControl = memo( props => {
...
 	// Update the internal value state if the prop changes.
 	useEffect( () => {
-		setValue( props.value === '' || ( isNaN( props.value ) && props.value !== 'auto' ) ? '' : props.value )
+		// Invalid values entered inside the number control will be omitted.
+		if ( props.value === '' || ( isNaN( props.value ) && props.value !== 'auto' ) ) {
+			setValue( '' )
+			// Only allow updating of the local state if the number control is not in focus.
+			// We do this so the clamping of values will not get triggered as long as the
+			// user is typing inside the number control. This happens in WP 5.5
+		} else if ( numberControlRef.current !== document.activeElement ) { // eslint-disable-line
```
bfintal, line 63:
```suggestion
		// Only allow updating of the local state if the number control is not in focus.
		// We do this so the clamping of values will not get triggered as long as the
		// user is typing inside the number control. This happens in WP 5.5
		} else if ( numberControlRef.current !== document.activeElement ) { // eslint-disable-line
```

PR #1602, opened 2021-09-16
Feat: More Card Styles

blankart:


src/block/card/variations.js
```diff
@@ -0,0 +1,68 @@
...
+ * @type {WPBlockVariation[]}
+ */
+const variations = applyFilters(
+	'stackable.card.variations',
+	[
+		{
+			name: 'default',
+			title: __( 'Default', i18n ),
+			isDefault: true,
+			description: __( 'Default Layout', i18n ),
```
blankart, line 29:
```suggestion
			name: 'basic',
			title: __( 'Basic', i18n ),
			isDefault: true,
			description: __( 'Basic Layout', i18n ),
```

bfintal (changes requested):
@blankart before I merge this PR, please apply the changes here to the other blocks and designs. Here are the changes:
- [ ] Apply the translation comments in the attributes of the layouts:
```js
{ text: _x( 'Subtitle for This Block', 'Subtitle placeholder', i18n ) }
```
- [ ] Removed the hardcoded styles from the `editor.scss` and `style.scss` as much as possible and move them to `attributes`:
```js
attributes: {
	className: 'is-style-faded',
	blockHeight: 600,
	imageFilter: 'opacity(0.3)',
},
```
- [ ] Place the icons in Stackable free so that they will be visible in the free version.
- [ ] Instead of overwriting the designs using the filter (and thereby repeating some properties such as the design's name, description, etc), place those information in the free plugin. The premium version should just add / override information in the designs. Here's the code that merges the information from premium:
```js
// Merge the variations
const newVariations = variations.filter( ( { name } ) => ! premiumVariations.find( v => v.name === name ) )
const updatedVariations = variations.map( variation => {
	const newData = premiumVariations.find( v => v.name === variation.name ) || {}
	return {
		...variation,
		...newData,
	}
} )
```
- [ ] To lessen the amount of translatable strings, use `sprintf` to suffix the variation name with "Layout":
```js
description: sprintf( _x( '%s Layout', 'Block layout name', i18n ), __( 'Horizontal', i18n ) ),
```

PR #1606, opened 2021-09-16
Feat: More button group features and styles

blankart:


src/block-components/button/edit.js
```diff
@@ -150,14 +151,20 @@ export const Size = props => {
 				title={ __( 'Button Size & Spacing', i18n ) }
 				id="button"
 			>
+				{ props.hasFullWidth && (
+					<AdvancedToggleControl
+						label={ __( 'Full Width' ) }
```
bfintal, line 156:
```suggestion
						label={ __( 'Full Width', i18n ) }
```

PR #1613, opened 2021-09-17
Feat: Blockquote Layouts

blankart:


bfintal (changes requested):
- [ ] Add the blockquote premium styles in the variation list of free. But these shouldn't be clickable and should be styled differently (just do `opacity: 0.5` for now for the premium ones when in free)
- [x] Premium variations should override the free ones if `isPro`

PR #1620, opened 2021-09-20
Feat: Call to Action Styles

blankart:
### Call to Action Styles

- [x] Default Style
- [x] Plain
- [x] Horizontal Style (premium but can be achieved using free features)
- [x] Horizontal 2 (premium but can be achieved using free features)
- [x] Horizontal 3 (premium but can be achieved using free features)
- [x] Split Centered (premium but can be achieved using free features)


src/block/call-to-action/variations.js
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @type {WPBlockVariation[]}
+ */
+const variations = applyFilters(
+	'stackable.call-to-action.variations',
+	[
+		{
+			name: 'default',
+			title: __( 'Default', i18n ),
+			description: __( 'Default Layout', i18n ),
```
blankart, line 23:
```suggestion
			name: 'basic',
			title: __( 'Basic', i18n ),
			description: __( 'Basic Layout', i18n ),
```

bfintal (changes requested):
- [x] In the horizontal layout, there's no way to center vertically the blocks when the inner blocks orientation is set to horizontal. Maybe we can enable the option in the block tab > alignment panel.
- [ ] Apply the changes from https://github.com/gambitph/Stackable/pull/1602#pullrequestreview-766952578

PR #1662, opened 2021-09-28
Fix: Enqueue global settings inline styles if there are no stackable …

blankart:
…blocks


src/deprecated/v2/init.php
```diff
@@ -136,6 +136,12 @@ function stackable_block_assets_v2() {
 
 		// Frontend only scripts.
 		if ( ! is_admin() ) {
+			// Add global colors.
```
blankart, line 139:
```suggestion
			// Add global colors and typography.
```

PR #1670, opened 2021-09-28
Fix: Change the `sliderMax` based on the Image's width and height value

blankart:


bfintal (changes requested):
The max should change depending on the width or height, whichever is higher

blankart:
> The max should change depending on the width or height, whichever is higher

There are units in both height and width. Border radius is strictly px

bfintal:
> There are units in both height and width. Border radius is strictly px

We can use the max value of the width and height whichever is in `px`. Then is both are not `px`, then use `100` instead.

src/block-components/image/edit.js
```diff
@@ -41,6 +42,25 @@ const Controls = props => {
...
+
+		if ( attributes.imageWidthUnit === 'px' ) {
+			return attributes.imageWidth
+		}
+
+		if ( [ '', 'px' ].includes( attributes.imageHeightUnit ) ) {
+			return ( ( attributes.imageHeight || 100 ) / 2 )
+		}
+
+		return 50
```
bfintal, line 61:
```suggestion
		return 100
```

src/block-components/image/edit.js
```diff
@@ -41,6 +42,25 @@ const Controls = props => {
 		return { imageData: { ...image } }
 	}, [ attributes.imageId ] )
 
+	const borderRadiusSliderMax = useMemo( () => {
+		if (
+			attributes.imageWidthUnit === 'px' &&
```
bfintal, line 47:
```suggestion
			[ '', 'px' ].includes( attributes.imageWidthUnit ) &&
```

src/block-components/image/edit.js
```diff
@@ -41,6 +42,25 @@ const Controls = props => {
 		return { imageData: { ...image } }
 	}, [ attributes.imageId ] )
 
+	const borderRadiusSliderMax = useMemo( () => {
+		if (
+			[ '', 'px' ].includes( attributes.imageWidthUnit ) &&
+			( [ '', 'px' ].includes( attributes.imageHeightUnit ) )
```
bfintal, line 48:
```suggestion
			[ '', 'px' ].includes( attributes.imageHeightUnit )
```

src/block-components/image/edit.js
```diff
@@ -41,6 +42,25 @@ const Controls = props => {
 		return { imageData: { ...image } }
 	}, [ attributes.imageId ] )
 
+	const borderRadiusSliderMax = useMemo( () => {
+		if (
+			[ '', 'px' ].includes( attributes.imageWidthUnit ) &&
```
bfintal, line 47:
```suggestion
			attributes.imageWidthUnit === 'px' &&
```

PR #1695, opened 2021-09-30
[V3] Fixes

blankart:


bfintal (changes requested):
- [ ] When setting the Feature grid wide/full, the content alignment should also adjust
- [ ] Feature block, the Style tab > General panel should be the first

PR #1935, opened 2021-10-18
[v3.0.3 Candidate] Feat: Image Overlays

blankart:


legaspielaine:
Gradient color option is missing:

v2 image block has this option (users may request for this as well)
Screen Shot 2021-12-03 at 5 22 25 PM

bfintal:
Another issue that's fixed also in this PR: some hover transition speed/durations cannot be adjusted - now they can be adjusted in the adv > transition panel.

PR #2006, opened 2021-11-10
Fix: Fixes icon block custom css selector in backend

blankart:
Connected PR:https://github.com/bfintal/Stackable-Premium/pull/229


bfintal (changes requested):
@blankart Let's change fix, adding a boolean `appendSpaceAfterUniqueClass` prop feels like a hacky fix.

bfintal:
Alternative fix: https://github.com/gambitph/Stackable/commit/89ff1e2787f67fc443864afcec40f35a6da72f93

PR #2015, opened 2021-11-11
Feat: Block alignment support for heading block

blankart:


src/block/heading/index.js
```diff
@@ -23,6 +23,7 @@ export const settings = applyFilters( 'stackable.block.metadata', {
 	attributes: schema,
 	supports: {
 		anchor: true,
+		align: true,
```
blankart, line 26:
```suggestion
		align: [ 'center', 'wide', 'full' ],
```

legaspielaine:
### Issues

- [ ] When setting Heading to `Align center`, the alignment in Backend not the same in Frontend > [link](https://github.com/gambitph/Stackable/issues/2016)

**Steps to reproduce the behavior:**
1. Add Heading block
2. set this to Align center in its toolbar
3. compare the alignment in Editor vs Frontend
4. See bug - in Frontend, the Heading is left aligned but in Backend it's center aligned

**Expected behavior**
The frontend should be the same as the one in backend (center aligned).

**Screenshots**

https://user-images.githubusercontent.com/28699204/141259305-cba06010-6ff0-4b5b-84d8-166ee66255d8.mov

legaspielaine:
this [issue](https://github.com/gambitph/Stackable/pull/2015#issuecomment-966143759) is still happening

src/styles/editor-block.scss
```diff
@@ -310,4 +310,3 @@
 	align-items: normal;
 	justify-content: var(--stk-alignment-justify-content);
 }
```
bfintal, line 312:
```suggestion
}

```

PR #2045, opened 2021-12-02
fix: column gap inherited from desktop collapses tablet columns

bfintal:
How to replicate:
1. create 2 columns
2. add column gap 100px
3. change the column width to 40% & 60%
4. preview in frontend, then resize browser to tablet size. See that the columns will collapse


legaspielaine:
> How to replicate:
> 
> 1. create 2 columns
> 2. add column gap 100px
> 3. preview in frontend, then resize browser to tablet size. See that the columns will collapse

I cannot replicate this issue with our released 3.0.6

bfintal:
Updated the replication steps. it can now be replicated properly

PR #2047, opened 2021-12-02
fix: removes the delay when the icon changes color when hovering.

### LINKED ISSUE #2046: There is a delay when an icon changes colors on hover
https://user-images.githubusercontent.com/1033611/144424870-6728e213-c9bf-4e48-b9ad-dbda7ad2e54e.mov

Steps to replicate:
1. Add a column with an icon and text
2. On hover, change the column background color
3. On hover parent, change the color of the icon and text
4. Preview in the frontend, hover mouse over the column. Notice a slight delay in the icon's color change.

bfintal:
fixes #2046


legaspielaine:
The issue is still happening in Backend but in Frontend, it's ok

PR #2051, opened 2021-12-03
feat: added an orange indicator to visually see when a responsive or hover state has a value

bfintal:


bfintal:
Closing in favor of https://github.com/gambitph/Stackable/pull/2649

PR #2077, opened 2021-12-29
Feat: Added image link

bfintal:


legaspielaine:
Tested the build: We might need to disable the Link panel on Image if it's inside Image box and Video popup blocks @bfintal

bfintal:
@legaspielaine great catch. I just added a fix to remove the image link when the block parent is the video popup block or the image box block

PR #2106, opened 2022-01-13
Add preventScroll to focus.

### LINKED ISSUE #2060: Clicking the Expand's show more button jumps to the bottom of the text
**To Reproduce**
1. Add an Expand block (populate the less text & more text - longer text)
2. In frontend, click the show more button.
3. See bug

**Expected behavior**
Clicking the button should not jump to the bottom of the text - this can be annoying.

**Screenshots**

https://user-images.githubusercontent.com/28699204/145511296-6d0d8969-ce31-406b-ba02-673121b14254.mov

gumacahin:
Fixes #2060

### DIFF
--- a/src/block/expand/frontend-expand.js
+++ b/src/block/expand/frontend-expand.js
@@ -18,7 +18,9 @@ class StackableExpand {
 			visibles.forEach( el => el.setAttribute( 'aria-hidden', 'true' ) )
 
 			// Refocus on the hide/show button.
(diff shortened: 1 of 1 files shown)


bfintal:
preventScroll has bug in Safari & iOS Safari that's preventing it from working correctly: https://bugs.webkit.org/show_bug.cgi?id=178583

I think for now let's use this fix and hope that eventually they will resolve the Safari bug

Issue #2093, opened 2022-01-11
Stackable colors do not have margins. 


prconcepcion:
![image](https://user-images.githubusercontent.com/49634311/148881497-5e8b8bdd-5b52-4867-8db5-757caa5c77a0.png)

1. Go to Stackable Settings.
2. Go to Global Color Palette.
3. Add a color.

PR #2112, opened 2022-01-13
Fixed misaligned stackable global colors.

prconcepcion:
Fixes issue #2093 
Works with 5.8

### DIFF
--- a/src/plugins/global-settings/colors/editor.scss
+++ b/src/plugins/global-settings/colors/editor.scss
@@ -19,7 +19,7 @@
 }
 
 .ugb-global-settings-color-picker__add-icon {
-	margin: 5px 5px 0px 5px;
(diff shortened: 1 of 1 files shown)


src/plugins/global-settings/colors/editor.scss
```diff
@@ -61,3 +62,6 @@
 	}
 }
 
+.components-circular-option-picker__option-wrapper {
```
bfintal, line 65: This selector might also affect other non-stackable color options. Add a parent selector to ensure only Stackable controls will be affected.

bfintal (changes requested):
Please see comment on the selector

bfintal (changes requested):
Please use `0 12px 12px 0` margins

bfintal (approved):

Issue #2022, opened 2021-11-15
Add: List gap option in Icon List


legaspielaine:
Let's add List/row gap option to Icon List. in version 2, we have this option.

PR #2147, opened 2022-01-24
Add slider for list gap.

gumacahin:
Fixes https://github.com/gambitph/Stackable/issues/2022.


bfintal (changes requested):
The row gap value doesn't stick. After adjusting the row gap, save the post and refresh the editor. You'll see that the row gap resets to the default.

You may need to add `rowGap` to the list of attributes in `schema.js`, be sure that it works in responsive attributes also.

Issue #2143, opened 2022-01-22
Button plus icon is in the wrong position


bfintal:
![image](https://user-images.githubusercontent.com/1033611/150640178-0dde9089-351e-4b00-8ec9-b5609a11c1e9.png)

Steps:
1. Add a button block (WP 5.9)
2. See that the plus button is overlapping the button block

PR #2149, opened 2022-01-25
Puts the button plus icon in the bottom right corner.

prconcepcion:
Fix for issues #2143.


prconcepcion:
One way to make it inline with the buttons is to not make the position absolute. 
image

Issue #2134, opened 2022-01-19
Column resize handles not visible in tablet & mobile preview in Firefox


bfintal:
https://user-images.githubusercontent.com/1033611/150059622-a21d0f7a-2d02-4d5c-9a74-adece739c988.mov

Steps to replicate:
1. In Firefox, add a columns block
2. Preview in tablet inside the editor, see that the resize handlers are not present

Note: there's a small box that shows up in tablet/mobile when hovering over the column, this might be just a CSS issue

bfintal:
Fixed by #2150 and fix/performance branch

PR #2150, opened 2022-01-26
Restore missing column resize handle when previewing tablet/mobile on Firefox.

gumacahin:
Fixes https://github.com/gambitph/Stackable/issues/2134.


bfintal:
Seems like we've already implemented this in `fix/performance`

Issue #2009, opened 2021-11-10
Accordion wont open on older version of browsers


legaspielaine:
**Describe the bug**
Our v3 accordion is using the browser when opening & closing the block. In older versions of browsers, this isn't supported. We need to add backwards compatibility for our accordion.

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Safari
 - Browser Version (not working): 13.1.2
 - Browser Version (working): 14.1.2

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy J5
 - OS: Android 7.0
 - Browser: Chrome
 - Browser Version (not working): 95.0.4638.50

bfintal:
Technical notes:
- Our accordion block relies on the browser's support for the `details` HTML tag. https://caniuse.com/?search=details
- The `details` feature isn't supported in older browsers
- Ideal fix: if an older browser is detected (or if support for `details` is not present), then the opening and closing of the accordion should be handled by a frontend JS script. It's important that for browsers that do support the `details` tag, the default browser behavior should be used.

PR #2155, opened 2022-01-30
Accordion wont open in old browser

gumacahin:
Fixes https://github.com/gambitph/Stackable/issues/2009.


src/block/accordion/index.php
```diff
@@ -18,3 +18,35 @@ function stackable_load_accordion_frontend_script() {
 	}
 	add_action( 'stackable/accordion/enqueue_scripts', 'stackable_load_accordion_frontend_script' );
 }
+
+if ( ! function_exists( 'stackable_load_accordion_frontend_polyfill_script' ) ) {
+	require_once( ABSPATH . 'wp-admin/includes/dashboard.php' );
```
bfintal, line 23: What is this `require_once` for? If this is needed, maybe we should add this inside the area / condition where this is used. With this current location, this `require_once` is always present per page load.

src/block/accordion/index.php
```diff
@@ -18,3 +18,35 @@ function stackable_load_accordion_frontend_script() {
...
+	/**
+	 * Adds polyfill for summary/details element that are * used in accordion blocks.
+	 *
+	 * TODO: confirm that this works on older browsers
+	 */
+	function stackable_load_accordion_frontend_polyfill_script() {
+
+		$browser_version = wp_check_browser_version();
```
bfintal, line 31: Let's not use `wp_check_browser_version`, this uses a lot more resources than necessary and also pings a remote URL.

Instead, just check `$_SERVER['HTTP_USER_AGENT']`. Be sure to check `empty( $_SERVER['HTTP_USER_AGENT'] )`

bfintal (changes requested):
Please see my comments about `wp_check_browser_version`

src/block/accordion/index.php
```diff
@@ -18,3 +18,43 @@ function stackable_load_accordion_frontend_script() {
...
+
+if ( ! function_exists( 'stackable_load_accordion_frontend_polyfill_script' ) ) {
+	/**
+	 * Adds polyfill for summary/details element that are * used in accordion blocks.
+	 *
+	 * TODO: confirm that this works on older browsers
+	 */
+	function stackable_load_accordion_frontend_polyfill_script() {
+
+		 $user_agent = ! empty($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : '';
```
bfintal, line 30:
```suggestion
		 $user_agent = ! empty( $_SERVER['HTTP_USER_AGENT'] ) ? $_SERVER['HTTP_USER_AGENT'] : '';
```

src/block/accordion/index.php
```diff
@@ -18,3 +18,43 @@ function stackable_load_accordion_frontend_script() {
...
+		 if ( ! $user_agent ) {
+			 return;
+		 }
+
+		 $matches = array();
+
+		 if ( preg_match('/(Edge|Chrome|Safari)\/(\d+)/', $user_agent, $matches) ) {
+			$name    = $matches[1];
+			$version = intval($matches[2]);
+		 }
```
bfintal, line 41:
```suggestion
		 if ( preg_match( '/(Edge|Chrome|Safari)\/(\d+)/', $user_agent, $matches ) ) {
			$name    = $matches[1];
			$version = intval( $matches[2] );
		 }
```

src/block/accordion/index.php
```diff
@@ -18,3 +18,43 @@ function stackable_load_accordion_frontend_script() {
...
+
+		if ( isset( $name ) && isset( $version) ) {
```
bfintal, line 43:
```suggestion
		if ( isset( $name ) && isset( $version ) ) {
```

bfintal (changes requested):
I added some minor spacing comments for the PHP part.

I tested the polyfill just to check what would happen. I commented out the first part to force it to run:

```js
//	if ( 'open' in document.createElement( 'details' ) ) {
//		return
//	}
```

I encountered a bunch of errors and the script didn't run:
1. `document.querySelectAll` is undefined
2. can't create element `document.createElement( '.stk-block-accordion .stk-block-accordion__heading' )`
3. `style` is referenced before initialization at `style.innerText === undefined ? 'textContent' : 'innerText',`

I suggest trying out the polyfill code to see whether it runs or not. While we can't test fully using an actual browser that doesn't support the `details` tag, trying to get it to run might be enough.

Issue #2151, opened 2022-01-27
Rem typography units not inherited in tablet/mobile


bfintal:
Steps to replicate:
1. Add a text block
2. Set font size to `2rem` in desktop
3. Preview in frontend. See the font-size goes to `2px` in tablet/mobile size in the frontend.

Note: Will need to check this on the typography module since other blocks might be affected by this issue

PR #2157, opened 2022-01-31
Fix/2151 font size unit not inherited

gumacahin:
Fixes https://github.com/gambitph/Stackable/issues/2151.


src/block-components/typography/attributes.js
```diff
@@ -3,6 +3,11 @@ const typographyAttributes = {
 		stkResponsive: true,
 		type: 'number',
 		default: '',
+		/**
+		 * Setting a unit here makes it the default value for all device types.
+		 * By leaving this empty the responsive settings will be correctly
+		 * inherited in the order: desktop -> tablet -> mobile
+		 */
```
bfintal, line 10:
```suggestion
```

Issue #2144, opened 2022-01-23
Column resize indicators show incorrectly


bfintal:
https://user-images.githubusercontent.com/1033611/150660171-4942fcef-da71-433e-9700-c78b735074ed.mov

Steps:
1. Open the Design Library and pick the Wireframes Blog Posts 1 design
2. Hover your mouse over the blog posts, see that the indicators are always shown

PR #2171, opened 2022-02-12
Add styles to correct display of resize handles.

gumacahin:
Fixes https://github.com/gambitph/Stackable/issues/2144.

### DIFF
--- a/src/block/column/editor.scss
+++ b/src/block/column/editor.scss
@@ -8,12 +8,12 @@
 	width: 100%;
 }
 
+// Ensure that only directly hovered columns will show their resizer. Nested
+// columns should not show their resizer if they are not directly hovered.
 .components-resizable-box__container.has-show-handle {
-	.components-resizable-box__handle {
-		display: none;
+	.components-resizable-box__container:not(.has-show-handle) {
+		.components-resizable-box__handle {
+			display: none !important;
+		}
 	}
-}
-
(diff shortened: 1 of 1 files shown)


src/block/column/editor.scss
```diff
@@ -7,3 +7,13 @@
...
+
+.components-resizable-box__container.has-show-handle {
+	.components-resizable-box__handle {
+		display: none;
+	}
+}
+
+.components-resizable-box__container.has-show-handle > div > div> div > div > div[data-type="stackable/columns"] > div > div > div > div > div[data-type="stackable/column"] > div > div > div > div.components-resizable-box__side-handle {
+	display: block;
+}
```
bfintal, line 19: This can easily break if something in the structure changes. Also, our implementation is already displaying the handlers, we just need to hide them for nested columns.

src/block/column/editor.scss
```diff
@@ -7,3 +7,13 @@
 .editor-styles-wrapper .stk-column-wrapper > .block-editor-inner-blocks {
 	width: 100%;
 }
+
+.components-resizable-box__container.has-show-handle {
+	.components-resizable-box__handle {
+		display: none;
+	}
+}
```
bfintal, line 15: Won't this hide all handlers?

bfintal:
@gumacahin I updated the implementation so that we force hide resize handlers of deeply nested columns

PR #2178, opened 2022-02-17
fix (variation picker): attributes not present in free version

bfintal:


gumacahin (commented):
Remove extra space.

gumacahin (commented):
Checked. Looks good to me.

PR #2181, opened 2022-02-21
Feat/2146 table of contents block

gumacahin:
WIP towards fixing: https://github.com/gambitph/Stackable/issues/2146

# Outstanding ToC issues

The table of contents block is an auto-generated table of contents section. It is still a work in progress.

# Quick Technical Notes

The Table of Contents Block (ToC) fetches all the `Core Heading` blocks and `Stackable Heading` blocks from the (other “[adv.] heading” blocks are unsupported).

The visibility of the `h1-h6` heading elements is controlled by individual toggle switches in the settings pane. Toggling these should toggle the visibility of the corresponding headings displayed on the ToC block.

When the ToC block is selected visibility toggle buttons will appear at the end of each of the items of the ToC block. These manage the visibility of the list item for the heading on the ToC block. When the ToC block is deselected, the vis. icons are hidden and list entries that have the visibility setting turned off will be hidden as well. The unselected view of the ToC block is the preview of the ToC block on the front-end.

All other functionality/settings are similar adapted from the `Icon List Block`, which means all outstanding issues from the `Icon List Block` are apparent on the `ToC Block`.  (see below)

# FAQ

Why is there no “Table of Contents” heading?

A: User can manually add their own heading with a heading block. This will appear on the ToC but they can opt to hide it.

# Priority Outstanding Issues

- [x]  Typography settings should apply to ToC preview.
- [x]  Should have a `Smoooth Scroll/Jump To Heading` setting.
- [x]  <del>Links on the ToC Block should take user to corresponding heading on the editor as a navigation aid. `Scroll setting` should apply.</del><ins>We are no longer doing this because clicking on the text should enable editing.</ins>
- [x]  When the block is selected all headings on the editor should be accounted for. Except those excluded by the `h1-h6` setting. Each line item should have a visual indication that they are hidden along with an **eye icon** to toggle the visibility. Hidden items are hidden when the ToC block is deselected. Eat breakfast
- [x]  Should allow the use to type over the auto generated item text.
- [x] Table of Contents Icon
- [x] Example should look like a ToC area

# Outstanding Icon List Issues

- [x]  Using large list icons causes the icons to become off-center vertically with the text.
- [x]  List Alignment not working as expected on the editor/front end.
- [ ]  Content overflow clipping when there are more than one column.


bfintal (changes requested):
I feel that we are getting overly complex with the table of contents block. Can we simply things a bit?

1. No need for a separate `notice.js` file just to display the note about auto generating ids. I feel that we can just include that inside `edit.js` and call it  day.
2. No need to use `__experimentalText` just to display text, I removed that already, just using normal text works okay
3. I don't think we need a new script file just for the smooth scroll, we can just generate the following style:
```
html {
    scroll-behavior: smooth;
}
```
4. Let's add an option "Scroll top offset" that's a slider, 0 - 200 which will output the style below. This is helpful if there's a floating header on the way.
```
html {
    scroll-padding-top: 100px;
}
```
5. The `generateAnchor` method can be simplified. Just use `kebabCase( content )` and then if it's been encountered before (e.g. check the anchors in the heading attribute, then just suffix `-2` or `-3` to make it unique.  In addition to this, add a filter call to generate this anchor, similar to the filter `stackable.block.table-of-contents.convert-block` so other plugins can add their own blocks.
6. Since we're not using icons anymore, remove references to icons such as the function `convertSVGStringToBase64`  or icons in the style generation
7. Let's not use `classNames` if we won't be styling or querying that element, e.g. we can remove the className `stk-table-of-contents__placeholder` since we won't be referencing this
8. Columns includes the notice in the layout. Recommend to place the notice outside the main `BlockDiv` so it won't interfere with normal block behavior

src/util/index.js
```diff
@@ -306,7 +306,10 @@ export const prependCSSClass = ( cssSelector, mainClassName = '', uniqueClassNam
 		.split( ',' )
 		.map( s => {
 			let newSelector = ''
-			if ( s.includes( '[data-block=' ) ) {
+			if ( s.includes( '[data-block=' ) ||
+					s === 'html' ||
+					s === 'body'
+			) {
```
bfintal, line 312:
```suggestion
			if ( s.includes( '[data-block=' ) ||
			     s === 'html' ||
			     s === 'body'
			) {
```

src/block/table-of-contents/util.js
```diff
@@ -0,0 +1,318 @@
...
+		// const allowedHeadingBlocks = [ ...headingBlocks ].filter( block => {
+		// 	// Find the heading DOM element
+		// 	const headingEl = ! block.classList.contains( 'wp-block-heading' )
+		// 		? block.querySelector( '.stk-block-heading__text' )
+		// 		: block
+
+		// 	return allowedHeadings.includes( headingEl.nodeName )
+		// } )
+
+		// return getHeadingsFromHeadingBlocks( allowedHeadingBlocks, attributes.headings )
```
bfintal, line 245:
```suggestion
```

src/block/table-of-contents/block.json
```diff
@@ -0,0 +1,15 @@
+{
+	"name": "stackable/table-of-contents",
+	"title": "Table of Contents",
+	"description": "Automatically generated table of contents. Supports native heading blocks and Stackable heading blocks.",
+	"category": "common",
+	"keywords": [
+		"ToC",
+		"Table of Contents",
```
bfintal, line 8:
```suggestion
```

bfintal:
Changes:
- ToC is a "special" block
- Updated block desrciption
- Changed example / block preview so it would look like a table of contents list
- Added new eslint rule to force object curly formatting for import & export statements

bfintal:
TODO:
- [ ] Style tab > General > List Alignment
- [ ] Change bullet option to a select control
- [ ] change "icons & numbers" to "Bullets & Numbers"

gumacahin:
Outstanding:

- [ ] Move list type to general panel 
- [ ] Number list/gradient settings does not function correctly
- [ ]  Initial click to to visibility button fails to toggle visibility

gumacahin:
- [X] Background image disappears when typography settings uses gradient
- [X] H1 should not be included by default
- [ ] Level indentation is wrong when level's are skipped
- [ ] Block > Horizontal alignment settings not working

gumacahin:
Progress:

- [X] Missing headings in frontend
- [ ] Skipping heading levels
- [ ] Duplicate anchors = warning
- [X] Disabling h1/h2 will use the wrong index in visibility toggle

Issue #2182, opened 2022-02-22
Default posts block does not render an image. 


prconcepcion:
1. Add a posts block.
2. Select default variation. 
3. Image is not seen.

https://user-images.githubusercontent.com/49634311/155043459-3429eb9d-66ca-4cd9-8a8e-8b36b35d5a8b.mov

PR #2188, opened 2022-02-28
Fixes missing featured image in default variation of Posts Block

gumacahin:
Fixes https://github.com/gambitph/Stackable/issues/2182 <del>and other variations that appears broken</del>

* <del>Add filter to reorganize content for vertical-2</del>
* <del>Add componentOrder array in variation.js for variations that require</del>
  them.</del>
  
  <ins>This is because of a quirk in source control</ins>

### DIFF
--- a/src/block/posts/design.js
+++ b/src/block/posts/design.js
@@ -38,37 +38,5 @@ const list = ( output, style, attributes, comps ) => {
 	)
 }
 
-// Vertical 2
-const invertedVertical = ( output, style, attributes, comps ) => {
-	if ( style !== 'inverted-vertical' ) {
-		return output
-	}
-
-	const {
-		contentOrder = [],
-	} = attributes
-
-	// If style is vertical, put the image wrapper outside of the article tag.
-	const contents = contentOrder.map( key => {
-		if ( key === 'featured-image' ) {
-			return null
-		}
-
-		const comp = comps[ key ]
-		return comp
-	} )
-
-	return (
-		<>
-			<article className="stk-container-padding">
-				{ compact( contents ).map( content => content ) }
-			</article>
-			{ comps[ 'featured-image' ] }
-		</>
-	)
-}
-
 addFilter( 'stackable.posts.edit.item.output', 'list', list )
 addFilter( 'stackable.posts.save.item.output', 'list', list )
-addFilter( 'stackable.posts.edit.item.output', 'invertedVertical', invertedVertical )
-addFilter( 'stackable.posts.save.item.output', 'invertedVertical', invertedVertical )
--- a/src/block/posts/design.js
+++ b/src/block/posts/design.js
@@ -38,5 +38,37 @@ const list = ( output, style, attributes, comps ) => {
 	)
 }
 
+// Vertical 2
+const invertedVertical = ( output, style, attributes, comps ) => {
+	if ( style !== 'inverted-vertical' ) {
+		return output
+	}
+
+	const {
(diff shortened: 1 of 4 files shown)


src/block/posts/variations.js
```diff
@@ -94,6 +95,13 @@ const variations = applyFilters(
 			name: 'horizontal',
 			description: sprintf( _x( '%s Layout', 'Block layout name', i18n ), __( 'Horizontal', i18n ) ),
 			attributes: {
+				contentOrder: [
+					'title',
+					'meta',
+					'category',
+					'excerpt',
+					'readmore',
+				],
```
bfintal, line 104: I think you can remove this since this doesn't do anything.

src/block/posts/variations.js
```diff
@@ -106,6 +114,13 @@ const variations = applyFilters(
 			name: 'horizontal-2',
 			description: sprintf( _x( '%s Layout', 'Block layout name', i18n ), __( 'Horizontal 2', i18n ) ),
 			attributes: {
+				contentOrder: [
+					'title',
+					'meta',
+					'category',
+					'excerpt',
+					'readmore',
+				],
```
bfintal, line 123: I think you can remove this since this doesn't do anything.

src/block/posts/variations.js
```diff
@@ -143,6 +158,13 @@ const variations = applyFilters(
 			description: sprintf( _x( '%s Layout', 'Block layout name', i18n ), __( 'Vertical Card', i18n ) ),
 			attributes: {
 				className: 'is-style-vertical-card',
+				contentOrder: [
+					'title',
+					'meta',
+					'category',
+					'excerpt',
+					'readmore',
+				],
```
bfintal, line 167: I think you can remove this since this doesn't do anything.

src/block/posts/variations.js
```diff
@@ -154,6 +176,13 @@ const variations = applyFilters(
 			name: 'vertical-card-2',
 			description: sprintf( _x( '%s Layout', 'Block layout name', i18n ), __( 'Vertical Card 2', i18n ) ),
 			attributes: {
+				contentOrder: [
+					'title',
+					'meta',
+					'category',
+					'excerpt',
+					'readmore',
+				],
```
bfintal, line 185: I think you can remove this since this doesn't do anything.

src/block/posts/design.js
```diff
@@ -38,5 +38,37 @@ const list = ( output, style, attributes, comps ) => {
...
+	return (
+		<>
+			<article className="stk-container-padding">
+				{ compact( contents ).map( content => content ) }
+			</article>
+			{ comps[ 'featured-image' ] }
+		</>
+	)
+}
+
```
bfintal, line 70: What does this do? It looks like without this, the inverted vertical works as intended

bfintal (changes requested):
Please double check this, as I'm checking it, the posts block looks broken because of two things:
1. The content order entries sometimes contains the featured image (or in the default layout, it was missing it)
2. Resetting the content order can produce an error.

It seems that fixing the content orders of the layouts in the premium fixes things already: https://github.com/bfintal/Stackable-Premium/commit/9e15a3b16e284ae7a6773e137bf491a5bffb794d

No need to add the other `contentOrder` values in the other variations (see my code comments) since they do not do anything. If free, they're not used, if premium, they're not used either.

To address the error visible when resetting sorting, it might be easier to just disable resetting the content order. I've added this in.

Please see the code comments

Issue #2183, opened 2022-02-23
Changing a Column's html tag stops it from displaying correcty.


bfintal:
Information from here: https://wordpress.org/support/topic/wp-editor-columns-container-bug/

> In the block “Columns / Container” there is a problem that when selecting a HTML tag other than “Default” or “DIV”, the display in the editor no longer matches the display on the live website.
The display on the live website is correct, the display in the WP Editor is incorrect.
Only with the DIV tag are the columns displayed in the WP Editor across the entire available width and the content within them accordingly.
>
> Or expressed technically:
> With the DIV, the width is set to 100% in the stylesheet.
> ```
> [data-type^=”stackable/”]>div {
> width: 100%;
> }
>```
> With the HTML tags – such as Section (which I like to use for semantic HTML) this specification is missing, although it is also an HTML block element.

bfintal:
Fixed by 770a10ad7171b1dacc713b4701909a1088a6b6a8

PR #2202, opened 2022-03-07
Add `width: 100%` to columns block inner element.

gumacahin:
Fixes https://github.com/gambitph/Stackable/issues/2183.


bfintal (commented):
This doesn't fix all issues of #2183, the css pointed out in the issue also affects other blocks in the editor as well.

bfintal:
Let's go with 770a10ad7171b1dacc713b4701909a1088a6b6a8 instead

Issue #2205, opened 2022-03-11
Inner content of Hero Block disappears when video is used as background.


gumacahin:
**Describe the bug**
On the editor the inner content of Hero Block disappears when block has a video background and one or both of the top/bottom separators are enabled. In addition, the separators appears to be invisible on the editor. On the frontend the inner content of the hero block is not visible. These features function properly when image instead of a video.

**To Reproduce**
1. Add a hero block
2. Enable Background
3. Optional: disable container background
4. Select a video background.
5. Enable top/bottom separator.
6. The block will resize to make room for the separator but they will be invisible
7. Go preview and the content of the hero block is not visible.

**Expected behavior**
When a video is used as a background the top/bottom seperators, when enabled should be visible in the editor. In addition, the contents of the hero block should be visible on the front end.

PR #2206, opened 2022-03-14
Fix: Inner content of hero and featured block disappears when using video background

gumacahin:
Fixes https://github.com/gambitph/Stackable/issues/2205.

### DIFF
--- a/src/block-components/separator/style.scss
+++ b/src/block-components/separator/style.scss
@@ -5,7 +5,7 @@
 		.stk-inner-blocks {
 			z-index: 5;
 		}
-		> * {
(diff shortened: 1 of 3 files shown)


bfintal (commented):
Setting the `z-index` to `0` works, but for this one it's better to fix the root cause of the issue - in the separator styles. The issue seems to be that when there's a top/bottom separator, the inner elements are moved to a higher `z-index`, this should not include the background video.

bfintal:
Closed in favor of https://github.com/gambitph/Stackable/commit/bbfa85137f4be52159b28cb25856ce0f345d84fb

PR #2213, opened 2022-03-16
Added a filter that would prevent the last column of the query loop block from wrapping whenever the global body text size was <= 16px. 

### LINKED ISSUE #2209: Query loop does not follow # of columns.
1. Add the native query loop.
2. Make global stackable body text <= 16px. 
3. See that the last column is being wrapped in back end and front end.

prconcepcion:
Fixes #2209.

### DIFF
--- a/src/global-settings.php
+++ b/src/global-settings.php
@@ -533,6 +533,9 @@ if ( ! class_exists( 'Stackable_Global_Settings' ) ) {
 				$selectors[] = $tag . '[class*="wp-block-"]';
 			}
 
+			// Fixes columns issue with Native Posts block.
+			add_filter( 'stackable_global_typography_selectors', array( $this, 'posts_block_columns_fix' ), 10, 2);
+
 			return apply_filters( 'stackable_global_typography_selectors', $selectors, $tag );
 		}
 
@@ -814,6 +817,15 @@ if ( ! class_exists( 'Stackable_Global_Settings' ) ) {
 				return substr_replace( $subject, $replace, $pos, strlen( $search ) );
 			}
 			return $subject;
+		}
+
+		public function posts_block_columns_fix( $selectors, $tag ) {
+			if ( $tag === 'li' ) {
+				$index = array_search('[data-block-type="core"] li', $selectors);
+				$selectors[$index] = '[data-block-type="core"] li:not(.wp-block-post)';
+			}
+
+			return $selectors;
 		}
 	}
 
--- a/src/global-settings.php
+++ b/src/global-settings.php
@@ -829,7 +829,7 @@ if ( ! class_exists( 'Stackable_Global_Settings' ) ) {
 		 * @return void
 		 */
 		public function posts_block_columns_fix( $selectors, $tag ) {
-			// Prevent global settings from affecting hte native wp block post.
+			// Prevent global settings from affecting the native wp block post.
 			if ( $tag === 'li' ) {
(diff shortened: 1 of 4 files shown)


src/global-settings.php
```diff
@@ -815,6 +818,15 @@ public function str_replace_first( $search, $replace, $subject ) {
 			}
 			return $subject;
 		}
+
+		public function posts_block_columns_fix( $selectors, $tag ) {
+			if ( $tag === 'li' ) {
+				$index = array_search('[data-block-type="core"] li', $selectors);
+				$selectors[$index] = '[data-block-type="core"] li:not(.wp-block-post)';
```
bfintal, line 825: Spacing standards:
```suggestion
				$index = array_search( '[data-block-type="core"] li', $selectors );
				$selectors[ $index ] = '[data-block-type="core"] li:not(.wp-block-post)';
```

src/global-settings.php
```diff
@@ -533,6 +533,9 @@ public function form_tag_selector( $tag ) {
 				$selectors[] = $tag . '[class*="wp-block-"]';
 			}
 
+			// Fixes columns issue with Native Posts block.
+			add_filter( 'stackable_global_typography_selectors', array( $this, 'posts_block_columns_fix' ), 10, 2 );
+
```
bfintal, line 538: This is the wrong location for an `add_filter`. Because the filter is added in here, it will be ran and added multiple times. The `add_filter` should only be called once. It would be better to place this in the `__construct` function where most of the hooks are added

src/global-settings.php
```diff
@@ -815,6 +818,15 @@ public function str_replace_first( $search, $replace, $subject ) {
 			}
 			return $subject;
 		}
+
+		public function posts_block_columns_fix( $selectors, $tag ) {
```
bfintal, line 822: Please add some documentation here, add why this is needed

bfintal: @prconcepcion I added comments on the function

src/global-settings.php
```diff
@@ -815,6 +818,25 @@ public function str_replace_first( $search, $replace, $subject ) {
...
+		 * block incorrectly wraps below and leaves a gap.
+		 *
+		 * @param array $selectors
+		 * @param string $tag
+		 * @return void
+		 */
+		public function posts_block_columns_fix( $selectors, $tag ) {
+			// Prevent global settings from affecting hte native wp block post.
+			if ( $tag === 'li' ) {
+				$index = array_search( '[data-block-type="core"] li', $selectors );
```
bfintal, line 834: @prconcepcion Just realized that the next line will produce an error if the `array_search` fails, you need to make sure that the next line only runs when `array_search` succeeds

bfintal (changes requested):
Please see comments

src/global-settings.php
```diff
@@ -815,6 +818,27 @@ public function str_replace_first( $search, $replace, $subject ) {
...
+		 *
+		 * @param array $selectors
+		 * @param string $tag
+		 * @return void
+		 */
+		public function posts_block_columns_fix( $selectors, $tag ) {
+			// Prevent global settings from affecting the native wp block post.
+			if ( $tag === 'li' ) {
+				$index = array_search( '[data-block-type="core"] li', $selectors );
+				if ( ! is_bool( $index ) ) {
```
bfintal, line 835:
```suggestion
				if ( $index !== false ) {
```

PR #2229, opened 2022-04-07
Added conditional rendering to prevent mini nav from rendering if ListView is null. 

prconcepcion:
Fixes block errors seen in sites that use Stackable ver. 3.1.5 and WP ver 5.8.4 and below.


bfintal:
@prconcepcion code review: by doing the condition check inside the return statement, you're performing additional rendering which will not be used anyway. It would be better to move up the condition and `return null` right away

PR #2262, opened 2022-04-25
Excerpt will display properly if there are chinese-japanese-korean characters. 

### LINKED ISSUE #2203: Chinese characters in posts block excerpt do not get reduced when adjusting excerpt length.
1. Add a post.
2. Add chinese characters to the post.
3. Publish the post.
4. Add a posts block.
5. Reduce the except length.
6. See that there is no effect on the except length with chinese characters. 

https://user-images.githubusercontent.com/49634311/157182169-8f0bd8f3-d4b0-4b3b-9120-7fdd95be3852.mov

prconcepcion:
Fixes #2203

### DIFF
--- a/src/block/posts/index.php
+++ b/src/block/posts/index.php
@@ -115,7 +115,8 @@ if ( ! function_exists( 'generate_render_item_from_stackable_posts_block' ) ) {
 			} else {
 				$excerpt = implode( ' ', $excerpt );
 			}
-
+			// Trim the excerpt differently if there are chinese characters.
+			$excerpt = filter_cjk_characters( $excerpt, $trim_to_length );
 			$excerpt = wp_kses_post( $excerpt );
 			$new_template = str_replace( '!#excerpt!#', $excerpt, $new_template );
 		} else {
@@ -130,6 +131,25 @@ if ( ! function_exists( 'generate_render_item_from_stackable_posts_block' ) ) {
 		$new_template = str_replace( '!#readmoreText!#', esc_html( $readmore_text ), $new_template );
 
 		return $new_template;
+	}
+
+	/**
+	 * Checks if there are cjk characters in the string
+	 * and cuts it based on character count.
+	 *
+	 * @since 3.3.2
+	 * @param string | string $excerpt
+	 * @param int | int $trim_to_length
+	 *
+	 * @return string trimmed excerpt if it contains a cjk character, original excerpt if not.
+	 */
+	function filter_cjk_characters( $excerpt, $trim_to_length ) {
+		$chinese_excerpt = preg_match_all("/\p{Han}+/u", $excerpt, $matches);
+		if($chinese_excerpt) {
+			$sample = mb_substr( $excerpt, 0, $trim_to_length ) . '...';
+			return $sample;
+		}
+		return $excerpt;
 	}
 }
 
--- a/src/block/posts/util.js
+++ b/src/block/posts/util.js
@@ -26,6 +26,16 @@ import { applyFilters } from '@wordpress/hooks'
  */
 import variations from './variations'
 
+// Check if there are east asian characters. Trim with regards to each char length rather than word count.
+const trimCJKCharacters = ( excerpt, excerptLength ) => {
+	const regexCJK = /[\u3040-\u30ff\u3400-\u4dbf\u4e00-\u9fff\uf900-\ufaff\uff66-\uff9f]/
+	if ( excerpt.match( regexCJK ) ) {
+		const outputExcerpt = excerpt.substring( 0, excerptLength ) + '...'
+		return outputExcerpt
+	}
+	return excerpt
+}
+
 export const META_SEPARATORS = {
 	dot: '·',
 	space: ' ',
@@ -204,6 +214,9 @@ export const generateRenderPostItem = attributes => {
 		} else {
 			excerptString = post.post_excerpt_stackable
 		}
+
+		// Trim the excerpt differently if there are cjk characters.
+		excerptString = trimCJKCharacters( excerptString, excerptLength )
 
 		const excerpt = excerptString && (
 			<div


src/block/posts/index.php
```diff
@@ -115,7 +115,8 @@ function generate_render_item_from_stackable_posts_block( $post, $attributes, $t
 			} else {
 				$excerpt = implode( ' ', $excerpt );
 			}
-
+			// Trim the excerpt differently if there are chinese characters.
+			$excerpt = filter_cjk_characters( $excerpt, $trim_to_length );
```
bfintal, line 119: While this would technically work, the logic is off. The CJK trimming shouldn't be called after the if/else statement, because this would mean that the strings will always be trimmed and will be trimmed twice. 

Our logic above is as follows (in the `if` statement):
1. If the excerpt is too long, trim it
2. If not, just use as-is

The logic in the PR is:
1. If the excerpt is too long, trim it
2. If not, just use as-is
3. After 1 & 2, trim CJK characters

The trimming of CJK should make the logic be:
1. If the excerpt has CJK characters, trim it if it's too long,
2. If the excerpt doesn't have CJK characters but it's too long, trim it
4. If not, just use as-is

bfintal (changes requested):

src/block/posts/index.php
prconcepcion: Got it.

src/block/posts/index.php
```diff
@@ -108,14 +108,23 @@ function generate_render_item_from_stackable_posts_block( $post, $attributes, $t
...
+				if( strlen( $untrimmed_excerpt ) > $trim_to_length ) {
+					// Trim according to string length.
+					$excerpt = mb_substr( $untrimmed_excerpt, 3, $trim_to_length ) . '...';
+				} else {
+					$excerpt = $untrimmed_excerpt;
+				}
+			} elseif ( count( $excerpt ) > $trim_to_length ) {
 				$excerpt = implode( ' ', array_slice( $excerpt, 0, $trim_to_length ) ) . '...';
 			} else {
 				$excerpt = implode( ' ', $excerpt );
```
bfintal, line 126:
```suggestion
				$excerpt = $untrimmed_excerpt;
```

src/block/posts/index.php
```diff
@@ -108,14 +108,23 @@ function generate_render_item_from_stackable_posts_block( $post, $attributes, $t
 
 		// Trim the excerpt.
 		if ( ! empty( $excerpt ) ) {
+			$untrimmed_excerpt = $excerpt;
 			$excerpt = explode( ' ', $excerpt );
 			$trim_to_length = (int) $excerpt_length;
-			if ( count( $excerpt ) > $trim_to_length ) {
+
+			// Check if there are CJK characters.
+			if ( preg_match_all("/\p{Han}+/u", $untrimmed_excerpt, $matches) ) {
```
bfintal, line 116: Mind the spaces
```suggestion
			if ( preg_match_all( "/\p{Han}+/u", $untrimmed_excerpt, $matches ) ) {
```

src/block/posts/index.php
```diff
@@ -108,14 +108,23 @@ function generate_render_item_from_stackable_posts_block( $post, $attributes, $t
...
+				if( strlen( $untrimmed_excerpt ) > $trim_to_length ) {
```
bfintal, line 117:
```suggestion
				if ( strlen( $untrimmed_excerpt ) > $trim_to_length ) {
```

src/block/posts/util.js
```diff
@@ -199,7 +199,22 @@ export const generateRenderPostItem = attributes => {
...
+			if ( postExcerptStackable.length > excerptLength || 55 ) {
+				if ( excerptLength === 1 ) {
+					excerptString = postExcerptStackable.substring( 3, 4 ) + '...'
+				} else {
+					excerptString = postExcerptStackable.substring( 3, excerptLength === '' ? 55 : excerptLength + 3 ) + '...'
+				}
+				// Outerconditional is always true even though postExcerptStackable.length < 55.
+				if ( postExcerptStackable.length < 55 && excerptLength === '' ) {
+					excerptString = postExcerptStackable
+				}
```
bfintal, line 215: 1. Why is there a different line for when the `excerptLength` is `1`, there should be no reason to split this into an `if`-`else`

2. In this `postExcerptStackable.substring( 3, 4 )`, why is the string trimming process started on index `3`? Shouldn't it start with `0`?

3. What is the last `if` statement for? Looks unnecessary.

Won't this do? Or is there an edge case where this doesn't work?
```suggestion
				excerptString = postExcerptStackable.substring( 0, excerptLength || 55 ) + '...'
```

bfintal (changes requested):
Please see inline comments for some enhancements

src/block/posts/index.php
```diff
@@ -108,14 +108,23 @@ function generate_render_item_from_stackable_posts_block( $post, $attributes, $t
...
+			$untrimmed_excerpt = $excerpt;
 			$excerpt = explode( ' ', $excerpt );
 			$trim_to_length = (int) $excerpt_length;
-			if ( count( $excerpt ) > $trim_to_length ) {
+
+			// Check if there are CJK characters.
+			if ( function_exists( 'mb_substr' ) && preg_match_all( "/\p{Han}+/u", $untrimmed_excerpt, $matches ) ) {
+				if ( strlen( $untrimmed_excerpt ) > $trim_to_length ) {
+					// Trim according to string length.
+					$excerpt = mb_substr( $untrimmed_excerpt, 3, $trim_to_length ) . '...';
```
bfintal, line 119: Why does the index start with `3`, shouldn't it be `0`?
```suggestion
					$excerpt = mb_substr( $untrimmed_excerpt, 0, $trim_to_length ) . '...';
```

src/block/posts/util.js
```diff
@@ -199,7 +199,19 @@ export const generateRenderPostItem = attributes => {
...
+		const regexCJK = /[\u3040-\u30ff\u3400-\u4dbf\u4e00-\u9fff\uf900-\ufaff\uff66-\uff9f]/
+		if ( postExcerptStackable.match( regexCJK ) ) {
+			// Extract the string inbetween the tags and newline.
+			const tempExcerptString = postExcerptStackable.match( '<p>(.*)<\/p>\\n' )
+
+			if ( tempExcerptString[ 1 ].length > ( excerptLength || 55 ) ) {
+				excerptString = tempExcerptString[ 1 ].substring( 0, excerptLength === '' ? 55 : excerptLength ) + '...'
+			} else {
+				excerptString = tempExcerptString[ 1 ]
+			}
```
bfintal, line 213: The `tempExcerptString[ 1 ]` errors out & I'm not sure why we need `tempExcerptString`, we can just use this directly:
```suggestion
			if ( postExcerptStackable.length > ( excerptLength || 55 ) ) {
				excerptString = postExcerptStackable.substring( 0, excerptLength === '' ? 55 : excerptLength ) + '...'
			} else {
				excerptString = postExcerptStackable
			}
```

bfintal (changes requested):
Please see comments. The PHP side gives the wrong substr string. The JS produces an error

src/block/posts/util.js
```diff
@@ -199,7 +199,19 @@ export const generateRenderPostItem = attributes => {
 
 		// Trim the excerpt.
 		let excerptString = postExcerptStackable.split( ' ' )
-		if ( excerptString.length > ( excerptLength || 55 ) ) {
+
+		//Checks if there are CJK characters present
+		const regexCJK = /[\u3040-\u30ff\u3400-\u4dbf\u4e00-\u9fff\uf900-\ufaff\uff66-\uff9f]/
+		if ( postExcerptStackable.match( regexCJK ) ) {
+			// Extract the string inbetween the tags and newline.
+			const tempExcerptString = postExcerptStackable.match( '<p>(.*)<\/p>\\n' )
```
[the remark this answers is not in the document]
prconcepcion: The string that I obtained when I would use the `postExcerptStackable` would contain the `<p>`  which is why I made it get the string between those tags and the reason why I made the function `substing(3,excerptLength+3)` start at 3 before and then also added a +3. Same reason for the `$untrimmed_excerpt` in the `php` file.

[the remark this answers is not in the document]
prconcepcion: The reason why I also made an `if (excerptLength === 1 ) ...` back then was because it would look like this in the backend if the excerpt length is 1:
image

src/block/posts/util.js
prconcepcion: `const tempExcerptString = postExcerptStackable.match( '<p>(.*)<\/p>\\n' )` returns an array and the 1st element is not the match. Though I am not sure if this is the best way to do it.

andeng1106:
Added this issue to Release Cycle Project since a user raised this issue. See https://www.facebook.com/groups/wpstackable/permalink/1195920727963105/

PR #2282, opened 2022-05-05
Margin-bottom is 0 in icon list so that themes' CSS wont affect it.

### LINKED ISSUE #2253: [Blocksy] Suggestion from Blocksy: A small improvement for icon list block would be to add this CSS
A small improvement for icon list block would be to add this CSS. You had such a CSS in V2, this will make the themes CSS to not apply their CSS to this list (we need to check whether this affects the how the icon list currently works)
```
.wp-block-stackable-icon-list ul {
	margin-bottom: 0;
}
.wp-block-stackable-icon-list ul li:last-child {
	margin-bottom: 0;
}
```

prconcepcion:
Fixes #2253.

### DIFF
--- a/src/block/icon-list/style.scss
+++ b/src/block/icon-list/style.scss
@@ -36,11 +36,11 @@
 	margin-right: var(--stk-alignment-margin-right);
 }
 
-//Suggestion from blocksy, prevents.
-.wp-block-stackable-icon-list ul {
+// Prevents theme's css from affecting the list block.
+.stk-block-icon-list :is(ul, ol) {
 	margin-bottom: 0;
-}
 
-.wp-block-stackable-icon-list ul li:last-child {
-	margin-bottom: 0;
+	li:last-child {
+		margin-bottom: 0;
+	}
 }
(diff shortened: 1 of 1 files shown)


src/block/icon-list/style.scss
```diff
@@ -35,3 +35,12 @@
...
 }
+
+//Suggestion from blocksy, prevents.
+.wp-block-stackable-icon-list ul {
+	margin-bottom: 0;
+}
+
+.wp-block-stackable-icon-list ul li:last-child {
+	margin-bottom: 0;
+}
```
bfintal, line 46: 1. This should follow the class naming we use. We use `.stk-block-icon-list` instead of `.wp-block-stackable-icon-list`

2. We also use `ol` here, you may need to use `:is(ul, ol)`

bfintal (changes requested):

src/block/icon-list/style.scss
```diff
@@ -35,3 +35,12 @@
 	margin-left: var(--stk-alignment-margin-left);
 	margin-right: var(--stk-alignment-margin-right);
 }
+
+//Suggestion from blocksy, prevents.
```
bfintal, line 39: Please also adjust the comment, no need to mention that it's a suggestion from blocksy

PR #2283, opened 2022-05-06
Container opacity and shadows can now be visible with colors from Blocksy color palette. 

### LINKED ISSUE #2226: Container color opacity does not change when you choose a blocksy color.
1. Activate blocksy theme. 
2. Add a block that has a container (ex. card).
3. Change the color opacity.
4. See that the color of the container becomes white and the opacity does not change. 

https://user-images.githubusercontent.com/49634311/161262091-bced1293-6bd6-4923-9ae9-a61e930206fd.mov

### LINKED ISSUE #2254: [Blocksy] Column block - The box shadow color is not working if I apply a color from the palette
**Describe the bug**
The box shadow color is not working if I apply a color from the palette - https://d.pr/v/W1JZDY - 
These colors are working in other places like text color, background color

Might be related to: https://github.com/gambitph/Stackable/issues/2226

**To Reproduce**
1. Add new page
2. Try adding Wireframes Card 1
3. In one of the columns, try clicking one Column from the list view
4. Once a column is selected, go to Style tab > Container Borders & Shadow
5. In Shadow/Outline, switch Hover State to Hover
6. Click the Tool icon
7. Set Vertical Offset to 20
8. Set Blur to 80
9. Change Shadow Color to one of the color from the Blocksy palette
10. Check if it doesn't apply as a shadow
11. Change Shadow Color to one of the color from the color picker
12. Check if it applies as a shadow

**Expected behavior**
Should be able to use Blocksy color palette as shadow color.

**Screenshots**

https://user-images.githubusercontent.com/103395655/164207679-537ebea3-f9bc-4166-b91b-6bc4e3b010d5.mp4

prconcepcion:
Fixes #2226, Fixes #2254

### DIFF
--- a/src/block-components/helpers/backgrounds/style.js
+++ b/src/block-components/helpers/backgrounds/style.js
@@ -6,6 +6,7 @@ import {
 	__getValue,
 	getStyles,
 	useStyles,
+	extractColor,
 } from '~stackable/util'
 import { Style as StyleComponent } from '~stackable/components'
 
@@ -41,13 +42,9 @@ const getStyleParams = ( options = {} ) => {
 					getAttribute( 'backgroundMediaUrl', 'mobile' )
 
 				if ( ! backgroundColorType && backgroundColorOpacity !== '' && ! hasBackground ) {
-					// Checks if color comes from Blocksy color pallete.
-					if ( value?.startsWith( 'var(--paletteColor' ) ) {
-						const hexColorIndex = value.search( /#[a-zA-Z0-9]{6}/g )
-						const hexColor = value.substring( hexColorIndex, hexColorIndex + 7 )
-						return `${ hexToRgba( hexColor || '#ffffff', backgroundColorOpacity || 0 ) }`
-					}
-					return `${ hexToRgba( value || '#ffffff', backgroundColorOpacity || 0 ) }`
+					// Checks if color comes from Non-stackable color palette.
+					const hexColor = extractColor( value )
+					return `${ hexToRgba( hexColor || '#ffffff', backgroundColorOpacity || 0 ) }`
 				}
 				return value
 			},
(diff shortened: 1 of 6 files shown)


src/block-components/helpers/backgrounds/style.js
```diff
@@ -41,9 +41,14 @@ const getStyleParams = ( options = {} ) => {
 					getAttribute( 'backgroundMediaUrl', 'mobile' )
 
 				if ( ! backgroundColorType && backgroundColorOpacity !== '' && ! hasBackground ) {
+					// Checks if color comes from Blocksy color pallete.
+					if ( value?.startsWith( 'var(--paletteColor' ) ) {
+						const hexColorIndex = value.search( /#[a-zA-Z0-9]{6}/g )
+						const hexColor = value.substring( hexColorIndex, hexColorIndex + 7 )
```
bfintal, line 47: If you check:
- `src/block-components/typography/edit.js`
- `src/block-components/button/edit.js`

they have this line:
```js
const value = _value?.startsWith( 'var(--stk-global-color' ) ? _value.match( /(#[^\)]*)/g )[ 0 ] : _value
```
which gets the color from a CSS custom property, similar to what you're doing.

I refactored the extracting code to support both `--stk-global-color` and `--paletteColor` so we can reuse it in your PR and in the 2 files I listed above. Add a new util function in `src/util` and use it in the codes mentioned (and in your fix):
```js
export const extractColor = value => {
	if ( color?.startsWith( 'var(--' ) ) {
		if ( color.match( /(#[^\)]*)/g ) ) {
			return color.match( /(#[^\)]*)/g )[ 0 ]
		}
	}
	return color
}

**Review by bfintal — CHANGES_REQUESTED** [review]

File: src/block-components/helpers/backgrounds/style.js (hunk 1 above)
Comment (prconcepcion) [reply src/block-components/helpers/backgrounds/style.js:47]: Got it

File: src/util/index.js [hunk 2]
```diff
@@ -437,3 +437,19 @@ export const sanitizeIdAttr = str => {
...
 }
+
+/**
+ * Extract the hexadecimal value of the color from a non-stackable color.
+ *
+ * @param {string} value Non-stackable color
+ * @return {string} Hexadecimal value of the color
+ */
+export const extractColor = value => {
+	const color = value
```
bfintal, line 448: This line doesn't really do anything, you can just use `value` instead of `color` in the rest of the function

bfintal (changes requested):
Almost there, just one change to do then we can move this for testing

PR #2331, opened 2022-07-12
Missing i18n on a string in block/map/edit.js

DAnn2012:
Update edit.js


bfintal:
Looks good! Thanks for the PR, you rock 🤘

PR #2347, opened 2022-08-01
fix (size & spacing issue): exclude horizontal alignment style when in full-width

### LINKED ISSUE #2338: Size & Spacing issue - Resetting Max. Content Width hides Content Horizontal Align settings but doesn't reset it
**Describe the bug**
If a user uses the Max. Content Width, that's where the feature Content Horizontal Align only shows. 
But if I unintentionally reset Max. Content Width and forgot to reset Content Horizontal Align, I might mistakenly saw a bug since the Content Horizontal Align is not resetting and automatically hides whenever I reset the Max. Content Width.

Note: This might also apply to all blocks that has Size & Spacing like:
- Inspector > Block tab > Size & Spacing
- Inspector > Style tab > Container Size & Spacing panel
- etc...

**To Reproduce**
1. Copy this code to your editor: 
[max content width & content horizontal align issue.txt](https://github.com/gambitph/Stackable/files/9177310/max.content.width.content.horizontal.align.issue.txt)

2. Follow the steps from the video provided in **Screenshots**
3. Check if the issue also happens

**Expected behavior**
Possible solutions:
1. If Max Content Width is reset, Content Horizontal Align settings should also reset
2. Do not hide Content Horizontal Align feature if it still has settings on and not yet reset even though Content Horizontal Align settings are reset
3. Show both Max. Content Width and Content Horizontal Align features by default

**Screenshots**

https://user-images.githubusercontent.com/103395655/180680022-35a7d307-c034-4aaa-9af3-3f40bb7d9ec4.mp4

vmc08:
Fixes #2338


src/block-components/helpers/size/style.js
```diff
@@ -123,6 +124,10 @@ const getStyleParams = ( options = {} ) => {
 			attrNameTemplate,
 			responsive: 'all',
 			hasUnits: 'px',
+			enabledCallback: getAttribute => {
+				const deviceType = getDeviceType()
```
bfintal, line 128: This will not work as intended because if you try and get the `deviceType` here, you will just end up getting the _current_ device type/preview of the editor (not the device type of the style being currently generated). Unfortunately the `device` isn't provided in the `enabledCallback` arguments.

I think we should try an alternative approach.. instead of `enabledCallback`, try using the `valuePreCallback`.. that callback is used to override the value of the attribute before it gets processed for style generation. If the function `valuePreCallback` returns an empty string `''`, the style isn't generated.

src/hooks/use-device-type.js
```diff
@@ -1,7 +1,20 @@
...
+	// In some editors, there is no edit-post / preview device type. If that
+	// happens, we just set our own internal device type.
+	if ( wpSelect( 'core/edit-post' ) ) {
+		deviceType = wpSelect( 'core/edit-post' ).__experimentalGetPreviewDeviceType()
+	} else {
+		deviceType = wpSelect( 'stackable/device-type' ).getDeviceType()
+	}
+	return deviceType || ''
+}
+
```
bfintal, line 17: Inline with the above suggestion, you won't need this anymore.

bfintal (changes requested):
Please see my comment about `valuePreCallback`.

Also, kindly clean up the commit, some entries are just linting fixes and the package-lock.json is also updated in this PR.

bfintal (changes requested):
Yes this is a much better looking solution 👍

Almost there, all that's left is to clean up the PR:
- remove the `package-lock.json` from the PR
- the file `src/block-components/block-div/edit.js` doesn't have relevant changes to the fix
- the file `src/block-components/helpers/size/edit.js` doesn't have relevant changes to the fix

vmc08:
>

bfintal:
Replaced by #2348

PR #2357, opened 2022-08-05
fix (button-group, button): Removes button margin if button is full width to prevent misaligned stacked buttons

### LINKED ISSUE #2249: [Blocksy] When buttons stack, there is a space on the right, this makes them not to be centered
**Describe the bug**
Reported by Blocksy. When there are two buttons and I changed view to mobile and tablet view, the first button is not aligned with the second one because of the additional space in padding.

**To Reproduce**
1. Add new page
2. Add a Feature block
3. In the button portion, add another button
4. Publish page and view it
5. Open console and toggle device toolbar and set dimensions to mobile or tablet view
6. Check if the two buttons are not aligned

Other observations:
If button group has Collapse Buttons on "Mobile" or "Tablet" it aligns. But if it's set to "Don't collapse", the buttons are not aligned.

**Expected behavior**
When buttons stack in mobile or tablet view, it should be aligned to center.

**Screenshots**
If Style > General > Collapse Buttons on is set to "Don't collapse", buttons align to center in Desktop view only. Mobile and Tablet view is not aligned to center.

https://user-images.githubusercontent.com/103395655/164181457-21f1db5d-4df5-4abe-b31d-61cb82473752.mp4

If Style > General > Collapse Buttons on is set to "Desktop" or "Tablet", buttons align to center in all views. - CORRECT

https://user-images.githubusercontent.com/103395655/164182837-2ceba2ac-af7e-43e5-be07-8b4991054c43.mp4

If Style > General > Collapse Buttons on is set to "Mobile", buttons align to center in Desktop and Mobile view only. Tablet view is not aligned to center.

https://user-images.githubusercontent.com/103395655/164182568-a7682097-f589-4b45-94a3-b932a0abe222.mp4

prconcepcion:
Fixes #2249


src/block/button/style.scss
```diff
@@ -13,3 +13,7 @@
 .stk-button {
 	box-sizing: border-box;
 }
+
+.stk-block-button-group--flex-wrap .stk-block-button.stk-button--full-width {
+	margin-inline-end: unset !important;
```
bfintal, line 18: Do not use `!important` here, instead, make sure that you're selector is more specific than the one you're trying to override: `.stk-button-group [data-block]:not(:last-child) .stk-block`

src/block/button-group/save.js
```diff
@@ -21,25 +21,33 @@ import {
...
 	const responsiveClass = getResponsiveClasses( props.attributes )
+	const flexWrapClass = flexWrap ? 'stk-block-button-group--flex-wrap' : flexWrap
 
-	const blockClassName = classnames( [
-		props.className,
-		'stk-block-button-group',
-		responsiveClass,
-	] )
+	const blockClassName = classnames(
+		applyFilters( 'stackable.buttonGroup.save.blockClassNames', [
```
bfintal, line 42: Please use `stackable.button-group.save.blockClassNames` as the name for the filter, the `button-group` corresponds to the name of the block `stackable/button-group`

src/block/button/save.js
```diff
@@ -25,25 +25,36 @@ import { useBlockProps } from '@wordpress/block-editor'
  * Internal dependencies
  */
 import { ButtonStyles } from './style'
+import { applyFilters } from '@wordpress/hooks'
 
 export const Save = props => {
 	const {
+		attributes,
```
bfintal, line 32: You extracted into its own variable `attributes`, so now it will not be anymore part of the `propsToPass` during the deconstruction. This will lead to unexpected issues

bfintal (changes requested):
Please see inline comments

src/block/button/style.scss
```diff
@@ -13,3 +13,17 @@
...
+.stk-block-button-group.stk-block-button-group--flex-wrap {
+
+	[data-block]:not(:last-child) .stk-block {
+		margin-inline-end: unset;
+	}
+
+	:is(.stk-block, .stk-button):not(:last-child) {
+		margin-inline-end: unset;
+	}
+
```
bfintal, line 27:
```suggestion
	[data-block]:not(:last-child) .stk-block {
		margin-inline-end: unset;
	}

	:is(.stk-block, .stk-button):not(:last-child) {
		margin-inline-end: unset;
	}
```

bfintal: We can combine these.

bfintal (changes requested):

prconcepcion:
A new solution towards the #2249  shall be created.

PR #2370, opened 2022-08-08
fix (icon-control): empty when clicking icon reset

### LINKED ISSUE #2329: Icon block issue - If resetting an icon, it doesn't show the default icon
**Describe the bug**
If I reset the icon, it doesn't show the default icon which is this star (see image below). Right now, it just shows blank.
image
image

**To Reproduce**
1. Add an icon block
2. Change icon
3. Change color or any settings
4. Click Reset
5. Check if icon resets back as a default icon (star)

**Expected behavior**
It should reset back to default icon (star)
image
image

**Screenshots**

https://user-images.githubusercontent.com/103395655/177743746-23eb96cb-5178-4fee-a963-6f67a159e0ca.mp4

vmc08:
Fixes #2329


bfintal (changes requested):
Since this adds a defaultValue across all components/blocks that use the `IconControl`, other controls that really have a blank icon as default now incorrectly have the `reset` button on them. 

Instead of setting `defaultIcon` as the default prop, try using:

```js
IconControl.defaultProps = {
    // ...
    defaultValue: '',
}
```

Then for the icon block (or for the ones that incorrectly shows the reset button), just pass the `defaultIcon` to the component instead.

bfintal:
Unfortunately this doesn't work as intended. Try this:

1. Add a button block, and select an icon for the button
2. Unselect & re-select the button block
3. Open the inspector to the button's icon settings
4. See that the reset button for the icon doesn't appear anymore.

The behavior should be that for the button block, the reset button always appears if there's a selected icon. Then for the icon block, the reset button appears if it's not the default icon.

I would like to suggest going back to the previous solution, but the default value should just be passed by icon block

bfintal (changes requested):
Unfortunately this doesn't work as intended. Try this:

1. Add a button block, and select an icon for the button
2. Unselect & re-select the button block
3. Open the inspector to the button's icon settings
4. See that the reset button for the icon doesn't appear anymore.

The behavior should be that for the button block, the reset button always appears if there's a selected icon. Then for the icon block, the reset button appears if it's not the default icon.

I would like to suggest going back to the previous solution, but the default value should just be passed by icon block

bfintal (changes requested):
In:
- src/block/icon-button/edit.js
- src/block/icon-list/edit.js
- src/block/icon/edit.js

I would prefer that in these files/components above, we just use the value directly in `defaultValue={ defaultIcon }` instead of adding `defaultProps` since these components won't be used any other way anyway. This is unlike `src/block-components/button/edit.js` which are components that are really used by others.

bfintal (changes requested):
Let's include the Accordion block's icon to this PR. Currently, you can reset the accordion icon and it will go back to a star.

bfintal (changes requested):
It seems that the default icon that the IconControl will reset to is being saved as a block attribute. We'll need to change this implementation because this would mean that all Icon blocks would have 2 icons saved in their HTML form: one is the actual icon, and the other is the icon it will reset to.

The icon default also doesn't belong inside an attribute. I think something like this would work: instead of passing the default icon inside an attribute, you can use a filter inside the IconControl (or the Icon block-component) where you can check whether the icon block is part of an accordion block then change the default icon.

Check the filters at the bottom of `src/block/expand/edit.js` for similar implementations

src/block/accordion/edit.js
```diff
@@ -227,3 +231,41 @@ addFilter( 'stackable.block-default-styles.use-saved-style', 'stackable/icon-lab
...
+		[ clientId ]
+	)
+
+	if ( isAccordionIcon ) {
+		const activeVariation = getActiveBlockVariation( block?.name, block?.attributes )
+		return ( activeVariation.name === 'plus' )
+			? applyFilters( 'stackable.block-component.plus.icon-open' ) : defaultIcon
+	}
+	return starIcon
+} )
```
bfintal, line 271: You may find the `parentTree` provided by the `useBlockContext` hook helpful here

src/block/accordion/edit.js
```diff
@@ -171,42 +170,16 @@ export default compose(
...
-				return false
-			}
-			return {
-				isAccordionIcon: ! getBlock( accordionClientId ).innerBlocks[ 0 ].clientId !== columnClientId,
-				block: getBlock( accordionClientId ),
-			}
-		},
-		[ clientId ]
-	)
+	const accordionBlock = parentTree.find( pt => pt.name === 'stackable/accordion' )
```
bfintal, line 178: Because this matches all the parents of the block, this will also affect other Icon pickers that are nested as the contents of the Accordion block. For example if you have an accordion block, then inside the actual accordion container, you have an icon block, this would also reset to a chevron.

You'll need to check specifically if the icon block belongs to the header part of the accordion block.

bfintal (changes requested):

bfintal (changes requested):
In all icon blocks, you can still reset the icon to blank. Expectation is that you can't reset the icon to blank, it should reset to the default icon.

bfintal (changes requested):

PR #2374, opened 2022-08-10
fix (color-picker): use debouncing to improve input field performance

### LINKED ISSUE #2339: Global Color Palette issue - Text is not registering correctly for Style name field
**Describe the bug**
When typing a text in Style name field in Global Color Palette, it's not registering correctly and seems to have a delay.

**To Reproduce**
1. In Editor, go to Global Settings > Global Color Palette panel
2. Click + to add new color
3. Click the color selected and edit Style name to your desired color (make sure to type fast or in normal speed)
4. Check if all the keys you inputted is not registered on the field
5. Try clicking backspace to delete all the texts

**Expected behavior**

- Should register all characters inputted through keyboard in Style name field

- Should be able to delete each character smoothly by clicking the backspace key

**Screenshots**

https://user-images.githubusercontent.com/103395655/181166474-9c5ef397-79ec-455e-a012-d5447c7e4a19.mp4

vmc08:
Fixes #2339


bfintal (changes requested):
I think debouncing might be the correct way to handle this. However, if you take a look at `ColorPickerTextarea` as a component by itself (ignoring where the component will be used), there seems to be no need to debounce the onChange prop since it's just a simple `input` handler. It might be better to do the debouncing where the `onChange` function is defined because that's where the slowdown is happening. Please see line 308 where the `ColorPickerTextarea` is used:

```jsx
<ColorPickerTextArea
	onChange={ onChangeStyleName }
	value={ color.name || '' }
/>
```

bfintal:
@vmc08 I actually like this solution 👍

PR #2395, opened 2022-08-22
Feat (count-up): Control the speed of the count up animation

### LINKED ISSUE #2355: Count Up block - Add a feature to adjust count speed
A user is requesting for a feature if they are able to adjust the count speed for Count Up block which is not present right now.

Ability to adjust Count Up block speed like a Count Speed

N/A

prconcepcion:
Fixes #2355

### DIFF
--- a/src/block/count-up/edit.js
+++ b/src/block/count-up/edit.js
@@ -66,14 +66,15 @@ const Edit = props => {
 				<PanelAdvancedSettings
 					title={ __( 'Counter', i18n ) }
 					id="count-up"
+					initialOpen={ true }
 				>
 					<AdvancedRangeControl
-						label={ __( 'Duration', i18n ) }
+						label={ __( 'Duration (ms)', i18n ) }
 						attribute="duration"
 						min={ 100 }
 						max={ 10000 }
 						step={ 100 }
-						placeholder="1000ms"
+						placeholder="1000"
 					>
 
 					</AdvancedRangeControl>
@@ -85,8 +86,8 @@ const Edit = props => {
 			<Transform.InspectorControls />
 			<Typography.InspectorControls
 				hasTextTag={ false }
-				initialOpen={ true }
 				hasTextShadow={ true }
+				initialOpen={ false }
 			/>
 			<EffectsAnimations.InspectorControls />
 			<CustomAttributes.InspectorControls />
--- a/src/block/count-up/edit.js
+++ b/src/block/count-up/edit.js
@@ -24,7 +24,9 @@ import {
 } from '~stackable/block-components'
 import { version as VERSION, i18n } from 'stackable'
 import classnames from 'classnames'
-import { InspectorTabs } from '~stackable/components'
+import {
+	InspectorTabs, InspectorBlockControls, PanelAdvancedSettings, AdvancedRangeControl,
+} from '~stackable/components'
 import { useBlockHoverClass } from '~stackable/hooks'
 import { withQueryLoopContext } from '~stackable/higher-order'
 
@@ -60,7 +62,21 @@ const Edit = props => {
 		<Fragment>
 
 			<InspectorTabs />
+			<InspectorBlockControls>
+				<PanelAdvancedSettings
+					title={ __( 'Speed', i18n ) }
+					id="count-up">
+					<AdvancedRangeControl
+						label={ __( 'Duration', i18n ) }
+						attribute="duration"
+						min={ 1 }
+						max={ 10 }
+						placeholder=""
+					>
 
+					</AdvancedRangeControl>
+				</PanelAdvancedSettings>
+			</InspectorBlockControls>
 			<Alignment.InspectorControls />
 			<BlockDiv.InspectorControls />
 			<Advanced.InspectorControls />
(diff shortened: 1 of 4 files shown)


src/block/count-up/edit.js
```diff
@@ -60,7 +62,21 @@ const Edit = props => {
 		<Fragment>
 
 			<InspectorTabs />
+			<InspectorBlockControls>
+				<PanelAdvancedSettings
+					title={ __( 'Speed', i18n ) }
+					id="count-up">
```
bfintal, line 68:
```suggestion
					id="count-up"
				>
```

src/block/count-up/edit.js
```diff
@@ -60,7 +62,21 @@ const Edit = props => {
...
+					<AdvancedRangeControl
+						label={ __( 'Duration', i18n ) }
+						attribute="duration"
+						min={ 1 }
+						max={ 10 }
+						placeholder=""
```
bfintal, line 74: I find it better if there's a placeholder value for the default.

src/block/count-up/edit.js
```diff
@@ -60,7 +62,21 @@ const Edit = props => {
...
 			<InspectorTabs />
+			<InspectorBlockControls>
+				<PanelAdvancedSettings
+					title={ __( 'Speed', i18n ) }
+					id="count-up">
+					<AdvancedRangeControl
+						label={ __( 'Duration', i18n ) }
+						attribute="duration"
+						min={ 1 }
+						max={ 10 }
```
bfintal, line 73: People might want more control over the duration, please use `ms` for the value of the duraction. min `100`, max `10000` and step `100`

src/block/count-up/frontend-count-up.js
```diff
@@ -13,9 +13,11 @@ class StackableCountUp {
 	callback = entries => {
 		entries.forEach( entry => {
 			const el = entry.target
+			const duration = el.getAttribute( 'data-duration' ) ? el.getAttribute( 'data-duration' ) * 1000 : 1000
+
```
bfintal, line 17: I think it might be good if we shorten this to:
```suggestion
			const duration = el.getAttribute( 'data-duration' ) || 1000
```

I suggested above to use `ms` instead of seconds, so I used `1000` instead of multiplying by 1000

src/block/count-up/edit.js
```diff
@@ -60,7 +62,21 @@ const Edit = props => {
 		<Fragment>
 
 			<InspectorTabs />
+			<InspectorBlockControls>
```
bfintal, line 65: I feel that this control is better suited to be inside the **Style tab** and as the first panel and should be open by default

bfintal (changes requested):
It works! But I think we can further improve this, please see inline comments for my requested changes.

src/block/count-up/edit.js
```diff
@@ -60,7 +62,21 @@ const Edit = props => {
...
+				<PanelAdvancedSettings
+					title={ __( 'Speed', i18n ) }
```
bfintal, line 67: It might be better if the title was "Counter" instead of "Speed" so that it would still be good if our panel were to contain other options aside from the duration.

src/block/count-up/frontend-count-up.js
```diff
@@ -13,9 +13,11 @@ class StackableCountUp {
 	callback = entries => {
 		entries.forEach( entry => {
 			const el = entry.target
+			const duration = el.getAttribute( 'data-duration' ) ? el.getAttribute( 'data-duration' ) : 1000
+
```
bfintal, line 17:
```suggestion
			const duration = el.getAttribute( 'data-duration' ) || 1000

```

src/block/count-up/edit.js
```diff
@@ -60,7 +62,23 @@ const Edit = props => {
...
+					title={ __( 'Counter', i18n ) }
+					id="count-up"
+				>
+					<AdvancedRangeControl
+						label={ __( 'Duration', i18n ) }
+						attribute="duration"
+						min={ 100 }
+						max={ 10000 }
+						step={ 100 }
+						placeholder="1000ms"
```
bfintal, line 76:
```suggestion
						placeholder="1000"
```

src/block/count-up/edit.js
```diff
@@ -60,7 +62,23 @@ const Edit = props => {
 		<Fragment>
 
 			<InspectorTabs />
+			<InspectorStyleControls>
+				<PanelAdvancedSettings
+					title={ __( 'Counter', i18n ) }
+					id="count-up"
+				>
+					<AdvancedRangeControl
+						label={ __( 'Duration', i18n ) }
```
bfintal, line 71:
```suggestion
						label={ __( 'Duration (ms)', i18n ) }
```

bfintal (changes requested):
Please see inline comments. Also, please have the Counter panel be open at the start (and the typography panel closed)

src/block/count-up/edit.js
```diff
@@ -60,15 +62,32 @@ const Edit = props => {
...
+				<PanelAdvancedSettings
+					title={ __( 'Counter', i18n ) }
+					id="count-up"
+					initialOpen={ true }
+				>
+					<AdvancedRangeControl
+						label={ __( 'Duration (ms)', i18n ) }
+						attribute="duration"
+						min={ 100 }
+						max={ 10000 }
```
bfintal, line 75:
```suggestion
						max={ 5000 }
```

PR #2401, opened 2022-08-25
refactor (default-block-style): do not replace inner blocks if it doe…

### LINKED ISSUE #2396: Inserting a Button should focus on the button text
When you insert a button, you should be able to type right away. Right now you have to click on the button text first after inserting it

This happens when there's a Default Button block set.

vmc08:
…sn't exist

Fixes #2396

### DIFF
--- a/src/hooks/use-saved-default-block-style.js
+++ b/src/hooks/use-saved-default-block-style.js
@@ -79,7 +79,7 @@ export const useSavedDefaultBlockStyle = blockProps => {
 				attributes[ attrName ] = newAttributes[ attrName ]
 			} )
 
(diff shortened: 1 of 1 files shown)


src/hooks/use-saved-default-block-style.js
```diff
@@ -79,15 +79,17 @@ export const useSavedDefaultBlockStyle = blockProps => {
 				attributes[ attrName ] = newAttributes[ attrName ]
 			} )
 
+			if ( blockData.innerBlocks.length ) {
```
bfintal, line 82:
```suggestion
			if ( blockData.innerBlocks?.length ) {
```

PR #2402, opened 2022-08-25
fix (i18n): missing i18n on strings in call-to-action/variations.js and some deprecated.js files

DAnn2012:
fix (i18n): missing i18n on strings in call-to-action/variations.js and some deprecated.js files


DAnn2012:
Hello @bfintal 

as soon as you can, could you take a look at this PR?

Thanks.

vmc08 (approved):
lgtm! thanks for addressing this translation issues.

PR #2419, opened 2022-09-08
fix (background): added prop to make transparent color visible in the 2nd gradient color palette

### LINKED ISSUE #2358: Gradient Background - When you set a gradient background, the transparent option doesn’t show for the 2nd color
**Describe the bug**
When you set a gradient background, the transparent option doesn’t show for the 2nd color
This applies to:
- Block tab > Background > Background Color # 2
- Container Background > Background Color # 2

**To Reproduce**
1. Add a block with background (e.g. Columns/Container)
2. Add gradient color for your background (Background > Gradient)
3. Check in Background Color # 2 if there's a transparent option in the palette

**Expected behavior**
There should be a transparent option for 2nd gradient color

**Screenshots**
image

prconcepcion:
Fixes #2358

### DIFF
--- a/src/block-components/helpers/backgrounds/edit.js
+++ b/src/block-components/helpers/backgrounds/edit.js
@@ -19,7 +19,7 @@ import { getAttributeName, urlIsVideo } from '~stackable/util'
 /**
  * WordPress dependencies
  */
-import { useCallback, Fragment } from '@wordpress/element'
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
Since the attribute `backgroundColor2` is now used in style generation, please also add it in the list of attribute dependencies (add it in the `dependencies` array)

PR #2420, opened 2022-09-08
refactor (color-palette): localize state changes and add debouncing

### LINKED ISSUE #2315: Highlight color feature issue - Unable to change color through color code system (Hex, RGB, HSL)
**Describe the bug**
- When I try to change the color (Text or Highlight color) through Custom by deleting the numbers, it cannot be deleted properly
- When copy pasting, it works fine. But when deleting then typing it doesn't.
- Notes:
	- Applies to all blocks that has text
	- Applies to three highlight options: Normal, Highlight, Low

**To Reproduce**
1. Add any block that has text (e.g. Text block or Button block etc.)
2. Add sample text
3. Select text then click Highlight from toolbar
4. Select text from palette
5. Click the custom color selector
6. Click the Settings icon
7. In Hex, try typing or deleting text in the field
8. In RGB, try typing or deleting text in the R, G, B fields
9. In HSL, try typing or deleting text in the H, S, L fields

**Expected behavior**
Should be able to change text color by typing or deleting in the fields.

**Screenshots**

https://user-images.githubusercontent.com/103395655/172298829-d18b2cda-ae8a-45e8-a7d2-ac3d25926e08.mp4

vmc08:
Fixes #2315

### DIFF
--- a/src/components/color-palette-control/index.js
+++ b/src/components/color-palette-control/index.js
@@ -41,15 +41,15 @@ const ColorPaletteControl = memo( props => {
 		className = '',
 	} = props
 
+	const [ _value, _onChange ] = useControlHandlers( props.attribute, props.responsive, props.hover, props.valueCallback, props.changeCallback )
+	const [ _propsToPass, controlProps ] = extractControlProps( props )
+
 	let value = typeof props.value === 'undefined' ? _value : props.value
 	if ( typeof value === 'string' && value.includes( '--stk-global-color' ) && value.match( /#[\d\w]{6}/ ) ) {
 		value = value.match( /#[\d\w]{6}/ )[ 0 ]
 	}
 
 	const [ colorValue, setColorValue ] = useState( value )
-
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
Some comments:
1. You get a block error when you add a Heading block. I had a Heading block prior to trying the code changes, when I refreshed the editor and selected the heading block, it resulted in an error. `undefined _value`
2. I see that you added a delay so that when adjusting the color value, it won't adjust on you right away. However, this doesn't seem to fix the inherent problem.

If you have a value `#f00000`, put your cursor on the end then press backspace and wait. The value will adjust to `#000000` after a while. The behavior should be that the value shouldn't adjust by itself and should remain in `#f0000` (with your cursor at the end).

Another scenario is that if you have `#f00000`, then delete the character until you get `#f00`, the value auto-adjusts to `#ff0000` (it should remain at `#f00` with your cursor at the end)

bfintal:
Let's not fix this. It appears that something built in with the text formatting is causing the issue. This issue is also present in the native highlight toolbar option

PR #2428, opened 2022-09-14
Feat/2427 progress circle block

### LINKED ISSUE #2427: Progress Circle Block
New block: `stackable/progress-circle`
Block spec: https://www.notion.so/Progress-Circle-Block-Design-Doc-eab886e266d84a0eab88c7884514d43d

vmc08:
Fixes #2427


src/block/progress-circle/meta-data.json
```diff
@@ -0,0 +1,13 @@
...
+	"title": "Progress Circle",
+	"description": "Adds a circular progress bar block.",
+	"category": "common",
+	"usesContext": [ "postId", "postType", "queryId" ],
+	"keywords": [
+		"Progress",
+		"Progress Circle"
+	],
+	"textdomain": "stackable-ultimate-gutenberg-blocks",
+	"stk-type": "essential"
```
bfintal, line 12: Better to move this to special because we want to keep the **essential** blocks to only the very basic ones
```suggestion
	"stk-type": "special"
```

src/components/advanced-range-control/index.js
```diff
@@ -69,15 +70,26 @@ const AdvancedRangeControl = props => {
 		placeholderRender = null
 	}
 
+	const dynamicContentProps = useDynamicContentControlProps( {
+		value: typeof props.value === 'undefined' ? value : props.value,
+		onChange: typeof props.onChange === 'undefined' ? onChange : props.onChange,
+		isFormatType: propsToPass.isFormatType,
```
bfintal, line 76: We can omit the `isFormatType` to make it always false. From what I know this option makes it so that the dynamic content can be multiple instances, or allows part of the value to be dynamic content, both cases are usually for text content. But for the range control, the value is always a whole number so we can not use this.
```suggestion
		isFormatType: propsToPass.isFormatType,
```

src/components/advanced-range-control/index.js
```diff
@@ -90,6 +102,8 @@ const AdvancedRangeControl = props => {
 
 AdvancedRangeControl.defaultProps = {
 	allowReset: true,
+	isDynamic: false,
+	isFormatType: true,
```
bfintal, line 106: In line with the `isFormatType` suggestion, we can remove this
```suggestion
```

src/components/dynamic-content-control/editor.scss
```diff
@@ -7,7 +7,6 @@
 .stk-dynamic-content-control {
 	display: flex;
 	align-items: flex-start;
-	overflow: hidden;
```
bfintal, line 10: Is this change okay with the other controls that use the dynamic content button?

src/block/progress-circle/save.js
```diff
@@ -0,0 +1,53 @@
...
+		>
+			<CustomCSS.Content attributes={ attributes } />
+			<div
+				className="stk-progress-circle"
+				style={ {
+					'--thickness': `${ attributes.thickness }px`,
+					'--percent': attributes.percentage,
+					'--progress-color': attributes.progressColor,
+					'--progress-background': attributes.progressBackgroundColor,
+				} }
```
bfintal, line 32: I'm inclined to do these in `style.js` instead of in `save.js` so that if we need to change things because of an update, we won't have to deprecate things.

bfintal (changes requested):
Please see me initial code review

src/components/dynamic-content-control/editor.scss
bfintal: I'd like to suggest that so we won't affect existing styles that use this class name, write a new selector for the range control that does `overflow: auto` instead.

bfintal (changes requested):
Some comments:

1. Now that I've been trying to actually use the new block, I think people might find the text block inside confusing. It's not obvious that you might need to click on the text to edit it, and changing the percentage value doesn't change the text. I think we might be better off changing our implementation. Instead of providing an InnerBlock with a text block.. let's just make things simple and add a built-in text in the middle of the progress circle for the percentage label, you can use the Typography block component for this. Let's not add a counting animation to the text.

2. Along with the suggestion above, add a new "Label" panel (add this after the progress bar panel):
  - label panel with a toggle (enabled by default, this would take the place of the "display percentage" option) - use the typography panel for this
  - label prefix (string, blank by default)
  - label suffix (string, `%` by default)

3. Add a new "Size" option to control the diameter of the progress circle
4. Rearrange the order of the options: (note that the "display percentage" option is not listed below, see comment 2)
  - Percentage
  - Size
  - Thickness
  - Bar Color (make the title "Bar Color")
  - Background Color
  - Rounded
  - Animate
  - Accessibility label
6. The progress circle should be centered by default - we should add a way to change the left/center/right align of this as well
7. Animate should be enabled by default
8. The "rounded" option doesn't do anything (it's always rounded even if disabled)
9. When adding a new percentage circle block, it animates (let's turn the animation off in the editor)
10. Let's add a gradient color option, you can add gradients using:

```jsx
<defs>
    <linearGradient id="gradient">
      <stop offset="0%"  stop-color="#f00"/>
      <stop offset="100%"  stop-color="#000"/>
    </linearGradient>
  </defs>
```

```css
// Only add this when using a gradient.
stroke: url(#gradient);
```

src/block-components/progress-bar/edit.js
```diff
@@ -0,0 +1,123 @@
...
+	InspectorStyleControls,
+	AdvancedToggleControl,
+	AdvancedRangeControl,
+	AdvancedTextControl,
+	ColorPaletteControl,
+} from '~stackable/components'
+import {
+	useBlockAttributesContext, useBlockSetAttributesContext, useAttributeEditHandlers,
+} from '~stackable/hooks'
+import { DEFAULT_PERCENT, DEFAULT_THICKNESS } from '~stackable/block-components/progress-bar/attributes'
```
bfintal, line 17: Let's import this internally so we won't have to run through aliasing.
```suggestion
import { DEFAULT_PERCENT, DEFAULT_THICKNESS } from './attributes'
```

src/block/progress-circle/save.js
```diff
@@ -0,0 +1,62 @@
...
+		>
+			<ProgressCircleStyles.Content { ...props } />
+			<CustomCSS.Content attributes={ attributes } />
+			<div
+				className={ divClassNames }
+				role="progressbar"
+				aria-valuemin="0"
+				aria-valuemax="100"
+				aria-valuenow={ attributes.progressPercent }
+				aria-valuetext={ attributes.progressAriaValueText }
```
bfintal, line 44: As a precautionary measure to prevent user input from messing up the block structure, please pass the text value through `striptags()`

src/block/progress-circle/save.js
```diff
@@ -0,0 +1,62 @@
...
+	const blockClassNames = classnames( [
+		className,
+		'stk-block-progress-circle',
+		responsiveClass,
+	] )
+
+	const divClassNames = classnames( [
+		'stk-progress-circle',
+		{
+			'with-animation': attributes.progressAnimate,
```
bfintal, line 27: Let's prefix this into a modifier class:
```suggestion
			'stk--with-animation': attributes.progressAnimate,
```

src/block/progress-circle/style.scss
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+}
+
+@media (prefers-reduced-motion) {
+  .stk-progress-circle {
+		&.with-animation {
+			.stk-progress-circle__bar {
+				transition: none;
+				will-change: unset;
```
bfintal, line 60: I think this will still work without this rule?
```suggestion
```

bfintal (changes requested):
Please see other inline code comments

src/block-components/progress-bar/edit.js
```diff
@@ -0,0 +1,137 @@
...
+						min="0"
+						sliderMax="100"
+						step="1"
+						placeholder=""
+						isDynamic
+					/>
+					<AdvancedRangeControl
+						label={ __( 'Size', i18n ) }
+						attribute="progressSize"
+						step="1"
```
bfintal, line 74: I think we can tweak this for more common-scenarios:
```suggestion
						min="0"
						sliderMin="50"
						sliderMax="300"
						step="1"
```

bfintal (changes requested):
Some UI and block-mechanic comments:
1. Please see inline code comments for minor tweaks
2. For Stackable blocks, when you insert a new block, the option controls as much as possible should be blank values, show the default value as the placeholder and should have no reset buttons present.

Here's an example of some controls found by adding an Image block:
image

Notice that the Zoom default value is `1`, but the range control itself doesn't have any value inside the input, instead it just shows the `1` as a placeholder (the text is color gray). The range slider is on the correct location of `1` as well, so that when the user slides it from the default position, it will be smooth from transitioning to a blank, to a value from the slider. There is no reset button present either.

In contrast, the controls of the progress circle all have values in them already - we should change these into placeholders. Some of them have reset buttons present as well.

To address this, for example for the **thickness** control, change these props:

```diff
<AdvancedRangeControl
	label={ __( 'Thickness', i18n ) }
	attribute="progressThickness"
	min="8"
	sliderMax="100"
	step="1"
-	placeholder=""
+	placeholder="8"
-	default={ DEFAULT_THICKNESS }
/>
```

Then for the attribute definition, we should almost always have `''` as the default value:

```diff
progressThickness: {
	type: 'number',
-	default: DEFAULT_THICKNESS,
+	default: '',
},
```

3. Ideally, when you add new blocks and leave everything by default (add the block and then hit update), the block itself should not generate any CSS `<style>` tags. Currently, if you add a new progress circle block, it will generate the following:

```html
<style>
.stk-7661e01 .stk-progress-circle {
	--progress-thickness: 8px !important;
	--progress-percent: 50 !important;
	--progress-color: #3498db !important;
	--progress-background: #f0f0f0 !important;
	--progress-size: 150px !important;
	--progress-rounded: unset !important;
}
</style>
```

Please make it so that no `<style>` tags are generated by default - when a style tag is present, it's just there to override the default values.

These default css values should be added into `style.scss` instead of all progress bars.

In connection to the previous comment, setting default values in the attribute definitions to blank strings `''` will help with this process.

4. I noticed that for certain attributes, the progress shows incorrectly. Try: percentage `100`, size `300` and thickness `3`. It will not show as `100%`:

image

I think we may need to change some CSS values, please check this implementation here: https://codepen.io/arrayzero/pen/LbvJvW

They have the `stroke-dasharray` set as `314` and then they compute for `stroke-dashoffset` depending on the radius of the circle.

src/block/progress-circle/style.scss
```diff
@@ -0,0 +1,72 @@
+@import "common";
+
+.stk-block-progress-circle {
+	display: flex;
+}
```
bfintal, line 5: Spaces and tabs seem to be mixed up here, you may need to install **Stylelint** in your VS Code to lint this css file properly.

src/block/progress-circle/edit.js
```diff
@@ -0,0 +1,118 @@
...
+	 const workAroundClass = classnames( {
+		'force-redraw': forceRedraw,
+	} )
+
+	useEffect( () => {
+		if ( attributes.progressSize ) {
+			setForceRedraw( true )
+			const timer = setTimeout( () => {
+				setForceRedraw( false )
+			}, 100 )
```
bfintal, line 61: The delay is a bit noticeable, I tried removing the delay (this defaults it to `0`) and it seems that this works.
```suggestion
			} )
```

src/block/progress-circle/style.scss
```diff
@@ -0,0 +1,72 @@
...
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    h4, .wp-block-stackable-count-up .stk-block-count-up__text {
+      font-weight: 200;
+      font-size: 3.5rem;
+      margin: 0;
+      word-break: inherit;
+    }
```
bfintal, line 48: For me it's better if we don't pre-style this for the user. If we remove this, the text style will inherit from the theme.
```suggestion
```

src/block/progress-circle/edit.js
```diff
@@ -0,0 +1,118 @@
...
+					aria-valuetext={ striptags( attributes.progressAriaValueText || undefined ) }
+				>
+					<svg viewBox={ `0 0 ${ attributes.progressSize } ${ attributes.progressSize }` } className={ workAroundClass }>
+						<circle className="stk-progress-circle__background" />
+						<circle className="stk-progress-circle__bar" />
+					</svg>
+					{ attributes.show && (
+						<div className="number">
+							<Typography
+								tagName="h4"
```
bfintal, line 102: Let's make this into a span. Headings will be incorrectly picked up by the ToC block and SEO.
```suggestion
								tagName="span"
```

src/block/progress-circle/save.js
```diff
@@ -0,0 +1,79 @@
...
+				aria-valuetext={ striptags( attributes.progressAriaValueText || undefined ) }
+			>
+				<svg>
+					<circle className="stk-progress-circle__background" />
+					<circle className="stk-progress-circle__bar" />
+				</svg>
+				{ attributes.show && (
+					<div className="number">
+						<Typography.Content
+							tagName="h4"
```
bfintal, line 66: Let's make this into a span. Headings will be incorrectly picked up by the ToC block and SEO.
```suggestion
							tagName="span"
```

DAnn2012:
Hello @vmc08 

I would appreciate your opinion on this comment https://github.com/gambitph/Stackable/pull/2333#issuecomment-1249655070

Thank you.

bfintal (changes requested):
1. If you don't provide an aria text value, the attribute `aria-valuetext` should not be present 
2. Let's namespace the classes `number` and `animate` to `stk-progress-circle__number' and `stk--animate` 
3. The suffix has a reset button if the value is `%` (that's the default value), the reset button should not be present if the value is `%`
4. Progress circle should be centered by default (but the toolbar button should have nothing selected)
5. If you increase the size to `200`, then duplicate the progress bar block, the duplicated block shows a smaller background circle

bfintal (approved):

src/block/progress-circle/save.js
```diff
@@ -0,0 +1,105 @@
...
+			<div className={ containerClassNames }>
+				<div
+					className={ divClassNames }
+					role="progressbar"
+					aria-valuemin="0"
+					aria-valuemax="100"
+					aria-valuenow={ derivedPercent }
+					{ ...( attributes.progressAriaValueText && {
+						'aria-valuetext': striptags( attributes.progressAriaValueText ),
+					} ) }
```
bfintal, line 69: I find this easier to read, kindly double check since I know `undefined` will prevent the attribute from showing:
```suggestion
					aria-valuetext={ attributes.progressAriaValueText ? striptags( attributes.progressAriaValueText ) : undefined }
```

src/block/progress-circle/save.js
```diff
@@ -0,0 +1,105 @@
...
+				>
+					<svg>
+						{ attributes.progressColorType === 'gradient' && (
+							<defs>
+								<linearGradient
+									id={ `gradient-${ attributes.uniqueId }` }
+									{ ...( attributes.progressColorGradientDirection && {
+										gradientTransform: `rotate(${ attributes.progressColorGradientDirection })`,
+									} ) }
```
bfintal, line 78: I find that this is easier to read:
```suggestion
									gradientTransform={ attributes.progressColorGradientDirection ? `rotate(${ attributes.progressColorGradientDirection })` : undefined }
```

bfintal (changes requested):

bfintal (changes requested):
Comments on the progress bar:
- add width control (place before the height slider), value can be %, px or vw
- height slider should be from 8 - 50
- radius max should be 25
- percentage label should work with dynamic content
- label default value should be "Design"
- prefix control label should be "Percentage Prefix"
- suffix control label should be "Percentage Suffix"
- if the label font size is too large and the percentage size too small, the text breaks. The word should not break (editor only)

src/block-components/progress-bar/style.js
```diff
@@ -0,0 +1,160 @@
...
+					// should use uniqueId upon saving
+					const color1 = getAttribute( 'progressColor1' ) || ''
+					const color2 = getAttribute( 'progressColor2' ) || ''
+					const direction = getAttribute( 'progressColorGradientDirection' ) || ''
+					const customGradientId = md5( color1 + color2 + direction )
+					return `url(#gradient-${ customGradientId })`
+				}
+				return value
+			},
+			dependencies: [ 'progressColorType', 'progressColor2' ],
```
bfintal, line 64:
```suggestion
			dependencies: [ 'progressColorType', 'progressColor2', 'progressColorGradientDirection' ],
```

src/block/progress-bar/save.js
```diff
@@ -0,0 +1,92 @@
...
+			<CustomCSS.Content attributes={ attributes } />
+			<div
+				className={ divClassNames }
+				role="progressbar"
+				aria-valuemin="0"
+				aria-valuemax="100"
+				aria-valuenow={ derivedPercent }
+				aria-valuetext={ deviedAriaValue ? striptags( deviedAriaValue ) : undefined }
+			>
+				<div className="stk-progress__background">
```
bfintal, line 67: I think it might be good if we eliminate one `div` here

src/block/progress-bar/save.js
```diff
@@ -0,0 +1,92 @@
...
+								<Typography.Content
+									tagName="span"
+									className={ classnames( [ textClassNames, 'stk-progress-text' ] ) }
+									value={ attributes.progressInnerText || attributes.text }
+								/>
+								<Typography.Content
+									tagName="span"
+									className={ classnames( [ textClassNames, 'stk-progress-percent-text' ] ) }
+									value={ derivedValue }
+								/>
```
bfintal, line 80: Let's change these class names to BEM classes: `stk-progress-bar__text` instead of `stk-progress-text`

src/block-components/progress-bar/style.js
```diff
@@ -87,6 +87,16 @@ const getStyleParams = ( { isCircle } ) => {
...
 		},
+		{
+			selector,
+			styleRule: '--progress-bar-border-radius',
+			attrName: 'progressApplyBarRadius',
+			valuePreCallback: ( value, getAttribute ) => {
+				const borderRadius = getAttribute( 'progressBorderRadius' )
+				return value ? borderRadius : undefined
+			},
+			hasUnits: 'px',
```
bfintal, line 104: 2 things:
1. `hasUnits` isn't needed here, if you need to format the value, you can do `format: '%spx'`
2. We'll need `dependencies: [ 'progressApplyBarRadius' ]` here

src/block/progress-bar/style.scss
```diff
@@ -18,6 +18,7 @@ div.stk-block-progress-bar {
 		align-items: center;
 		justify-content: space-between;
 		background-color: var(--progress-color-1, #3498db);
+		border-radius: var(--progress-bar-border-radius, unset);
```
bfintal, line 17: Can probably use `border-radius: var(--progress-bar-border-radius, 0)` here instead

PR #2438, opened 2022-09-30
Feat/2432 progress bar block

### LINKED ISSUE #2432: Progress Bar Block
New block: stackable/progress-bar
Block spec: https://www.notion.so/Progress-Bar-Block-Design-Doc-dc30739cc9e441a38e80b9f517818f36

vmc08:
Fixes #2432

### DIFF
--- a/src/block/progress-bar/edit.js
+++ b/src/block/progress-bar/edit.js
@@ -97,7 +97,7 @@ const Edit = ( {
 				>
 					<div className="stk-progress__background">
 						<div className={ barClassNames }>
-							{ attributes.show && (
+							{ attributes.showText && (
 								<>
 									<Typography
 										tagName="span"
--- a/src/block-components/typography/attributes.js
+++ b/src/block-components/typography/attributes.js
@@ -70,10 +70,6 @@ const typographyAttributes = {
 		type: 'boolean',
 		default: false,
 	},
-	show: {
-		type: 'boolean',
-		default: true,
-	},
 	textPrefix: {
 		type: 'string',
 		default: '',
--- a/src/block-components/typography/edit.js
+++ b/src/block-components/typography/edit.js
@@ -380,8 +380,8 @@ export const Edit = props => {
 				title={ label }
 				initialOpen={ initialOpen }
 				{ ...( hasToggle ? {
-					checked: getAttribute( 'show' ),
-					onChange: updateAttributeHandler( 'show' ),
+					checked: getAttribute( attrNameTemplate !== '%s' ? 'show' : 'showText' ),
+					onChange: updateAttributeHandler( attrNameTemplate !== '%s' ? 'show' : 'showText' ),
 				} : {} ) }
 				id="text"
 			>
--- a/src/block/progress-bar/save.js
+++ b/src/block/progress-bar/save.js
@@ -68,7 +68,7 @@ const Save = props => {
 			>
 				<div className="stk-progress__background">
 					<div className={ barClassNames }>
-						{ attributes.show && (
+						{ attributes.showText && (
 							<>
 								<Typography.Content
 									tagName="span"
--- a/src/block/progress-circle/edit.js
+++ b/src/block/progress-circle/edit.js
@@ -131,7 +131,7 @@ const Edit = ( {
 							<circle className="stk-progress-circle__background" />
 							<circle className="stk-progress-circle__bar" />
 						</svg>
-						{ attributes.show && (
+						{ attributes.showText && (
 							<div className="stk-number">
 								<Typography
 									tagName="span"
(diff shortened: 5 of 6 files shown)


src/block-components/progress-bar/style.js
```diff
@@ -1,43 +1,49 @@
...
 
-const getStyleParams = () => {
+const getStyleParams = ( { isCircle } ) => {
 	return [
 		{
-			selector: '.stk-progress-circle',
-			styleRule: '--progress-thickness',
-			attrName: 'progressThickness',
-			format: '%spx',
+			selector: '[class*="stk-progress-"]',
```
bfintal, line 12: **Attribute contains** selectors are relatively slow. I'd like to suggest that instead of using `[class*="stk-progress-"]`, just change the selectors directly:

```js
const selector = isCircle ? '.stk-progress-circle' : '.stk-progress-bar'
```

```diff
- selector: '[class*="stk-progress-"]',
+ selector,
```

```

File: src/block-components/progress-bar/style.js [hunk 2]
```diff
@@ -1,43 +1,49 @@
...
+			} ),
 		},
 		{
 			selector: '.stk-progress-circle',
 			styleRule: '--progress-color-1',
 			attrName: 'progressColor1',
-			valuePreCallback: ( value, getAttribute, _, state ) => {
-				if ( getAttribute( 'progressColorType', 'desktop', state ) === 'gradient' ) {
+			valuePreCallback: ( value, getAttribute ) => {
+				if ( getAttribute( 'progressColorType' ) === 'gradient' && isCircle ) {
```
bfintal, line 24: Seems that this style is only applicable for when `isCircle` is `true`? If yes, then I recommend using only adding this entry only for circles

src/block-components/progress-bar/style.js
```diff
@@ -75,12 +87,42 @@ const getStyleParams = () => {
...
+			},
+			dependencies: [ 'progressColorType', 'progressColor1', 'progressColor2', 'progressColorGradientLocation1', 'progressColorGradientLocation2', 'progressColorGradientDirection' ],
 		},
+	 ] ),
 	]
 }
 
 export const Style = props => {
-	const progressBarStyles = useStyles( getStyleParams() )
+	const progressBarStyles = useStyles( getStyleParams( { isCircle: props.isCircle } ) )
```
bfintal, line 125: I think we can pass the props directly here.

```suggestion
	const progressBarStyles = useStyles( getStyleParams( props ) )
```

src/block-components/progress-bar/style.js
```diff
@@ -93,6 +135,7 @@ export const Style = props => {
 }
 
 Style.defaultProps = {
+	isCircle: false,
```
bfintal, line 138: Let's follow how the other block-components are passing options:

```suggestion
```

src/block-components/progress-bar/style.js
```diff
@@ -103,7 +146,7 @@ Style.Content = props => {
 		...propsToPass
 	} = props
 
-	const progressBarStyles = getStyles( propsToPass.attributes, getStyleParams() )
+	const progressBarStyles = getStyles( propsToPass.attributes, getStyleParams( { isCircle: props.isCircle } ) )
```
bfintal, line 149:
```suggestion
	const {
		attributes,
		options = {},
		...propsToPass
	} = props

	const progressBarStyles = getStyles( attributes, getStyleParams( options ) )
```

src/block/progress-bar/edit.js
```diff
@@ -0,0 +1,128 @@
...
+		'stk-block-progress-bar',
+		blockAlignmentClass,
+	] )
+
+	const textClassNames = classnames( [
+		'stk-progress-bar__inner-text',
+		textClasses,
+	] )
+
+	const barClassNames = classnames( 'stk-progress__bar', {
```
bfintal, line 59: I think the aim here is to have a common class for the **bar**, it might be better if we have 2 classes here: 
- `stk-progress-bar__bar` to satisfy BEM, and
- `stk-progress__bar` as the shared bar class

src/block/progress-bar/edit.js
```diff
@@ -0,0 +1,128 @@
+/**
+ * Internal dependencies
+ */
+import ProgressBarStyles from './style'
+import { DEFAULT_PERCENT } from '../../block-components/progress-bar/attributes'
```
bfintal, line 5: I just realized that this is a reference to `src/block-components` which is built to a separate module from this file `stk.js`, this might include duplicate code into our `editor_blocks.js`.

Please add another declaration for this value somewhere in `src/block/progress-bar` instead, so we won't have to reference `src/block-components`

bfintal (changes requested):
Please see my initial review

Issue #2350, opened 2022-08-03
Separator block issue - Layer 2 and 3 cannot be toggled OFF if panel is open


andeng1106:
**Describe the bug**
When toggling on Separator block's Layer 2 and 3 feature and when I tried to toggle it off, it doesn't toggle off unless I collapse it

**To Reproduce**
1. Add a Separator block
2. In Inspector > Style, toggle on Layer 2 and 3
3. (optional) Add settings or leave it as is
4. Try to toggle off the panel while it's expanded/open
5. Check if it doesn't toggle off
6. Try to toggle off the panel while it's collapsed/close
7. Check if it toggles off

**Expected behavior**
While Layer 2 and 3 is expanded it should: 
- Be able to toggle off 
- settings should be reset when toggling off
- panel should collapse automatically if toggled off

**Screenshots**

https://user-images.githubusercontent.com/103395655/182514260-6b0e8355-b9ee-45b7-91ad-34acffb3a6fb.mp4

PR #2446, opened 2022-10-12
Try performance improvements on style generation

### LINKED ISSUE #2190: Inverse the behavior of the Heading's top margin from theme
image

### LINKED ISSUE #2483: Icon Popover overlapping when scrolling further
WP 6.0 (Expected):

https://user-images.githubusercontent.com/103395655/197179019-7fe2b6ee-46f6-4227-ae96-4e527fc2b10f.mp4

WP 6.1 (Actual):

https://user-images.githubusercontent.com/103395655/197179119-72ad2021-d9f6-49da-ab54-2262e605d91a.mp4

bfintal:
Fixes #2190, fixes #2483

# Issues to be fixed

https://www.notion.so/11-03-22-Speed-Improvement-Monitor-Issues-Found-807323725a934ccda7b19551dbd67751

TODO:
- [x] FontAwesomeIcon's `prependRender` always keeps firing when hovering over other blocks (to check, add a wireframe that has lots of icons, then hover mouse over different blocks)
- [x] Change `useDispatch` to use `dispatch` instead
- [x] Add custom eslint rule to disallow using `useDispatch`, use `dispatch` instead (causes rerenders)
- [x] rewrite progress bar and progress circle style generation to use `BlockCss` component
- [ ] ~removing `isHovered` in `withBlockWrapperIsHovered` greatly reduces renders when hovering over multiple blocks quickly. Try moving things that use it in the `BlockWrapper` to remove passing `isHovered`.~
- [x] `useAttributeEditHandlers` still uses `useBlockAttributesContext()` that can rerender every time an unrelated attribute changes
- [x] Optimize `BlockCss` component so it will just return right away if no value
- [ ] Replace `AdvancedAutosuggestControl` component with `FormTokenFields`
  - [ ] Remove 'react-autosuggest' package
- [x] Remove unused `GlobalStyleControl` component
- [x] Fix highlight toolbar, it should show the expanded color picker

---

# New eslint rules

## eslint rule: <a id="no-use-dispatch">no-use-dispatch</a>

Using `const { foo } = useDispatch( 'core/bar' )` in a component can cause rerenders. Instead, just call the `dispatch` directly `dispatch( 'core/bar' ).foo()` where you are going to call the function.

```js
// const { __unstableMarkNextChangeAsNotPersistent } = useDispatch( 'core/block-editor' )
// __unstableMarkNextChangeAsNotPersistent()
// New usage
```

---

# Testing Notes

These are the things that have been improved, so these behaviors would need to be tested:
- I changed the way how styles are generated, so we'll need to check whether all the styles are generated correctly
- text block dynamic content has been improved
- hover states have been improved
- left/center/right alignment from the toolbar
- margin bottom resizer has been improved
- block splitting (pressing enter in a text block)
- block duplication
- color pickers have been improved (affected things are global colors, highlight toolbar, color and background color pickers)
- clicking on an image block to upload / change the image has been improved
- showing/hiding of visual indicators/resizers have been improved (column resizers, image resizers, image/column size indicators)
- controls with buttons that open a popover (like the typography settings or adv gradient settings) have been improved (might need to check whether it closes properly, or you can select a value without it closing prematurely)
- icon picker in the inspector has been improved
- icon block unnecessary rerendering has been prevented
- outlines that show when you hover your mouse over a block has been improved
- fixed https://github.com/gambitph/Stackable/issues/2190
- fixed https://github.com/gambitph/Stackable/issues/2350


src/block-components/container-div/index.js
```diff
@@ -22,11 +23,12 @@ export const ContainerDiv = props => {
 	const instanceId = useQueryLoopInstanceId( attributes.uniqueId )
 	let uniqueBlockClass = getUniqueBlockClass( attributes.uniqueId )
 	uniqueBlockClass = instanceId ? uniqueBlockClass + `-${ instanceId }` : uniqueBlockClass
+	const uniqueContainerClass = applyFilters( 'stackable.container-div.uniqueClass.edit', `${ uniqueBlockClass }-container`, uniqueBlockClass )
```
kmgalanakis, line 26: Same here, you most probably forgot to also pass the attributes as a filter argument.

src/components/block-css/index.js
```diff
@@ -433,7 +441,8 @@ const BlockCssEdit = props => {
 		instanceId,
 	} )
 
-	const output = useDynamicContent( css )
+	let output = useDynamicContent( css )
+	output = applyFilters( 'stackable.block-styles.edit', output, getUniqueBlockClass( attributes.uniqueId ) )
```
kmgalanakis, line 445: Same here, you most probably forgot to also pass the attributes as a filter argument.

src/block-components/block-div/index.js
```diff
@@ -37,6 +37,7 @@ export const BlockDiv = props => {
 	const instanceId = useQueryLoopInstanceId( attributes.uniqueId || tempUniqueId )
 	let uniqueBlockClass = getUniqueBlockClass( attributes.uniqueId || tempUniqueId )
 	uniqueBlockClass = instanceId ? uniqueBlockClass + `-${ instanceId }` : uniqueBlockClass
+	uniqueBlockClass = applyFilters( 'stackable.block-div.uniqueClass.edit', uniqueBlockClass )
```
kmgalanakis, line 40: Here you most probably forgot to also pass the attributes as a filter argument.

kmgalanakis (changes requested):
Regarding Toolset I'm still checking how this works, but for sure we are going to need the `attributes` argument for the `*.edit` filters. Thank you

bfintal:
> Regarding Toolset I'm still checking how this works, but for sure we are going to need the `attributes` argument for the `*.edit` filters. Thank you

Hey @kmgalanakis thanks for checking this out! The `attributes` object wasn't intentionally passed on the **edit** side in order to prevent component re-renderings. Instead, the `attributes` object now only contains attributes needed by the current component - e.g. the ones needed for container styles in the Container component - so it doesn't have all the attributes of the block.

For the reliable `attributes` object from the within the hook, I'd like to recommend this instead:

```js
const { clientId } = wp.blockEditor.useBlockEditContext()
const attributes = wp.data.select( 'core/block-editor' ).getBlockAttributes( clientId )
```

DAnn2012:
Hello @bfintal @vmc08  @prconcepcion 

as soon as you can, could you take a look at these PR #2472 #2498 ?

I think these changes are important and urgent for users who are not using English-language WordPress.

Thanks.

bfintal:
> > Regarding Toolset I'm still checking how this works, but for sure we are going to need the `attributes` argument for the `*.edit` filters. Thank you
> 
> Hey @kmgalanakis thanks for checking this out! The `attributes` object wasn't intentionally passed on the **edit** side in order to prevent component re-renderings. Instead, the `attributes` object now only contains attributes needed by the current component - e.g. the ones needed for container styles in the Container component - so it doesn't have all the attributes of the block.
> 
> For the reliable `attributes` object from the within the hook, I'd like to recommend pulling the block's attributes instead:
> 
> ```js
> const { clientId } = wp.blockEditor.useBlockEditContext()
> const attributes = wp.data.select( 'core/block-editor' ).getBlockAttributes( clientId )
> ```

Hey @kmgalanakis just checking up whether this change is good for your end.

kmgalanakis:
@bfintal, what you wrote in that comment made perfect sense and served the purpose of what I needed back then.

Though, I'm no longer working on this project because I'm not part of OnTheGoSystems anymore. You can reach out to them to discuss how to proceed with this.

Thank you.

bfintal:
@kmgalanakis thanks! I'll reach out to them and ask about it :)

PR #2474, opened 2022-10-20
fix (icon-list): remove unused unstable function for wp 6.1 update

### LINKED ISSUE #2442: Icon List encounters block error and can't attempt recovery
Other observations:
- Icon List encounters block error and can't attempt recovery but upon refreshing it displays correctly
- Upon clicking the displayed correctly Icon List near the bullet where you'll type, it encounters a block error again
- It looks okay in frontend

Might be related to this:
https://www.wpbeginner.com/news/whats-coming-in-wordpress-6-1-features-and-screenshots/#:~:text=Ever%20wanted%20to,down%20without%20editing.

https://user-images.githubusercontent.com/103395655/195225962-5a741032-ff68-43b6-a7a1-c5e4b59a239f.mp4

vmc08:
Fixes #2442

### DIFF
--- a/src/block/icon-list/edit.js
+++ b/src/block/icon-list/edit.js
@@ -52,6 +52,17 @@ import { createIconListControls } from './util'
 
 const listTypeOptions = [
 	{
+		label: __( 'Ordered List', i18n ),
+		value: 'ordered',
+	},
+	{
+		label: __( 'Unordered List', i18n ),
+		value: 'unordered',
+	},
+]
+
+const listStyleTypeOptions = [
+	{
 		label: __( 'Number', i18n ),
 		value: 'decimal',
 	},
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
Not working as expected, it's now showing both unordered and ordered buttons active.

image

If the `__unstableIsActiveListType` isn't available anymore, then I would recommend removing these two buttons and adding a select option instead in the icon inspector settings to pick an icon or numbered list.

src/block/icon-list/edit.js
```diff
@@ -255,9 +274,9 @@ const Edit = props => {
 					/>
 
 					<AdvancedSelectControl
-						label={ __( 'List Type', i18n ) }
+						label={ __( 'List Style Type', i18n ) }
```
bfintal, line 277: Revert the string so we retain the translations
```suggestion
						label={ __( 'List Type', i18n ) }
```

PR #2485, opened 2022-10-24
fix (columns): fix column cloning issue

### LINKED ISSUE #2476: Blocks styling is not applied when columns are cloned for the first time
**Describe the bug**
- Issue happens when it's the first time you duplicated columns inside the container but when you move slider back to "1" then next number, it applies correctly
- For the styling issue, happens to all blocks including columns itself
- For the blocks with cloning columns feature, here are the blocks affected:
  - Columns/Container
  - Feature Grid

**To Reproduce**
1. Add a Columns/Container block
2. Add any block and make sure to add any styling (background color, alignment, etc.)
4. Go to Columns/Container > Style > General > Columns 
5. Click the icon near "Columns" feature to clone columns
6. Set Columns to "2"
7. Check the 2nd or next blocks if have the same styling with the 1st block (should also be cloned)
8. Move Columns slider to "1"
9. Repeat steps 6-7 and check if it's now implemented correctly

**Expected behavior**
Should also clone the nested block's styling when cloning columns

**Screenshots**

https://user-images.githubusercontent.com/103395655/196918018-4a29c590-71dc-4508-924b-54c906d47622.mp4

vmc08:
Fixes #2476

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -23,7 +23,7 @@ if ( function_exists( 'sugb_fs' ) ) {
 }
 
 defined( 'STACKABLE_SHOW_PRO_NOTICES' ) || define( 'STACKABLE_SHOW_PRO_NOTICES', true );
-defined( 'STACKABLE_BUILD' ) || define( 'STACKABLE_BUILD', 'free' );
+defined( 'STACKABLE_BUILD' ) || define( 'STACKABLE_BUILD', 'premium' );
 defined( 'STACKABLE_VERSION' ) || define( 'STACKABLE_VERSION', '3.6.0' );
 defined( 'STACKABLE_FILE' ) || define( 'STACKABLE_FILE', __FILE__ );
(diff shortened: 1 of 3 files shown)


plugin.php
```diff
@@ -23,7 +23,7 @@
 }
 
 defined( 'STACKABLE_SHOW_PRO_NOTICES' ) || define( 'STACKABLE_SHOW_PRO_NOTICES', true );
-defined( 'STACKABLE_BUILD' ) || define( 'STACKABLE_BUILD', 'free' );
+defined( 'STACKABLE_BUILD' ) || define( 'STACKABLE_BUILD', 'premium' );
```
bfintal, line 26:
```suggestion
defined( 'STACKABLE_BUILD' ) || define( 'STACKABLE_BUILD', 'free' );
```

PR #2497, opened 2022-10-26
fix (wp6.1): removed redundant text when "Show button text labels" is on

### LINKED ISSUE #2486: Design Library text is odd when "Display button labels" is enabled
**To Reproduce**
1. Add a new page
2. Click the Editor Options and scroll down to "Preferences"
image
3. In Preferences > General, toggle "Show button text labels" on
4. Exit the Preferences pop-up
5. Observe how the Design Library label looks like:
image

**Expected behavior**
Should be labeled as "Design Library"

**Screenshots**

https://user-images.githubusercontent.com/103395655/197668032-9b5dd7ea-c642-4a53-8780-7d8116684310.mp4

prconcepcion:
Fixes #2486

### DIFF
--- a/src/plugins/design-library-button/design-library-button.js
+++ b/src/plugins/design-library-button/design-library-button.js
@@ -34,7 +34,6 @@ const DesignLibraryButton = () => {
 		<Button
 			onClick={ onClick }
 			className="ugb-insert-library-button"
-			label={ __( 'Open Design Library', i18n ) }
 			icon={ <SVGStackableIcon /> }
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
This fix only visibly hides the text from the screen. Seeing the issue, I think at the core of the problem, we probably have wrong labels (or wrong attributes or wrong html structure). If you check the Stackable settings button, it displays correctly, maybe check why that is working correctly, and the design library button isn't.

PR #2507, opened 2022-11-07
Feature/2494 horizontal scroller block

vmc08:
Feature #2494


src/block/horizontal-scroller/block.json
```diff
@@ -0,0 +1,13 @@
...
+	"title": "Horizontal Scroller",
+	"description": "Adds a horizontal scroller block.",
+	"category": "stackable",
+	"usesContext": [ "postId", "postType", "queryId" ],
+	"keywords": [
+		"Scroller"
+	],
+	"textdomain": "stackable-ultimate-gutenberg-blocks",
+	"stk-type": "essential"
+}
```
bfintal, line 13:
```suggestion
{
	"apiVersion": 2,
	"name": "stackable/horizontal-scroller",
	"title": "Horizontal Scroller",
	"description": "A slider that scrolls horizontally.",
	"category": "stackable",
	"usesContext": [ "postId", "postType", "queryId" ],
	"keywords": [
		"Slider",
		"Carousel"
	],
	"textdomain": "stackable-ultimate-gutenberg-blocks",
	"stk-type": "special"
}
```

src/block/horizontal-scroller/edit.js
```diff
@@ -0,0 +1,190 @@
...
+
+	 useGeneratedCss( props.attributes )
+
+	 const blockAlignmentClass = getAlignmentClasses( props.attributes )
+	 const { hasInnerBlocks } = useBlockContext()
+	 const [ columnProviderValue, columnTooltipClass ] = ColumnInnerBlocks.useContext()
+
+	 const blockClassNames = classnames( [
+		 className,
+		 'stk-block-horizontal-scroller editor',
```
bfintal, line 65: I think you added an `editor` class here so you can check whether the scroller is in the editor or in the frontend. From what I can see, I think you're using it for 2 things:

1. Frontend js script for the horizontal scroller in `frontend-horizontal-scroller.js`
2. Styling for the frontend only

I believe you can remove the `editor` class here and do this instead:

1. in `frontend-horizontal-scroller.js`, remove the reference to `.editor`, this script is only ran in the frontend anyway
2. In `horizontal-scroller/style.scss`, remove the `:not(.editor)`, instead, add an `editor.scss` and override the styles you don't want in the editor, e.g. display grid

src/block/horizontal-scroller/edit.js
```diff
@@ -0,0 +1,190 @@
...
+								value={ props.attributes.horizontalScrollerSnap || 'center' }
+								default="center"
+							/>
+							<AdvancedRangeControl
+								label={ __( 'Item Width', i18n ) }
+								attribute="horizontalScrollerColumnWidth"
+								responsive="all"
+								units={ [ 'px', 'em', '%' ] }
+								min={ [ 0, 0, 0 ] }
+								sliderMax={ [ 100, 1000, 1000 ] }
```
bfintal, line 114: These values seem too small or large.

What we want to achieve with our `AdvancedRangeControl` is that when you drag the handle to min and drag the handle to max, they shouldn't look too extreme / broken.

Try using max values of `[ 500, 40, 50 ]`

src/block/horizontal-scroller/edit.js
```diff
@@ -0,0 +1,190 @@
...
+								step="1"
```
bfintal, line 115: Use also different steps, because a step of `1em` is very big.

`[ 1, 0.1, 1 ]`

src/block/horizontal-scroller/edit.js
```diff
@@ -0,0 +1,190 @@
...
+								placeholder="0"
+							/>
+							<AdvancedRangeControl
+								label={ __( 'Left Offset', i18n ) }
+								attribute="horizontalScrollerLeftOffset"
+								responsive="all"
+								units={ [ 'px', 'em', '%' ] }
+								min={ [ 0, 0, 0 ] }
+								sliderMax={ [ 300, 500, 100 ] }
+								step="1"
```
bfintal, line 141: `[ 500, 40, 50 ]`
`[ 1, 0.1, 1 ]`

src/block/horizontal-scroller/edit.js
```diff
@@ -0,0 +1,190 @@
...
+				>
+					<ColumnInnerBlocks
+						providerValue={ columnProviderValue }
+						orientation="horizontal"
+						renderAppender={ false }
+						template={ [
+							[ 'stackable/column' ],
+							[ 'stackable/column' ],
+							[ 'stackable/column' ],
+							 ] }
```
bfintal, line 176: Define this array outside the function:

```js
const TEMPLATE = [
	[ 'stackable/column' ],
	[ 'stackable/column' ],
	[ 'stackable/column' ],
]
```

then:

```js
template={ TEMPLATE }
```

src/block/horizontal-scroller/frontend-horizontal-scroller.js
```diff
@@ -0,0 +1,61 @@
+/* eslint-disable @wordpress/no-global-event-listener */
+/* eslint-disable no-mixed-spaces-and-tabs */
```
bfintal, line 2: Why do we have this? Should be okay to just use tabs only.
```suggestion
```

src/block/horizontal-scroller/style.scss
```diff
@@ -0,0 +1,21 @@
+.stk-block-horizontal-scroller:not(.editor) > .stk-block-content {
```
bfintal, line 1: Instead of targeting this for the frontend `:not(.editor)`, create an `editor.scss` (this only loads in the editor) then override the styles. We want to keep all `style.scss` lean and frontend only.

bfintal (changes requested):
Please see inline comments for the initial code review.

Please check if adding a new Horizontal Scroller block (without any settings) generates CSS / a style tag when you preview the page. Ideally, adding a block shouldn't produce any CSS / styles unless you change a block attribute.

src/block/horizontal-scroller/edit.js
vmc08: @bfintal this script is also being run on the editor, this is so user can check the actual scroll snapping behavior even when on the editor.

vmc08: @bfintal for this one, I'm targeting a different element between backend vs frontend, this is to avoid applying the styles in the editor for the class `.stk-block-content` as those styles are not needed on the editor.

for the frontend script, will remove the editor selector

bfintal (changes requested):
Some issues I found during my testing:
- [ ] Remove the column resizers on the sides of columns
- [ ] Dragging the scroller using your mouse is a bit slow / laggy (and then speeds up after you release). Ideally it should be smooth. Scrolling using the mousewheel is smooth already though.

https://user-images.githubusercontent.com/1033611/214216013-e9bcb1c7-82ec-4049-8058-b0eb19cd24f2.mov

- [ ] If you set a left offset and make the block width **wide** or **center,** the contents spill over to the side (doesn't happen if the block is **none** aligned)

image

TODO:
- [ ] New block icon

PR #2514, opened 2022-11-10
fix (column-block): fix horizontal scrolling on content align

### LINKED ISSUE #2408: Column block issue - When changing Size & Spacing > Content Vertical Align settings, alignment looks weird on editor
**Describe the bug**
- When changing Block tab > Size & Spacing > Content Vertical Align settings, position change looks weird on editor. This issue is much visible if you add background and change Min. Height settings to higher than default.
- It seems that the column suddenly has a scrollbar and blocks moves position to the right

Affects:
- Block content vertical align (Block tab > Size & Spacing > Content Vertical Align)

Note: 
- Might affect different blocks with nested columns such as:
  - Columns/Container
  - Accordion
  - Feature
  - Feature Grid
  - Image Box
- For Block content vertical align, it seems to be working but have a weird alignment in editor if content vertical align settings is changed.
- Issue only happens on editor, frontend looks fine.
- Using WP 6.1 & Stackable Version 3.6.1

**To Reproduce**
1. Add a Columns/Container block or any block that has a nested column (see list above)
2. Add a block background for the Columns/Container block
3. (optional) In Block > Size & Spacing, change Min. Height to 600 or more (issue is more noticeable)
5. In Block > Size & Spacing, change Content Vertical Align to any (Align Top, Center, or Bottom)
6. Observe the behavior happening on editor/backend
7. Save page and go to frontend/preview page
8. Compare behavior happening in editor vs. expected behavior in frontend

**Expected behavior**
Check behavior happening in frontend.

**Screenshots**

**Block content vertical align**

https://user-images.githubusercontent.com/103395655/194039990-880825e0-d507-4725-b809-221b1718fd33.mp4

vmc08:
Fixes #2408

### DIFF
--- a/src/block/column/editor.scss
+++ b/src/block/column/editor.scss
@@ -4,8 +4,13 @@
 	z-index: 2;
 }
 
-.editor-styles-wrapper .stk-column-wrapper > .block-editor-inner-blocks {
-	width: 100%;
+.editor-styles-wrapper {
+	.stk-block.stk-block-column .stk-column-wrapper {
+		width: auto;
+		> .block-editor-inner-blocks {
+			width: 100%;
+		}
+	}
 }
 
 // Ensure that only directly hovered columns will show their resizer. Nested
--- a/src/block/column/style.scss
+++ b/src/block/column/style.scss
@@ -1,7 +1,7 @@
 @import "common";
 
 .stk-block.stk-block-column .stk-column-wrapper {
-	width: auto;
+	width: 100%;
 }
 .stk-block-column__content {
 	box-sizing: border-box;
--- a/src/block/column/style.scss
+++ b/src/block/column/style.scss
@@ -1,7 +1,7 @@
 @import "common";
 
 .stk-block.stk-block-column .stk-column-wrapper {
-	width: 100%;
+	width: auto;
 }
 .stk-block-column__content {
 	box-sizing: border-box;


src/block/column/style.scss
```diff
@@ -1,7 +1,7 @@
 @import "common";
 
 .stk-block.stk-block-column .stk-column-wrapper {
-	width: 100%;
+	width: auto;
```
bfintal, line 4: I'm not that comfortable with this fix because this would affect all Stackable column blocks, even in the frontend.

Since the issue in https://github.com/gambitph/Stackable/issues/2408 is only in the editor, it might be better if the fix only affects the editor

bfintal (changes requested):

src/block/column/editor.scss
```diff
@@ -4,8 +4,13 @@
 	z-index: 2;
 }
 
-.editor-styles-wrapper .stk-column-wrapper > .block-editor-inner-blocks {
-	width: 100%;
+.editor-styles-wrapper {
+	.stk-block.stk-block-column .stk-column-wrapper {
```
bfintal, line 8: @vmc08 question here, the new selector uses `.stk-block.stk-block-column`, this limits the styles to only the column block, so now the `width: 100%` might not apply to other blocks that might use `.stk-column-wrapper` (I think the Feature block or Feature Grid block uses them). Is this okay?

bfintal (changes requested):

bfintal:
Closed in favor of #2548

PR #2539, opened 2022-12-13
fix (icon): render premium icon controls

vmc08:
Fixes
* Icon controls not rendering correctly


src/block-components/icon/edit.js
```diff
@@ -50,6 +46,10 @@ export const Edit = props => {
 		defaultValue,
 	} = props
 
+	const PremiumColorControls = useMemo( () => applyFilters( 'stackable.block-component.icon.color-controls', null ), [ props ] )
+	const PremiumShapeColorControls = useMemo( () => applyFilters( 'stackable.block-component.icon.shape-color-controls', null ), [ props ] )
+	const PremiumBackgroundShapeControls = useMemo( () => applyFilters( 'stackable.block-component.icon.edit.background-shape', null ), [ props ] )
+
```
bfintal, line 52: I believe you can use `[]` as the dependencies here because the filter just returns the component to be rendered.

bfintal (changes requested):
Because of this PR, I think there are other similar parts that would need fixing as well:

I used this regex to find similar code `const Premium[^\s]+ = applyFilters`

image

src/block-components/icon/edit.js
```diff
@@ -50,6 +46,10 @@ export const Edit = props => {
 		defaultValue,
 	} = props
 
+	const PremiumColorControls = useMemo( () => applyFilters( 'stackable.block-component.icon.color-controls', null ), [ ] )
+	const PremiumShapeColorControls = useMemo( () => applyFilters( 'stackable.block-component.icon.shape-color-controls', null ), [ ] )
+	const PremiumBackgroundShapeControls = useMemo( () => applyFilters( 'stackable.block-component.icon.edit.background-shape', null ), [ ] )
+
```
bfintal, line 52:
```suggestion
	const PremiumColorControls = useMemo( () => applyFilters( 'stackable.block-component.icon.color-controls', null ), [] )
	const PremiumShapeColorControls = useMemo( () => applyFilters( 'stackable.block-component.icon.shape-color-controls', null ), [] )
	const PremiumBackgroundShapeControls = useMemo( () => applyFilters( 'stackable.block-component.icon.edit.background-shape', null ), [] )

```

PR #2541, opened 2022-12-15
fix (typography): revert dynamic content fetcher

vmc08:
Fixes:
* dynamic content inconsistency issue


bfintal (changes requested):
I don't think this is the right approach here. The hook was removed because it was triggering a lot of rerenders and was one of the major causes of slowdowns in the editor.

vmc08:
> I don't think this is the right approach here. The hook was removed because it was triggering a lot of rerenders and was one of the major causes of slowdowns in the editor.

The reason we are not getting the right value here is bc the `useSelect` was removed. The only difference between the new implementation and the older one is that the `useMemo` is used but it's memoizing the wrong value since the correct and actual value is only returned after completing a certain request (which the useSelect would be able to fetch and return the correct value later)

PR #2544, opened 2022-12-20
fix (advance-style): fix position selector

vmc08:
Fixes:
* position style selector


vmc08:
Wrong code adjustment made

vmc08:
Wrong code adjustment made

PR #2560, opened 2023-01-11
fix (php): php warnings

vmc08:
Fixes:
* Fix attempt php warning


src/css-optimize.php
```diff
@@ -262,7 +262,8 @@ public function strip_optimized_block_styles( $block_content, $block ) {
 			}
 
 			// Only do this to our blocks.
-			if ( ! empty( $block ) && is_array( $block ) && stripos( $block['blockName'], 'stackable/' ) === 0 ) {
+			$blockName = isset($block['blockName']) ? $block['blockName'] : ''
```
bfintal, line 265: @vmc08 a few standards checks here since we don't have a php linter:
- Mind the spacing between parenthesis `isset( $block['blockName'] )`
- Use snake case in php `$block_name` instead of `$blockName`
- Semi-colon at line end `;`

bfintal (changes requested):

andeng1106:
@vmc08 @bfintal I'm still seeing` strpos():` and `str_replace():` errors when I tried these:
- I activated the build from this PR.
- I added these four lines on wp-config.php
```
	define( 'WP_DEBUG', true );

	define( 'SCRIPT_DEBUG', true );

	define( 'WP_DEBUG_DISPLAY', true );

	define( 'WP_DEBUG_LOG', true );
```
- PHP version is 8.1.9. Tried versions below 8.1.9 and errors doesn't appear.
image

PR #2566, opened 2023-01-26
fix (block-div): fix max content width when center aligned

### LINKED ISSUE #2502: Max Content Width issue - Block's max. content width is not reflecting properly on frontend
**Describe the bug**
When changing max. content width for the nested columns/container block, it doesn't apply correctly on backend.

Observations:
- Saw this issue on 3.4.5 also so this happens before performance enhancements
- This might only happen for Columns/Container block only so make sure other blocks will not be affected.

**To Reproduce**
Shortcut:
1. Add a new page
2. Copy and paste these blocks on your page:
[max content width frontend issue.txt](https://github.com/gambitph/Stackable/files/9887395/max.content.width.frontend.issue.txt)
3. Save page
4. Observe editor (expected) vs. frontend view (issue)

Exact steps:
1. Add a new page
2. Add a Columns/Container block
3. Inside the nested column from Step 2, add another Columns/Container block inside
4. Add two columns and one Heading block each
   - This is how the structure would look like:
   image
5. (optional) Add background color for both Columns/Container block
6. Add two co
7. In the nested Columns/Container block, set Max. Content Width to 750px and Content Horizontal Align as center (Block tab > Size & Spacing > Max. Content Width / Content Horizontal Align)
8. Save page
9. Observe editor (expected) vs. frontend view (issue)

**Expected behavior**
Frontend view should be same with editor when adjusting Max. Content Width

**Screenshots**

https://user-images.githubusercontent.com/103395655/198563142-11f5919a-1d0d-414c-b292-76ec166a1aab.mp4

vmc08:
Fixes #2502

### DIFF
--- a/src/block-components/helpers/size/edit.js
+++ b/src/block-components/helpers/size/edit.js
@@ -7,9 +7,7 @@ import {
 	AdvancedRangeControl,
 	FourRangeControl,
 } from '~stackable/components'
-import {
-	useAttributeEditHandlers, useDeviceType,
-} from '~stackable/hooks'
+import { useAttributeEditHandlers, useDeviceType } from '~stackable/hooks'
 
 /**
  * WordPress dependencies
--- a/src/block-components/helpers/size/style.js
+++ b/src/block-components/helpers/size/style.js
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import { BlockCss } from '~stackable/components'
+import compareVersions from 'compare-versions'
 
 const Styles = props => {
 	const propsToPass = {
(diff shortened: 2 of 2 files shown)


bfintal (changes requested):
2 things:

1. This seems to be adjusting the **width** of the block, while the option is for the **max content width**. This might not produce the expected behavior on smaller screens where the block might get need to be resize smaller.

2. We do not need to conditionally render here because CSS rendering will never trigger a deprecation / block error even if newly added. (please test this)

PR #2580, opened 2023-02-08
fix (separator): prevent the separator from being infront of the blocks. 

### LINKED ISSUE #2268: Bring to Front for Top/Bottom Separator is not working properly
**Describe the bug**
Bottom separator overlaps for columns contents even if Bring to Front is toggled OFF
Top separator works fine for editor but for some blocks, it doesn't work in frontend.

Happens on blocks with separators such as:
- Call to Action - top and bottom (top doesn't work on frontend only)
- Columns/Container - bottom only
- Feature - top and bottom (top doesn't work on frontend only)
- Feature Grid - bottom only
- Hero - top and bottom (top doesn't work on frontend only)

**To Reproduce**
1. add a columns/container block or any block listed in description above
2. add contents inside the column
3. make sure container background is toggled on and has a visible background color
4. In Block indicator, toggle Bottom Separator
5. Add separator color
6. Set Height to max (400)
7. Check if it overlaps
8. Check if Bring to Front is toggled on

Additional steps to see to add Separator Layer 2:
1. If it works fine, add another separator by toggling Separator Layer 2 on
2. Add separator color
3. Set opacity to 1
4. Check if it overlaps
5. Check if Bring to Front is toggled on

**Expected behavior**
Bottom Separator should not overlap contents inside the column if Bring to Front is toggled OFF

**Screenshots**

https://user-images.githubusercontent.com/103395655/165459637-bbd3e2ff-327f-4453-87a6-545570187830.mp4

Sample from client's websites:
- Visit https://www.coursebuildershub.com/ - Overlapping happens here
image

- Visit https://www.earlyyears.tv/ - Overlapping doesn't happen here and it seems to work fine
image

prconcepcion:
Fixes #2268


src/block-components/separator/style.scss
```diff
@@ -5,8 +5,8 @@
 		.stk-inner-blocks {
 			z-index: 5;
 		}
-		> *:not(.stk-video-background) { // Don't change the order of any video backgrounds, they stay at the back.
-			z-index: 2;
+		> *:not(.stk-video-background, .stk-inner-blocks) { // Don't change the order of any video backgrounds, they stay at the back.
+			z-index: -5;
```
bfintal, line 9: You cannot do `-5` here because this can possibly hide the content behind other content that we don't control. Won't `2` work?

bfintal (changes requested):
I don't think `-5` is good here because this can possibly hide the content behind other content that we don't control. Won't `2` work?

prconcepcion:
> I don't think `-5` is good here because this can possibly hide the content behind other content that we don't control. Won't `2` work?

Tried `2` but the separators of the Hero and CTA block wouldn't go behind the container in the frontend. Updated it to `-1`.

src/block-components/separator/style.scss
```diff
@@ -5,8 +5,8 @@
 		.stk-inner-blocks {
 			z-index: 5;
 		}
-		> *:not(.stk-video-background) { // Don't change the order of any video backgrounds, they stay at the back.
-			z-index: 2;
+		> *:not(.stk-video-background, .stk-inner-blocks) { // Don't change the order of any video backgrounds, they stay at the back.
+			z-index: -1;
```
bfintal, line 9: I'd recommend doing something else here because this currently changes the z-index of all existing content. The fix should only affect the broken parts.

bfintal (changes requested):

src/block-components/separator/style.scss
```diff
@@ -5,11 +5,17 @@
...
 			z-index: 5;
 		}
-		> *:not(.stk-video-background) { // Don't change the order of any video backgrounds, they stay at the back.
+		> *:not(.stk-video-background, .stk-inner-blocks) { // Don't change the order of any video backgrounds, they stay at the back.
 			z-index: 2;
 		}
 	}
 
+	&:is(.stk-block-call-to-action, .stk-block-columns, .stk-block-feature-grid, .stk-block-feature, .stk-block-hero) {
+		> *:is(.stk-separator) { // Don't change the order of any video backgrounds, they stay at the back.
```
bfintal, line 14: Isn't this the same?
```suggestion
		> .stk-separator { // Don't change the order of any video backgrounds, they stay at the back.
```

bfintal (changes requested):
Can you try exchanging the places of the selector block of `z-index` `2` and `5`

Issue #2588, opened 2023-02-21
Fix mechanism to style blocks with modified uniqueClass


decodekult:
**Describe the bug**
In https://github.com/gambitph/Stackable/pull/2446/commits/7ec605c4fc175d3a1fd5c756053e752b994264a6 a number of filters were introduced to support custom `uniqueClass` values for blocks.

Changes include:
- a pair of `stackable.block-div.uniqueClass.edit`/`stackable.block-div.uniqueClass.save`  filters targeting the block `div` wrapper.
- a pair of `stackable.container-div.uniqueClass.edit`/`stackable.container-div.uniqueClass.save` filters targeting the block container.
- a pair of `stackable.block-styles.edit`/`stackable.block-styles.save` filters to modify the styles generated for a given block.

However, the `stackable.block-styles.save` filter has no effect over block styles. The mechanism to generate the block styles on block saving uses `addCssToCssSaveObject`, which **calls `addStyle`** on the `CssSaveCompiler` object related to the block, and returns an empty string. Thus the newly introduced filter `stackable.block-styles.save` filters nothing, just an empty string.

As a result, it becomes **impossible to match** the generated styles selector with the modified unique class for the `div` or container in a given block.

**To Reproduce**
1. Generate a callback for the `save` filters on the block `div` and container:
```
addFilter( 'stackable.block-div.uniqueClass.save', 'namespace', ( className ) => {
	return `${ className }_mod`;
} );
addFilter( 'stackable.container-div.uniqueClass.save', 'namespace', ( className ) => {
	return `${ className }_mod`;
} );
```
2. Try to apply the new unique class to the generated styles for the block:
```
addFilter( 'stackable.block-styles.save', namespace', ( css, className, attributes ) => {
	return css.replace( className, `${ className }_mod` );
} );
```
3. Add a `Heading` block to an editor, and add some styles to it (for example, set some border radius and some box shadow). Note that the styles are visible properly in the editor.
4. Save the edited page and visit it on the frontend.
5. Note that the styles are not applied to the block output, but they are indeed added to the DOM, **with the un-edited block class selector**.

**Expected behavior**
Selectors should have a way to match between the generated styles and the generated HTML nodes.

**Screenshots**
![índice](https://user-images.githubusercontent.com/1232205/220332776-ad00dd88-3707-4adc-b478-3cf6831eda28.png)
[...]

PR #2589, opened 2023-02-21
fix: filter on the block unique classname for saving styles

decodekult:
Add a way to match the modified block unique class on styles and on HTML nodes. Also, add a note on a filter since the original value will always be empty - please consider deprecating the filter altogether.

Ref: https://github.com/gambitph/Stackable/issues/2588

### DIFF
--- a/src/components/block-css/index.js
+++ b/src/components/block-css/index.js
@@ -347,6 +347,8 @@ const BlockCss = props => {
 
 	if ( ! props.editorMode ) {
 		blockUniqueClassName = applyFilters( 'stackable.block-css.uniqueClass.save', blockUniqueClassName, attributes )
+	} else {
+		blockUniqueClassName = applyFilters( 'stackable.block-css.uniqueClass.edit', blockUniqueClassName )
 	}
 
 	// Selectors can be arrays, flatten them.
--- a/src/components/block-css/index.js
+++ b/src/components/block-css/index.js
@@ -345,6 +345,10 @@ const BlockCss = props => {
 		collapsedSelector = collapsedSelector.replace( /[^^?](.%s)([^-])/g, `$1-${ instanceId }$2` )
 	}
 
+	if ( ! props.editorMode ) {
+		blockUniqueClassName = applyFilters( 'stackable.block-css.uniqueClass.save', blockUniqueClassName, attributes )
+	}
+
 	// Selectors can be arrays, flatten them.
 	if ( Array.isArray( selector ) ) {
 		selector = selector.join( ', ' )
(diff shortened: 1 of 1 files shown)


src/components/block-css/index.js
```diff
@@ -345,6 +345,10 @@ const BlockCss = props => {
 		collapsedSelector = collapsedSelector.replace( /[^^?](.%s)([^-])/g, `$1-${ instanceId }$2` )
 	}
 
+	if ( ! props.editorMode ) {
+		blockUniqueClassName = applyFilters( 'stackable.block-css.uniqueClass.save', blockUniqueClassName, attributes )
+	}
+
```
bfintal, line 351: Looks good to me!

I was thinking that it also might be a good idea to also do this for the `edit` part just for the sake of completion:

```suggestion
	blockUniqueClassName = applyFilters( `stackable.block-css.uniqueClass.${ props.editorMode ? 'edit' : 'save' }`, blockUniqueClassName, attributes )

```

decodekult: Hi Benjamin!

Yes, I noticed that there is some asymmetry between `save` and `edit` processes. I decided not to move to the edit part for some reasons:
- All the `edit`-related hooks introduced to modify unique class names only pass the filtering `className` value, never the `attributes`, since the set of attributes being passed to components executing those filters is always a subset of the block attributes, as stated [here](https://github.com/gambitph/Stackable/pull/2446#issuecomment-1370968384). If we want to have a `save` and an `edit` set of filters, I would suggest following the same schema and pass `attributes` only to the `save` version, and in fact I am pushing this change to the current pull request.
- The `edit` flow it a little more entangled. It passes over the `BlockCssEdit` component. In it, we gather the `instanceId` from the WordPress native query loop block, matching the ID of the current post (I assume), and we pass that `instanceId` to the `BlockCss` component. With that `instanceId`, the `BlockCss` component composes one unique classname by appending it to the result of a `getBlockUniqueClassname` call (note that the `BlockCss` component itself has some commented-out code to gather the `instanceId` by itself). In the end, the `BlockCss` composes the CSS output for the `edit` flow by generating an unique classname counting on the `instanceId`. **But** the `BlockCssEdit` component also applies the `stackable.block-styles.edit` filter, and passes to it the generated CSS for the editing block, and also a block unique classname that **does not include** any reference to the `instanceId` in case it exists.

In short, I feel that there is also a mismatch between unique classnames on the edit process, especially when the block is edited in the WordPress new site editor and especially when dealing with the current post context when inside a query loop, but I could not reproduce any issue related to this, so I decided to skip it and focus on the saving mechanism.

Let me know if you would like me to open that party can :-) I would rather investigate slightly deeper and open a new issue if I find anything.

bfintal:
@decodekult good points, this new change looks good to me already :)

PR #2598, opened 2023-03-07
feat (new-block): Tabs Block

### LINKED ISSUE #694: Tabs Block
Tabs Block

prconcepcion:
Fixes #694


src/block/tab-content/edit.js
```diff
@@ -0,0 +1,98 @@
...
+				blockHoverClass={ props.blockHoverClass }
+				clientId={ props.clientId }
+				attributes={ props.attributes }
+				className={ blockClassNames }
+				data-tab={ tabCount }
+			>
+				<div className="stk-block-tab-content__wrapper">
+					<ColumnInnerBlocks
+						providerValue={ columnProviderValue }
+						orientation="horizontal"
```
bfintal, line 82: The inner blocks are all vertical
```suggestion
```

src/block/tab-content/edit.js
```diff
@@ -0,0 +1,98 @@
...
+	] )
+
+	return (
+		<>
+			{ isSelected && (
+				<>
+					<InspectorTabs />
+					<Alignment.InspectorControls hasRowAlignment={ true } />
+					<BlockDiv.InspectorControls />
+					<Separator.InspectorControls />
```
bfintal, line 61: Go through all of the Block Components here and see whether you'll need them for the Tab Content block. One thing for sure is you won't need the `Separator` and `Conditional Display` here. Adjust the `save` and `style` parts accordingly.

Also go through the Block Components of the Tab Labels and Tabs block and check if they're relevant.

src/block/tab-labels/block.json
```diff
@@ -0,0 +1,15 @@
...
+	"title": "Tab Labels",
+	"description": "Container for Tabs Block labels",
+	"category": "stackable",
+	"usesContext": [ "postId", "postType", "queryId" ],
+	"parent": [
+		"stackable/tabs"
+	],
+	"keywords": [],
+	"textdomain": "stackable-ultimate-gutenberg-blocks",
+	"stk-type": "special",
```
bfintal, line 13: You'll need to set this and Tab Content to `hidden` so that the block won't show up in the enable/disable block are in the settings
```suggestion
	"stk-type": "hidden",
```

src/block/tab-labels/edit.js
```diff
@@ -0,0 +1,147 @@
...
+		const tabContentBlock = document.querySelector( `[data-block="${ tabContent }"]` )
+		const columns = tabContentBlock.querySelectorAll( '[data-type="stackable/column"]' )
+		columns.forEach( ( element, index ) => {
+			if ( dataTab === index + 1 ) {
+				element.style.display = 'flex'
+			} else {
+				element.style.display = 'none'
+			}
+		} )
+	}
```
bfintal, line 77: This implementation looks like it's written for vanilla JS for the frontend. You'll need to change this. As much as possible, we don't want to manipulate CSS directly here.

I think one approach can be to add a `className` or `data-*` attribute to the `BlockDiv` (block-level css) in `edit` in the Tab Labels block, then use some CSS in the editor to hide the corresponding Tab Content column.

src/block/tab-labels/edit.js
```diff
@@ -0,0 +1,147 @@
...
+	const {
+		insertBlock, updateBlockAttributes,
+	} = dispatch( 'core/block-editor' )
+
+	const {
+		getBlockRootClientId, getNextBlockClientId,
+	} = select( 'core/block-editor' )
+
+	const tabContent = getNextBlockClientId( props.clientId )
+	const tabsBlock = getBlockRootClientId( props.clientId )
```
bfintal, line 59: This will slow down the editor because it will traverse the block tree every render. I don't think you need to know the `tabContent` and `tabsBlock` every time, and I think you'll only need to reference these when changing/clicking on tabs. If so, then the best way would be only do these during the action itself.

src/block/tabs/block.json
```diff
@@ -0,0 +1,13 @@
+{
+	"apiVersion": 2,
+	"name": "stackable/tabs",
+	"title": "Tabs",
+	"description": "Insert your content inside tabs like a browser",
+	"category": "stackable",
+	"usesContext": [ "postId", "postType", "queryId" ],
+	"keywords": [],
```
bfintal, line 8:
```suggestion
	"keywords": [
		"toggle"
	],
```

src/block/tabs/edit.js
```diff
@@ -0,0 +1,188 @@
...
+		if ( props.attributes.tabCount ) {
+			const tabContentBlock = document.querySelector( `[data-block="${ innerBlocks[ 1 ].clientId }"]` )
+			const columns = tabContentBlock.querySelectorAll( '[data-type="stackable/column"]' )
+			columns.forEach( ( element, index ) => {
+				if ( index !== 0 ) {
+					element.style.display = 'none'
+				}
+			} )
+		}
+	}, [ props.attributes.tabCount ] )
```
bfintal, line 90: Please see my previous comment about not directly manipulating CSS aside if it's the last resort.

src/block/tabs/style.js
```diff
@@ -0,0 +1,89 @@
...
+	return (
+		<>
+			<BlockCss
+				{ ...propsToPass }
+				selector=""
+				styleRule="columnCount"
+				attrName="columns"
+				key="columns"
+				responsive="all"
+			/>
```
bfintal, line 36: I think this is unused.

src/block/tabs/edit.js
```diff
@@ -0,0 +1,188 @@
...
+						>
+							<AdvancedRangeControl
+								label={ __( 'Tabs', i18n ) }
+								min={ 1 }
+								sliderMax={ 10 }
+								attribute="tabCount"
+								onChange={ value => {
+									const tabsInnerBlocksClientIds = innerBlocks.map( ( { clientId } ) => clientId )
+									const tabsUpdatedAttributes = {}
+									const block = getBlockFromExample( 'stackable/column', {} ) // eslint-disable-line
```
bfintal, line 111: Remove this, this seems to be unused. Also, a new block should be created every time you call an `insertBlock` like what's shown here: https://github.com/gambitph/Stackable/pull/2598/files#diff-3d5143887e532f9de77c7aa181b255da8f18433e0d7f53ee82f78a3df7c9f74cR127

bfintal (changes requested):
Good start. My advice is for you to connect the 3 blocks together first, clean things up before you proceed. Right now the tabs block isn't working yet, it looks like you can't add another tab and it doesn't yet use the Tab Label block and Tab Content block.

The next steps should be to implement the following behavior:
- [ ] Adding a Tabs block should have 1x Tab Label block and 1x Tab Content block
- [ ] Adding a Tabs block should default to having 3 tabs (the Tab Content block would have 3x columns)
- [ ] Clicking on a Tab label should show the corresponding Column in the Tab Content block

Other things to consider:
- [ ] In the editor, only manipulate CSS directly as a last resort. When showing the currently selected tab, there are a number of ways to do this with React / class names / CSS
- [ ] Add a toggle similar to the Columns block "columns option" where when you slide the columns number, it can duplicate the last tab when creating new tabs

PR #2629, opened 2023-03-21
feat (global settings): add our global settings button in the site editor

bfintal:


andeng1106:
@andeng1106 To monitor:
- It doesn't change responsive view when changing responsive related settings

Issue #2651, opened 2023-04-24
Query Loop - Some Stackable blocks causes a block error when inside Query Loop > Post Template block


andeng1106:
**Describe the bug**
When adding some Stackable blocks inside the Query Loop > Post Template block, it causes a block error (not able to attempt recovery).
Blocks such as:
- Testimonial block: Error in the nested Image block
- Team Member block: Error in the nested Image block
- Image Box: Error in the nested icon block

**To Reproduce**
1. Add a new page
2. Add a Query Loop block
3. Inside the nested Post Template block from Query Loop block, add a Testimonial & Image Box block
4. See if there are block errors

**Expected behavior**
Should not have block errors when adding Stackable blocks inside Query Loop > Post Template block

**Screenshots**
**Testimonial & Team Member Block**

https://user-images.githubusercontent.com/103395655/233913824-0f712e41-b2b4-4f8b-a695-508c2ccc7948.mp4

**Image Box block**

https://user-images.githubusercontent.com/103395655/233914818-7649d04f-a8e2-402f-afdd-175fc81114cb.mp4

andeng1106:
Will close this alnog https://github.com/gambitph/Stackable/issues/2577 since this is a duplicate.

PR #2653, opened 2023-04-25
fix (query-loop): prevent some blocks from encountering a block error inside a query loop

### LINKED ISSUE #2577: Query Loop issue - Some blocks with icons encounters a block error when placed inside Post Template
**Describe the bug**
Some blocks with icons encounters a block error (not usable)  when placed inside the Post Template from Query Loop.
 
Blocks with icons that are affected:
- Image Box
- Icon Label
- Icon Box
- Notification

Some blocks with icons that are NOT affected:
- Icon
- Accordion
- Icon List
- Map
- Blockquote

**To Reproduce**
1. Add a new page
2. Add a Query Loop block
3. In the Post Template block, add the ff. blocks mentioned in the description
4. Check if the icon in the block encounters a block error

**Expected behavior**

**Screenshots**

https://user-images.githubusercontent.com/103395655/216928700-fdcf0ec2-ae81-4cca-8894-832d1ef7a95c.mp4

prconcepcion:
Fixes #2577 , #2651

### DIFF
--- a/src/components/block-css/index.js
+++ b/src/components/block-css/index.js
@@ -338,11 +338,13 @@ const BlockCss = props => {
 			blockUniqueClassName = blockUniqueClassName + `-${ instanceId }`
 		}
 
+		if ( ! Array.isArray( selector ) ) {
 		// Add instance id to classes. ( e.g. `stk-abc123` -> `stk-abc123-2`, where 2 is `instanceId`. )
-		selector = selector.replace( /[^^?](.%s)([^-])/g, `$1-${ instanceId }$2` )
-		hoverSelector = hoverSelector.replace( /[^^?](.%s)([^-])/g, `$1-${ instanceId }$2` )
-		parentHoverSelector = parentHoverSelector.replace( /[^^?](.%s)([^-])/g, `$1-${ instanceId }$2` )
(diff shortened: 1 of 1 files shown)


src/components/block-css/index.js
```diff
@@ -338,11 +338,13 @@ const BlockCss = props => {
 			blockUniqueClassName = blockUniqueClassName + `-${ instanceId }`
 		}
 
+		if ( ! Array.isArray( selector ) ) {
```
bfintal, line 341: Since the fix seems to be about fixing the error when the `replace` function is not available, maybe the `if` statement should be only when `selector` is a `string` instead of checking if it's **not** an array.

bfintal (changes requested):
Since the fix seems to be about fixing the error when the `replace` function is not available, maybe the `if` statement should be only when `selector` is a `string` instead of checking if it's **not** an array.

src/components/block-css/index.js
```diff
@@ -338,11 +338,13 @@ const BlockCss = props => {
 			blockUniqueClassName = blockUniqueClassName + `-${ instanceId }`
 		}
 
+		if ( typeof selector === 'string' || selector instanceof String ) {
```
bfintal, line 341:
```suggestion
		if ( typeof selector === 'string' ) {
```

PR #2659, opened 2023-05-17
fix (modified-indicator): fixes multiple issues

prconcepcion:


src/components/responsive-toggle/index.js
```diff
@@ -70,14 +70,22 @@ const ResponsiveToggle = props => {
...
 
-		const tabletAttributes = [ attributes[ `${ props.attribute }Tablet` ] ]
+		// const attribute = props.attribute
+
+		// console.log( props.attribute, attributes, props.suffix )
+
+		// if ( attribute === 'backgroundMedia' ) {
+		// 	ret
+		// }
+
```
bfintal, line 80: Please remove comments
```suggestion
```

src/components/four-range-control/index.js
```diff
@@ -300,6 +360,8 @@ FourRangeControl.defaultProps = {
 	placeholderLeft: '',
 	initialPosition: '',
 
+	splitMode: undefined,
```
bfintal, line 363: I feel that `splitMode` doesn't describe best what our intention is. I'd recommend using `vhMode` instead because it shows clearly from the name that it's a "vertical-horizontal" mode

src/components/font-family-control/google-fonts.json
bfintal, line 1: Please do not commit this file because it is not related to the fix. It will cause merging issues, we only update this during builds

src/components/responsive-toggle/index.js
```diff
@@ -107,6 +115,8 @@ const ResponsiveToggle = props => {
 		}
 	} )
 
+	// console.log( screens, props.attribute, responsiveValues )
+
```
bfintal, line 119: Please cleanup comments

src/components/image-control2/index.js
```diff
@@ -83,6 +83,7 @@ const ImageControl = memo( props => {
 	return (
 		<AdvancedControl
 			{ ...controlProps }
+			suffix="Url"
```
bfintal, line 86: I understand why you need this. However I don't think this is the ideal solution because if you don't know the inner workings of the `AdvancedControl` component, you will not know what the `suffix` prop is for (that it's solely used for checking whether there's a value). I recommend changing this.

Let's go back to why you need this new prop, you need it so that the control will know whether the attribute has a value or not. Usually, the `prop.attribute` that's passed here is enough, but you need need to check a different attribute for images. So why not instead of a `suffix` prop, add a new `valueCheckAttribute` prop that's a string. Then inside the `AdvancedControl` component, instead of appending a suffix to the attribute, just use `valueCheckAttribute` if it's given.

You can pass it like this:

```jsx
valueCheckAttribute={ props.attribute + 'Url' }
```

In the `defaultProps` for `AdvancedControl` make sure to add comments so we can document what this prop does.

bfintal (changes requested):
Hey @prconcepcion please do some code cleanup, and see my comments about `suffix` and `splitMode`

src/components/four-range-control/index.js
```diff
@@ -300,6 +360,8 @@ FourRangeControl.defaultProps = {
 	placeholderLeft: '',
 	initialPosition: '',
 
+	vhMode: undefined,
```
bfintal, line 363:
```suggestion
	vhMode: false,
```

src/components/base-control2/index.js
```diff
@@ -172,6 +173,8 @@ AdvancedControl.defaultProps = {
 
 	after: null,
 
+	valueCheckAttribute: '', // Checks the value of different variations of the attribute name i.e. backgroundMedia -> backgroundMediaUrl
```
bfintal, line 176:
```suggestion
	valueCheckAttribute: '', // Checks the value of different variations of the attribute name i.e. backgroundMedia -> backgroundMediaUrl. This is mainly used for the modified indicator.
```

PR #2662, opened 2023-05-18
fix: updated css for subtitle global typography

### LINKED ISSUE #1903: UI issue: question mark icon in subtitle - global typography goes to the right side
**To Reproduce**
1. adjust the subtitle options in global typography > see bug

expected behavior: it should not go to the right most side (should be beside the `subtitle` text)

Screen Shot 2021-10-15 at 3 35 39 PM

mxkae:
fixed #1903

### DIFF
--- a/src/plugins/global-settings/typography/editor.scss
+++ b/src/plugins/global-settings/typography/editor.scss
@@ -36,15 +36,16 @@
 		overflow: hidden !important;
 		white-space: nowrap !important;
 		background: transparent !important;
+		display: flex !important;
+		justify-content: flex-start !important;
 		&::before,
 		&::after {
 			display: none !important;
 		}
 	}
 	.dashicons-editor-help {
-		position: absolute;
-		top: 0;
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
It works, but we lot some spacing between the text and the tooltip. Please add a `12px` space between them.

bfintal (changes requested):
Hey @mxkae I was looking into this further, and I noticed that the new styles:

```css
display: flex !important;
gap: 12px
```

...are also affecting other selectors which are not related to the issue being fixed, namely: `.block-editor-block-list__layout`, `.block-editor-block-list__block`, and `.editor-styles-wrapper`. Including other selectors might cause side-effects in other areas.

Based on your solution of adding `display: flex`, I think you'll need to add a new entry that targets instead:

```css
.ugb-global-typography-preview {
	.editor-styles-wrapper {
		display: flex !important;
		gap: 12px;
	}
}
```

Issue #2528, opened 2022-11-28
Text block - Add support for justifying text


andeng1106:
**Is your feature request related to a problem? Please describe.**
Yes. User wants to justify text using the Text block.

**Describe the solution you'd like**
Add a feature to justify text.

More info: https://practicaltypography.com/justified-text.html

**Describe alternatives you've considered**
Adding "text-align: justify;" to the block custom CSS

PR #2665, opened 2023-05-19
feat: justify text option for text block

mxkae:
added justify text option (#2528)


bfintal (changes requested):
Good implementation. I'd prefer to have less buttons though, please see my comment.

PR #2666, opened 2023-05-19
fix: prevent redirecting when dragging columns with links

### LINKED ISSUE #2657: Horizontal Scroller block - Should be able to scroll drag horizontal scroller even if there's a column link
Yes. When dragging the horizontal scroller with column links, it redirects to the link URL instead of dragging only

Should be able to drag the horizontal scroller block even if columns have links to it.

1. Using the scrollbar feature
2. Alternative scrolling such as:
Windows: Shift + mouse scroll wheel
Mac: Swipe mousepad horizontally

mxkae:
fixes #2657

### DIFF
--- a/src/block/horizontal-scroller/frontend-horizontal-scroller.js
+++ b/src/block/horizontal-scroller/frontend-horizontal-scroller.js
@@ -11,8 +11,10 @@ class StackableHorizontalScroller {
 		let initialClientX = 0
 
 		els.forEach( el => {
+			// get all <a> tags for inner columns
 			const children = el.querySelectorAll( '.stk-block-link' )
 
+			// prevents redirecting to the inner column link
 			const onClickHandler = function( e ) {
 				e.preventDefault()
 			}
@@ -28,6 +30,7 @@ class StackableHorizontalScroller {
 
 				// Prevent selection of contents because of dragging.
 				e.preventDefault()
+				// adds the on click event listener to each inner column
 				children.forEach( child => {
 					child.addEventListener( 'click', onClickHandler )
 				 } )
@@ -53,6 +56,7 @@ class StackableHorizontalScroller {
 
 				dragTimeout = setTimeout( () => {
 					el.classList.remove( 'stk--snapping-deactivated' )
+					// removed the on click event listeners to allow redirecting to inner column links
 					children.forEach( child => {
 						child.removeEventListener( 'click', onClickHandler )
 					 } )
--- a/src/block/horizontal-scroller/frontend-horizontal-scroller.js
+++ b/src/block/horizontal-scroller/frontend-horizontal-scroller.js
@@ -11,6 +11,12 @@ class StackableHorizontalScroller {
 		let initialClientX = 0
 
 		els.forEach( el => {
+			const children = el.querySelectorAll( '.stk-block-link' )
+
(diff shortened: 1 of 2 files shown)


bfintal:
@mxkae kindly add some code comments to document this and we're good to go

bfintal (changes requested):
@mxkae kindly add some code comments to document this and we're good to go

src/block/horizontal-scroller/frontend-horizontal-scroller.js
```diff
@@ -11,6 +11,14 @@ class StackableHorizontalScroller {
 		let initialClientX = 0
 
 		els.forEach( el => {
+			// get all <a> tags for inner columns
+			const children = el.querySelectorAll( '.stk-block-link' )
```
bfintal, line 15:
```suggestion
			// Get all blocks that have links enabled.
			const children = el.querySelectorAll( '.stk-block-link' )
```

src/block/horizontal-scroller/frontend-horizontal-scroller.js
```diff
@@ -22,6 +30,10 @@ class StackableHorizontalScroller {
 
 				// Prevent selection of contents because of dragging.
 				e.preventDefault()
+				// adds the on click event listener to each inner column
+				children.forEach( child => {
+					child.addEventListener( 'click', onClickHandler )
+				 } )
```
bfintal, line 36: I'd like to suggest that for commenting this one, while we do add the click event listener here, this is already apparent in the code. It would be more helpful for other people to know why we are adding a click event listener in this step. I'd probably write something like this:

```suggestion
				// Links will trigger while dragging, this disabled the links.
				children.forEach( child => {
					child.addEventListener( 'click', onClickHandler )
				 } )
```

src/block/horizontal-scroller/frontend-horizontal-scroller.js
```diff
@@ -44,6 +56,10 @@ class StackableHorizontalScroller {
 
 				dragTimeout = setTimeout( () => {
 					el.classList.remove( 'stk--snapping-deactivated' )
+					// removed the on click event listeners to allow redirecting to inner column links
```
bfintal, line 59: You can adjust this one :)

src/block/horizontal-scroller/frontend-horizontal-scroller.js
```diff
@@ -11,6 +11,14 @@ class StackableHorizontalScroller {
 		let initialClientX = 0
 
 		els.forEach( el => {
+			// get all <a> tags for inner columns
```
bfintal, line 14: While this comment says exactly what the line does, I prefer comments that explains what the chunk of code is for, or explain what it does for complex ones. For example:

```suggestion
			// Get all links, because we will need to disable them during drag
```

src/block/horizontal-scroller/frontend-horizontal-scroller.js
bfintal, line 15: I just realized that all links would suffer from the problem, so I added `, a` in the selector

```suggestion
			const children = el.querySelectorAll( '.stk-block-link, a' )
```

bfintal (changes requested):
Hi, just some more minor comments

PR #2669, opened 2023-05-22
fix: reset button global typography removed on default value

### LINKED ISSUE #2667: Reset button in global typography does not go away
**Describe the bug**
In Stackable's Global Settings, under Global Typography, the option Apply Typography Styles to always has a reset button even though the value is the default.

**To Reproduce**
1. Edit a page
2. Open Stackable's Global Settings, under Global Typography, make sure the Apply Typography Styles is the default (first value)
3. Refresh the editor
4. Open the option again, see that the reset button is there

**Expected behavior**
There should be no reset button

**Screenshots**

https://github.com/gambitph/Stackable/assets/1033611/855ae3b0-1860-4ff3-9b6e-7f73331473a3

mxkae:
fixes #2667

### DIFF
--- a/src/plugins/global-settings/typography/index.js
+++ b/src/plugins/global-settings/typography/index.js
@@ -162,6 +162,7 @@ addFilter( 'stackable.global-settings.inspector', 'stackable/global-typography',
 					] }
 					value={ applySettingsTo }
 					onChange={ changeApplySettingsTo }
+					default={ 'blocks-stackable-native' }
 				/>
 				<ControlSeparator />
 				{ TYPOGRAPHY_TAGS.map( ( {


src/plugins/global-settings/typography/index.js
```diff
@@ -162,6 +162,7 @@ addFilter( 'stackable.global-settings.inspector', 'stackable/global-typography',
 					] }
 					value={ applySettingsTo }
 					onChange={ changeApplySettingsTo }
+					default={ 'blocks-stackable-native' }
```
bfintal, line 165: Because this is a static string, it's better if you pass it as a string instead of using an expression. This wasn't in the linter before, I just added it in `develop` now
```suggestion
					default="blocks-stackable-native"
```

bfintal (changes requested):

PR #2678, opened 2023-05-25
fix (carousel): fix multiple issues

prconcepcion:


src/block/carousel/edit.js
```diff
@@ -262,9 +261,13 @@ const Edit = props => {
...
-			clearInterval( interval )
-			sliderRef.current.parentElement.removeEventListener( 'mouseenter', onMouseEnter )
-			sliderRef.current.parentElement.removeEventListener( 'mouseleave', onMouseLeave )
+			try {
+				clearInterval( interval )
+				sliderRef.current.parentElement.removeEventListener( 'mouseenter', onMouseEnter )
+				sliderRef.current.parentElement.removeEventListener( 'mouseleave', onMouseLeave )
+			  } catch ( err ) {
+
+			  }
```
bfintal, line 270: I think the error here is from `sliderRef.current` being undefined? If it is, then I think optional chaining would be better here: 

```js
sliderRef.current?.parentElement.removeEventListener( 'mouseenter', onMouseEnter )
```

src/block/carousel/save.js
```diff
@@ -77,31 +77,33 @@ export const Save = props => {
...
+								</button>
+								<button className="stk-block-carousel__button stk-block-carousel__button__next" aria-label={ attributes.ariaLabelNext || 'Next slide' }>
+									<SvgIcon.Content
+										value={ attributes.arrowIconNext || defaultIconNext }
+									/>
+								</button>
+							</div>
+						) }
+					</div>
+					{ attributes.showDots && <div className="stk-block-carousel__dots" role="list" data-label={ attributes.ariaLabelSlide || 'Slide %%d' }></div> }
```
bfintal, line 105: Can you check if the different combinations of layout/locations for the dots and arrows work because the DOM arrangement of the dots and arrows were specifically made this way.

src/block/carousel/style.js
```diff
@@ -169,7 +169,7 @@ const Styles = props => {
 			/>
 			<BlockCss
 				{ ...propsToPass }
-				selector=".stk-block-carousel__dots"
+				selector=".%s > .stk-block-carousel__content-wrapper > .stk-block-carousel__dots"
```
bfintal, line 172: `>` is not allowed for saved styles because this will produce an error in multisite. Please revert, let's not encourage nested carousels

```suggestion
				selector=".stk-block-carousel__dots"
```

bfintal (changes requested):

PR #2682, opened 2023-05-27
fix: auto-enable hasBackground when selecting background color/image

### LINKED ISSUE #2673: When adding a background image or background color, auto-enable hasBackground
Across all block backgrounds, when the background toggle is disabled, and then a background image or background color is selected, it should enable the `hasBackground` attribute for the block.

https://github.com/gambitph/Stackable/assets/1033611/d69b318c-da0e-4963-90a5-eaa6a6100355

mxkae:
fixes #2673

### DIFF
--- a/src/block-components/block-div/edit.js
+++ b/src/block-components/block-div/edit.js
@@ -71,7 +71,7 @@ export const Edit = props => {
 					onChange={ hasBackground => setAttributes( { hasBackground } ) }
 					initialOpen={ initialOpen === 'background' }
 				>
-					<BackgroundControls attrNameTemplate="block%s" isBlockBackground={ true } />
+					<BackgroundControls attrNameTemplate="block%s" onBackgroundEnableAttribute="hasBackground" />
 				</PanelAdvancedSettings>
 				<PanelAdvancedSettings
 					title={ __( 'Borders & Shadows', i18n ) }
--- a/src/block-components/block-div/edit.js
+++ b/src/block-components/block-div/edit.js
@@ -71,7 +71,7 @@ export const Edit = props => {
 					onChange={ hasBackground => setAttributes( { hasBackground } ) }
 					initialOpen={ initialOpen === 'background' }
 				>
-					<BackgroundControls attrNameTemplate="block%s" />
+					<BackgroundControls attrNameTemplate="block%s" isBlockBackground={ true } />
 				</PanelAdvancedSettings>
 				<PanelAdvancedSettings
 					title={ __( 'Borders & Shadows', i18n ) }
--- a/src/block-components/helpers/backgrounds/edit.js
+++ b/src/block-components/helpers/backgrounds/edit.js
@@ -13,7 +13,11 @@ import {
 	AdvancedRangeControl,
 	ImageControl2,
 } from '~stackable/components'
-import { useAttributeEditHandlers, useDeviceType } from '~stackable/hooks'
+import {
+	useAttributeEditHandlers,
+	useDeviceType,
+	useBlockSetAttributesContext,
+} from '~stackable/hooks'
 import { getAttributeName, urlIsVideo } from '~stackable/util'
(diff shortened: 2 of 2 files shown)


bfintal (changes requested):
I just realized that the changes will be placed in a Block Component & a helper component, which means that other components/blocks will use this. The current changes will also turn on `hasBackground` if any background color or background image is changed.

Since what we want is for the background to turn on when Style > Background > Color / Image is changed, it might be better to do an adjustment in `src/block-components/block-div/edit.js`, and pass a new prop here:

```jsx
<BackgroundControls attrNameTemplate="block%s" />
```

..so that the Block Div component would be in charge of specifying that `hasBackground` should be the attribute that's enabled.

src/block-components/block-div/edit.js
```diff
@@ -71,7 +71,7 @@ export const Edit = props => {
 					onChange={ hasBackground => setAttributes( { hasBackground } ) }
 					initialOpen={ initialOpen === 'background' }
 				>
-					<BackgroundControls attrNameTemplate="block%s" />
+					<BackgroundControls attrNameTemplate="block%s" isBlockBackground={ true } />
```
bfintal, line 74: While this works, I'm thinking that ideally `BackgroundControls` should have no knowledge of an attribute named `hasBackground`. Also, because `isBlockBackground` isn't reusable because it's coupled tightly with `hasBackground`.

What if instead of a `isBlockBackground={ true }` prop which will enable `hasBackground` every time a background color/image is modified inside the component.. we add a `onBackgroundEnableAttribute` that would specify what attribute would be made `true` when the background color/image is modified:

```suggestion
					<BackgroundControls 
						attrNameTemplate="block%s"
						onBackgroundEnableAttribute="hasBackground"
					/>
```

src/block-components/helpers/backgrounds/edit.js
```diff
@@ -74,6 +80,18 @@ export const BackgroundControls = props => {
...
+					const attributes = {
+						[ getAttrName( 'backgroundColor' ) ]: value,
+						hasBackground: true,
+					}
+
+					const attributesNoHasBackground = Object.fromEntries(
+						Object.entries( attributes ).filter( ( [ key ] ) => key !== 'hasBackground' )
+					  )
+
+					setAttributes( props.isBlockBackground ? attributes : attributesNoHasBackground )
```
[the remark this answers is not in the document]
bfintal: My earlier suggestion was about optimization, combining that into my earlier suggestion about changing into `onBackgroundEnableAttribute`:

```suggestion
					const attributes = {
						[ getAttrName( 'backgroundColor' ) ]: value,
					}

					if ( props.onBackgroundEnableAttribute ) {
						attributes[ props.onBackgroundEnableAttribute ] = true
					}

					setAttributes( attributes )
```

Also, make sure to add `onBackgroundEnableAttribute` in `defaultProps`

src/block-components/helpers/backgrounds/edit.js
```diff
@@ -168,6 +186,29 @@ export const BackgroundControls = props => {
...
+						const attrHeightAttribute = getAttributeName( `${ getAttrName( 'backgroundMedia' ) }WidthAttribute`, deviceType )
+						const attrAlt = getAttributeName( `${ getAttrName( 'backgroundMedia' ) }Alt`, deviceType )
+
+						const attributes = {
+							[ attrNameId ]: image.id,
+							[ attrNameUrl ]: image.url,
+							[ attrWidthAttribute ]: image.width || '',
+							[ attrHeightAttribute ]: image.height || '',
+							[ attrAlt ]: image.alt || '',
+							hasBackground: true,
```
bfintal, line 202: Please apply my earlier suggestions about the new prop and about optimization:

```suggestion
```

bfintal (changes requested):

PR #2683, opened 2023-05-27
fix: save and retrieve deleted columns when changing column count

### LINKED ISSUE #2675: Columns: When typing in a number in the column count, the columns get deleted
**Describe the bug**
When typing in a number (backspace then type a number) in the column count control, the columns get deleted.

**To Reproduce**
1. Create 2 columns, add content inside each column
2. Click on the columns control, then hit backspace then type in "3"
3. See that all the columns are now blank

**Expected behavior**
The result should be as if you slid the range control to "3". The issue occurs because when you. hit backspace, all the columns are deleted, then when you enter in "3" by typing it in, new blank blocks are created. Solution can be that when backspacing, it should keep in memory the previous columns that got deleted, so when making the columns, they're not blank.

**Screenshots**
Expected:

https://github.com/gambitph/Stackable/assets/1033611/479e163b-7338-495a-9caf-d92bffe8c63e

Actual:

https://github.com/gambitph/Stackable/assets/1033611/57b0d0a5-fb4a-4d08-a1a3-d2587d6a671a

mxkae:
fixes #2675

### DIFF
--- a/src/block-components/columns/column-settings-button.js
+++ b/src/block-components/columns/column-settings-button.js
@@ -16,52 +16,46 @@ export const ColumnsControl = ( { sliderMax = 6 } ) => {
 		numInnerBlocks, innerBlocks,
 	} = useBlockContext()
 	const [ isDuplicate, setIsDuplicate ] = useLocalStorage( 'stk__columns_new_duplicate', false )
-	let debounceTimeout
 
 	const setColumns = numColumns => {
 		const { insertBlock, removeBlocks } = dispatch( 'core/block-editor' )
 
-		// prevents the previous setTimeouts from executing
-		clearTimeout( debounceTimeout )
-		// the timeout allows the user to finish typing before making changes to the number of columns
-		debounceTimeout = setTimeout( () => {
-			// do nothing if input field is blank
-			if ( numColumns === '' ) {
+		// do nothing if input field is blank
+		if ( numColumns === '' ) {
 
-				// Remove the columns.
-			} else if ( numColumns < numInnerBlocks ) {
-				const columnClientIds = innerBlocks.slice( numColumns ).map( ( { clientId } ) => clientId )
-				removeBlocks( columnClientIds, false )
+			// Remove the columns.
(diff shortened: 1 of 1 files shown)


src/block-components/columns/column-settings-button.js
```diff
@@ -16,6 +18,8 @@ export const ColumnsControl = ( { sliderMax = 6 } ) => {
 		numInnerBlocks, innerBlocks,
 	} = useBlockContext()
 	const [ isDuplicate, setIsDuplicate ] = useLocalStorage( 'stk__columns_new_duplicate', false )
+	const [ localNumInnerBlocks, setLocalNumInnerBlocks ] = useLocalStorage( 'stk__columns_num_inner_blocks' )
+	const [ localInnerBlocks, setLocalInnerBlocks ] = useLocalStorage( 'stk__columns_inner_blocks' )
```
bfintal, line 22: You don't need to use `useLocalStorage` here, this can be just `useState` because you don't need this to persist

bfintal (changes requested):
Seeing this PR, I'd like to suggest another approach that might be easier and might be more ideal. If you go back to the original cause issue, the columns are all getting deleted because by hitting `backspace` on the number field, it's behaving like you're setting the columns to `0`. Then if you type in a number, let's say `3`, the value now becomes `03` because the backspace / empty value is treated as a `0`.

What if we change the approach here, instead of saving and bringing back the inner blocks, what if you make it so that if the input field is blank `""` (no value because you hit `backspace` on it), then it won't do anything - it won't treat it as a `0` and delete columns?

If you do this, then the columns won't get deleted in the first place and there might be no more need for saving and retrieving them.

mxkae:
I added a timeout because I noticed that if the number of columns is initially set to a single digit, for example, 4, and you want to change it to a double-digit like 12, then the number of columns will be like from 4 -> 1 -> 12, making the 2nd-4th columns get deleted. With the timeout, the number of columns will be like from 4 -> 12.

PR #2694, opened 2023-06-13
fix: Block max content width

### LINKED ISSUE #2502: Max Content Width issue - Block's max. content width is not reflecting properly on frontend

mxkae:
fixes #2502

### DIFF
--- a/src/block/columns/style.scss
+++ b/src/block/columns/style.scss
@@ -8,6 +8,10 @@
 	max-width: none;
 }
 
+.stk-inner-blocks > .stk-block {
+	width: inherit;
+}
+
 .stk-block-columns > .stk-block-content {
 	--stk-column-gap: 0px; // For nested columns, this takes precedence.
 	column-gap: var(--stk-column-gap, 0px);


bfintal (changes requested):
Does the bug only happen in nested columns like that's described in the issue? If it so, then maybe the fix should only be applied only in that scenario.

I'm afraid the current fix (adding a style to `.stk-inner-blocks > .stk-block`) will affect all our blocks and might have unforeseen consequences.

Please refactor to fix only the specific scenario where the issue is seen: so far the issue says that it happens on a columns > inner column > columns nesting. are there other scenarios?

mxkae:
I checked if the bug only happens on nested columns and turns out it also happens to other (inner) blocks.

Expected output (Editor view):
Screenshot 2023-06-13 at 6 40 19 PM

Frontend view (w/o fix):
Screenshot 2023-06-13 at 6 40 33 PM

Frontend view (w/ fix):
Screenshot 2023-06-13 at 6 40 53 PM

src/block-components/helpers/size/style.js
```diff
@@ -53,6 +53,20 @@ const Styles = props => {
...
+				selector={ wrapperSelector || selector }
+				styleRule="width"
+				attrName="width"
+				key="width"
+				attrNameTemplate={ attrNameTemplate }
+				responsive="all"
+				hasUnits="px"
+				valueCallback={ ( value, getAttribute ) => {
+					const blockWidth = getAttribute( 'width' )
+					return blockWidth ? 'inherit' : ''
```
bfintal, line 67: Please use `undefined` here, I think there's a chance that the style might show up if it's an empty string.
```suggestion
					return blockWidth ? 'inherit' : undefined
```

bfintal (changes requested):
Found an issue: if you perform the issue in tablet/mobile only, then the issue still persists. There's a 3rd argument `device` that you can use:

```js
valueCallback={ ( value, getAttribute, device ) => {
```

src/block-components/helpers/size/style.js
```diff
@@ -53,6 +53,20 @@ const Styles = props => {
...
 			/>
+			<BlockCss
+				{ ...propsToPass }
+				selector={ wrapperSelector || selector }
+				styleRule="width"
+				attrName="width"
+				key="width"
+				attrNameTemplate={ attrNameTemplate }
+				responsive="all"
+				hasUnits="px"
```
bfintal, line 64:
```suggestion
```

andeng1106:
The issue has been fixed but it affected single blocks.

Remaining issue:
- Single blocks looks different outside the columns when there's a Max. Content Width added (issue only in editor)

https://github.com/gambitph/Stackable/assets/103395655/6f421544-b089-4bc8-b45d-d4db2f78a915

@mxkae

PR #2695, opened 2023-06-13
fix: Blocksy Theme Button and Header Block links

### LINKED ISSUE #2387: [Blocksy] Button & Heading block - No underline for links in frontend
**Describe the bug**
For buttons: When selecting Style > Link for buttons, there is no underline in frontend compared to editor.
For text block links: When I set a text block as "Show as Link", it doesn't appear as link (no underline)
Note: 
- This issue only happens for Blocksy theme
- This issue happens for blocks such as:
  -  Button block > Link style
  - Heading block

**To Reproduce**
1. Change theme to Blocksy

- For Button block:
   2. Add a button block
   3. Change button style to "Link" (Inspector > Style > Styles > Link)

- For Heading block:
   2. Add a Heading block
   3. Add a URL for that link (using toolbar or dynamic content)

4. Save page
5. Check editor if there's an underline (looks like a link; expected appearance)
6. Check frontend if there's no underline

**Expected behavior**
There should be an underline in frontend button if Button style is set to "Link"

**Screenshots**
**Button block**
https://user-images.githubusercontent.com/103395655/184845225-8097e62c-9e15-41b3-a9e2-94d48e8d5ee4.mp4

**Heading block**
https://user-images.githubusercontent.com/103395655/211230096-ef8db3af-9902-440d-906d-7cf8b2e29041.mp4

mxkae:
fixes #2387

### DIFF
--- a/src/block/button/style.scss
+++ b/src/block/button/style.scss
@@ -5,6 +5,10 @@
 	--stk-block-margin-bottom: 0;
 }
 
+.stk-link {
+	text-decoration: underline;
+}
+
 // This makes it so that when the button becomes too small for the viewport,
 // that's the time we wrap text. If the button is fullwidth, and other
 // non-fullwidth button presses against it, this stops the button text from
--- a/src/block/heading/style.scss
+++ b/src/block/heading/style.scss
@@ -10,3 +10,7 @@
 		margin: 0 !important;
 	}
 }
+
+.stk-block-heading__text a {
+	text-decoration: underline;
+}


bfintal (changes requested):
While these newly added CSS will directly fix the issue, I think these might not be the correct fix.

If we add the `text-decoration: underline` ourselves, this means that all websites that use our link buttons and just any link inside a heading block would get an underline. However, I think the real issue is when we are used with Blocksy, there should be an underline for links but because of our styles, the underline is gone.

I imagine that the fix should be to figure out why the underline is being overwritten. The "plain" style in the button should make the button look like a normal link - so if the theme's normal link has an underline, then our button should also. Same goes with heading links, if the theme normally places an underline on heading links, then ours should have one too.

bfintal:
Let's not proceed with this

PR #2696, opened 2023-06-13
fix: Progress bar max to full width

### LINKED ISSUE #2519: [WP 6.2] Progress Bar block - When setting block to full width, it doesn't max to full width in frontend
**Describe the bug**
- This happens when I use the Twenty Twenty Three Theme
- Issue only in frontend

**To Reproduce**
1. Activate Twenty Twenty Three theme
2. Add a new page
3. Add a Progress Bar block
4. (optional) Add a block background
5. Set block to full width
6. Check editor (expected)
7. Check frontend (issue)

**Expected behavior**
Should max to full width on frontend.

**Screenshots**

https://user-images.githubusercontent.com/103395655/201816133-510b99ef-13db-49de-ad60-7138f415e697.mp4

mxkae:
fixes #2519

### DIFF
--- a/src/block/progress-bar/style.scss
+++ b/src/block/progress-bar/style.scss
@@ -1,5 +1,9 @@
 @import "common";
 
+.stk-block-progress-bar:not(.alignfull) {
+	width: 100%;
+}
+
 .stk-progress-bar {
 	position: relative;
 	width: var(--progress-bar-width, 100%);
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
This fix will affect all existing progress bars. Checking the issue, it says the issue only happens if you set the block to fullwidth, so I would suggest adding the fix to the block when it's in fullwidth so that the existing working behavior is not modified.

Also, it's said that it happens in the Twenty Twenty Three theme, please check if your new fix works in that theme and in other themes as well

PR #2699, opened 2023-06-14
Made the placeholder texts of the Carousel block translatable

DAnn2012:
Made the placeholder texts of the Carousel block translatable

### DIFF
--- a/src/block/carousel/edit.js
+++ b/src/block/carousel/edit.js
@@ -564,7 +564,7 @@ const Edit = props => {
 									__( 'Previous slide', i18n )
 								) }
 								attribute="ariaLabelPrev"
-								placeholder="Previous slide"
+								placeholder={ __( 'Previous slide', i18n ) }
 							/>
 							<AdvancedTextControl
 								label={ sprintf(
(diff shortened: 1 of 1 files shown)


bfintal:
Hey @DAnn2012 I see what what you mean here, but for this case, the placeholders were made English by default and untranslatable intentionally. Mainly because the accessibility values are saved with the rendered block, so we can't dynamically change them to match the language settings of the website because the block will produce an error.

If the placeholders were made translatable, when the language changes, the placeholders would change to match the languages, but in reality the accessibility values would still remain the same - so this may cause confusion because the default values didn't change even though the placeholders show it.

DAnn2012:
Hello @bfintal 

Thank you for the answer and explanation.

I thought I'd submit the change, because I was inspired by the Button block where the placeholder is translatable:

https://github.com/gambitph/Stackable/blob/2588bc34cbeba23e266c530851af574fc22de1c1/src/block/button/edit.js#L144

I'm sorry for the trouble.

bfintal:
No worries @DAnn2012! Closing this now :)

PR #2705, opened 2023-06-22
feat: new UI skin

mxkae:
- updated UI colors 
- updated image control


bfintal (changes requested):
Please do these changes:
- [ ] Add dark border on disabled toggles
- [ ] Update the style control (check the button block's style control for reference)
- [ ] Change the background color of the top part of the inspector to `#F5F5F5`
- [ ] Disabled toolbar buttons/labels are too light
- [ ] Make sure the skin colors are only applied to our components/controls, add `stk-*` classes if needed
- [ ] Update image control to add preview of image and change the text to **Replace** if an image/video is present

bfintal (changes requested):
Some comments about the image picker:
1. Please move the image preview before the upload/replace button
2. For the upload button, since it now looks like a regular button, try and use the `Button` component: `import Button from '../button'`. Currently, the active state of the button doesn't show correctly

PR #2718, opened 2023-07-04
fix (new ui skin): bug fixes

mxkae:


src/styles/editor-skin.scss
```diff
@@ -161,4 +163,8 @@
 	.block-editor-list-view-leaf.is-selected td {
 		background-color: var(--stk-skin-dark);
 	}
+
+	.ugb-global-settings-typography-control .components-base-control__label {
+		overflow: hidden;
+	}
```
bfintal, line 169: This feels like this should belong to the global typography component and not in the editor skin

bfintal (changes requested):
Just one minor change @mxkae

PR #2720, opened 2023-07-04
feat (carousel): infinite looping and slide speed

bfintal:
TODO:
- [ ] Add a new block attribute and setting to adjust the Slide Speed - this should add a `data-speed` attribute, defaults to `500` ms (only supported by slide type)
- [ ] Add a new block attribute and setting to toggle on/off the infinite scroll - this should add a class `stk--is-infinite-slide` if toggled on, defaults to `false`(only supported by slide type)
- [ ] The infinite scroll currently only supports if the carousel shows 1 slide at a time, will need to make it support having 2 or more slides at a time.


bfintal:
closed in favor of #3060

PR #2731, opened 2023-07-13
Removed placeholder text from getting-started.js

DAnn2012:
Removed placeholder text from getting-started.js


DAnn2012:
Hello @bfintal 

I would like to point out that, of course, the placeholder text string was also included in the .POT file.

Thanks.

PR #2732, opened 2023-07-13
Removed WordPress versions prior to 6.0 in php-5.6.39.yml, php-7.3.5.yml and php-8.0.0.yml

DAnn2012:
Hello @bfintal 

Removed WordPress versions prior to 6.0 in php-5.6.39.yml, php-7.3.5.yml and php-8.0.0.yml

Also, I take this opportunity to point out that [Node.js version 14.x is "end-of-life" as of April 30, 2023](https://github.com/nodejs/release#end-of-life-releases).


bfintal:
Just to add, these tests are failing right now, but will get into those in the future :|

PR #2751, opened 2023-07-18
fix (inner-column): prevent margin bottom visual indicator from overlapping with the block

### LINKED ISSUE #2672: Template Locking changes block spacing
**Describe the bug**
If template locking the contents of a column, the spacing between blocks change.

**To Reproduce**
1. Add a columns block
2. Add 2 text/heading blocks inside the column
3. Select the Inner columns block, from the 3-dots in the toolbar, lock the block (apply to all children)
4. See that the block spacing changes

**Expected behavior**
The block spacing should not change

**Screenshots**

Image

https://github.com/gambitph/Stackable/assets/103395655/55e6b380-1ea0-4926-be3e-3c10b23fef33

prconcepcion:
fixes #2672

### DIFF
--- a/src/styles/editor-block.scss
+++ b/src/styles/editor-block.scss
@@ -237,9 +237,6 @@
 	// Last of type since there's a .block-list-appender
 	> [data-block]:nth-last-child(2) {
 		margin-bottom: 0;
-		&[data-type^="stackable/"] > .stk-block {
-			margin-bottom: 0;
-		}
 	}
 	> [data-block]:first-child:last-child {
 		margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


bfintal (commented):
@prconcepcion is there a way to only target when template locking is used? I think a fix as general as this would have unwanted side effects.

src/styles/editor-block.scss
```diff
@@ -235,7 +235,7 @@
 		margin-top: 0;
 	}
 	// Last of type since there's a .block-list-appender
-	> [data-block]:nth-last-child(2) {
+	&:is([data-is-drop-zone="true"]) > [data-block]:nth-last-child(2) {
```
bfintal, line 238:
```suggestion
	// If the block parent is locked, there will be no appender block and data-is-drop-zone will not be present.
	&:is([data-is-drop-zone="true"]) > [data-block]:nth-last-child(2) {
```

PR #2757, opened 2023-07-20
fix (accordion): collapsed settings are now properly applied

### LINKED ISSUE #2723: Accordion block - Colors not working for Collapsed state in Accordion
**Describe the bug**
When adding colors in Collapsed state for Accordion block, it's not working for both editor and frontend

**To Reproduce**
1. Add a new page
2. Add an Accordion block
3. Try out the scenarios below (Screenshots part)
4. Make sure to check both frontend and editor
5. See issues

**Expected behavior**
Collapsed state color should work for all blocks in Accordion

**Screenshots**
Some scenarios in the video:
1. Added Collapsed state color for Icon Label's Heading & Icon - not working for both frontend and editor
7. Added Collapsed state color for Inner Column 2's Text block - working
8. Added an Image block inside Inner Column 2 and added overlay color - not working properly for both frontend and editor, might have a conflict with hover state color

https://github.com/gambitph/Stackable/assets/103395655/a5239e99-00c8-46db-9ab7-a3d31d032cfd

prconcepcion:
fixes #2723

### DIFF
--- a/src/components/block-css/index.js
+++ b/src/components/block-css/index.js
@@ -292,9 +292,9 @@ const BlockCss = props => {
 	// TODO: why do we have this condition for the collapsedSelector, but they just do the same prepending??
 	if ( hasCollapsed ) {
 		if ( blockState === 'collapsed' ) {
-			collapsedSelector = prependClass( selector, ':where(.stk-block-accordion.stk--is-open) .%s, .%s.stk--is-open' )
+			collapsedSelector = prependClass( selector, ':where(.stk-block-accordion.stk--is-open) .%s' )
 		} else {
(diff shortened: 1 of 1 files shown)


src/components/block-css/index.js
```diff
@@ -292,9 +292,9 @@ const BlockCss = props => {
 	// TODO: why do we have this condition for the collapsedSelector, but they just do the same prepending??
 	if ( hasCollapsed ) {
 		if ( blockState === 'collapsed' ) {
```
prconcepcion, line 294: It also goes in here when you click on the inner column when the accordion is open @bfintal

bfintal: I think that's expected because this is called during style generation, the collapsed styles are always generated regardless whether the accordion is collapsed or not

PR #2775, opened 2023-07-24
fixes (columns): columns content max width/horizontal align saves upon refresh

### LINKED ISSUE #2742: Columns block - Content Max. Width & Content Horizontal Align settings resets if editor is refreshed
**Describe the bug**
- Issue only on editor
- This issue happens for settings in:
   - Layout > Content Max. Width
   - Layout > Content Horizontal Align
Screenshot 2023-07-18 at 11 28 20 AM

**To Reproduce**
1. Add a Columns block
2. In settings mentioned above, change both settings
3. Refresh editor
4. See that both settings resets

**Expected behavior**
Content Max. Width & Content Horizontal Align settings should not reset if editor is refreshed

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/ba6ca045-9a31-4c9f-9c78-9a4824adeb1b

prconcepcion:
fixes #2742

### DIFF
--- a/src/block-components/alignment/editor.scss
+++ b/src/block-components/alignment/editor.scss
@@ -37,7 +37,3 @@
 		}
 	}
 }
-
-.stk-block-columns .block-editor-block-list__layout {
-	width: 100%;
-}
--- a/src/block-components/alignment/editor.scss
+++ b/src/block-components/alignment/editor.scss
@@ -37,3 +37,7 @@
 		}
 	}
 }
+
+.stk-block-columns .block-editor-block-list__layout {
+	width: 100%;
+}
--- a/src/block/columns/style.js
+++ b/src/block/columns/style.js
@@ -10,10 +10,9 @@ import {
 	Separator,
 	Transform,
 	Columns,
-	ContainerDiv,
 } from '~stackable/block-components'
 import { useBlockAttributesContext } from '~stackable/hooks'
-import { BlockCssCompiler } from '~stackable/components'
+import { BlockCssCompiler, BlockCss } from '~stackable/components'
 
 /**
  * WordPress dependencies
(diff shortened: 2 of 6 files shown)


src/block-components/alignment/editor.scss
```diff
@@ -37,3 +37,7 @@
 		}
 	}
 }
+
+.stk-block-columns .block-editor-block-list__layout {
+	width: 100%;
+}
```
bfintal, line 43: What is this for? This will affect a lot of our blocks, I would suggest not having this general selector

src/block/columns/style.js
```diff
@@ -39,10 +119,10 @@ const BlockStyles = memo( props => {
 
 	return (
 		<>
+			<Styles { ...props } />
 			<Alignment.Style { ...props } { ...alignmentOptions } />
 			<BlockDiv.Style { ...props } />
 			<MarginBottom.Style { ...props } />
-			<ContainerDiv.Style { ...props } { ...containerDivOptions } />
```
bfintal, line 45: I feel that we're using some styling from the Container Div (maybe in the layout tab, something with the size/max width/spacing). Unless you are sure that we do not use this, and if it interferes with the new styles, then I think you can retain this.

src/block/columns/style.js
```diff
@@ -25,10 +24,91 @@ const alignmentOptions = {
...
+					if ( value === 'flex-start' || value === 'center' ) {
+						return 'auto'
+					}
+					if ( value === 'flex-end' ) {
+						return '0'
+					}
+				} }
+			/>
+		</>
+	)
```
bfintal, line 111: I think these are the same styles being applied by `ContainerDiv`?

bfintal (changes requested):

prconcepcion:
@bfintal, regarding the inner columns, the content maximum width and container maximum width control share the same value when adjusting it. Is that the expected behavior?

andeng1106:
@prconcepcion Issue still happens:

https://github.com/gambitph/Stackable/assets/103395655/58fd4fd6-5011-4de3-8b08-4f311acee9bc

PR #2781, opened 2023-07-28
fix (lightbox): prevent lightbox from not working if "srcset" has a different variation

### LINKED ISSUE #2724: Lightbox is not working if WP Rocket's Lazy Loading feature is enabled
**Describe the bug**
After disabling the lazy loading, lightbox began working again.

**Expected behavior**
Lightbox should work whether WP Rocket's Lazy Loading is enabled or disabled.

prconcepcion:
fixes #2724

### DIFF
--- a/src/lightbox/frontend-image-lightbox.js
+++ b/src/lightbox/frontend-image-lightbox.js
@@ -15,9 +15,10 @@ const getImageFromSrcset = el => {
 		}
 
 		// Find an attribute that ends with 'srcset'
-		return el.getAttributeNames().find( v => {
+		const srcsetName = el.getAttributeNames().find( v => {
 			return v.endsWith( 'srcset' )
 		} )
+		return el.getAttribute( srcsetName )
 	}
 
 	const srcset = findSrcsetAttribute()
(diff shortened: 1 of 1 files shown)


src/lightbox/frontend-image-lightbox.js
```diff
@@ -7,11 +7,12 @@ const getImageFromSrcset = el => {
...
-	const srcset = el.getAttribute( 'nitro-lazy-srcset' ) ||
-		// Support for Perfmatters lazy loading.
-		el.getAttribute( 'data-srcset' ) ||
-		el.getAttribute( 'srcset' )
+	// Search and return an attribute that ends with "srcset"
+	const srcsetName = el.getAttributeNames().find( v => {
+		return v.endsWith( 'srcset' )
+	} )
+
+	const srcset = el.getAttribute( srcsetName )
```
bfintal, line 15: Please try and use the exact `srcset` attribute first, and then if that's not available, resort to looking through the attribute names for a close match

bfintal (changes requested):

PR #2783, opened 2023-07-29
Fix: i18n missing in three strings of color-palette-control/index.js

DAnn2012:
Hello @bfintal Do these three strings need to be translated?

Fix: i18n missing in three strings of color-palette-control/index.js

Update src/components/color-palette-control/index.js


bfintal:
Hey @DAnn2012, no need for translation for these since we're using the strings (and their translated counterparts) used by Gutenberg for matching. Hopefully we get proper ids or slugs for the color categories that we can use, until then we need to resort to name matching.

DAnn2012:
Thank you for the answer and explanation.

I'm sorry for the trouble.

bfintal:
@DAnn2012 No worries!

PR #2785, opened 2023-08-01
fix (WP6.3): added styling to prevent design library from being misaligned 

### LINKED ISSUE #2752: [WP 6.3] Design Library UI issues
**Describe the bug**
Issues:
- Scrolling in the Category section
- Navigation Menu lines
- Add Designs button is missing
   - This button should show up if I'm selecting two or more designs

**Expected**

https://github.com/gambitph/Stackable/assets/103395655/eb872a21-ba98-40ad-b3ae-86566a666910

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/bb32896d-1e49-4991-b560-2ac7e816f767

prconcepcion:
fixes #2752


bfintal (changes requested):
The side-panel also scrolls with the rest of the content, they should have their own scrolling area.

Also, the X button is too close to the side

https://github.com/gambitph/Stackable/assets/1033611/317b01c9-da71-4e8e-8e95-837d331c394a

bfintal (changes requested):
The top bar (the bar with the select button should stay fixed at the top).

Please go through the different functionality and UI of the design library first to make sure that everything works okay before re-submitting for code review.

src/components/modal-design-library/editor.scss
```diff
@@ -57,7 +61,7 @@
 		display: flex;
 		justify-content: flex-end;
 		padding: 0 24px;
-		height: 80px;
+		height: 50px;
```
bfintal, line 64: I suggest bringing this back because the toolbar became smaller
```suggestion
		height: 80px;
```

bfintal (changes requested):
Aside from the inline comment, I found that the "no designs found" is not centered

image

PR #2800, opened 2023-08-03
 fix: timeline block issues

mxkae:


bfintal (changes requested):
@mxkae I also added `dependencies` in the `BlockCss`, please see my other comment also. Once you fix that, please set this issue for **QA**

PR #2810, opened 2023-08-07
fix: removes extra divider in top toolbar

### LINKED ISSUE #2803: [WP 6.3] Top Toolbar - there's an extra divider when Stackable is active
In Stackable Premium, when the top toolbar enabled, selected a native paragraph block, and Stackable is **deactivated** (expected behavior):

Image

When Stackable is **activated**:

Image

mxkae:
fixes #2803

### DIFF
--- a/src/editor.scss
+++ b/src/editor.scss
@@ -63,6 +63,11 @@ svg.stk-stackable-icon-gradient {
 	}
 }
 
+// removes extra divider in Top Toolbar
+.components-toolbar.block-editor-block-toolbar__block-controls::after {
+	content: none !important;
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
I don't think this is the right fix for it. If you check out the other toolbar buttons, they don't have this extra divider. I think it might have something to do with how the button component was used or added.

This is low priority though because it only happens when the top toolbar feature is enabled, re-visit this after all the other WordPress 6.3 issues are done

bfintal:
Closed in favor of #2816

Issue #2774, opened 2023-07-24
[WP 6.3] Group > Row or Stack - When copying then pasting the Row or Stack block, the Dimensions set in Stackable blocks resets


andeng1106:
**Describe the bug**
- Issue happens for both Row & Stack
- Issue applies to Dimensions Width (Row) or Height (Stack)
- Issue happens when using Dimensions Fill or Fixed
- Issue happens when using Stackable blocks but when using Native blocks, it displays as expected

**To Reproduce**
1. Add a Group block and change it to Row or Stack
2. Inside the block, add two Stackable Heading blocks
3. In List View, select Heading 1
4. Go to Styles tab
<img width="279" alt="Screenshot 2023-07-24 at 4 45 26 PM" src="https://github.com/gambitph/Stackable/assets/103395655/d47d83a4-eac0-490f-8c91-feae007e0c18">

5. Change Dimension Width (Row) or Height (Stack) to Fill or Fixed (try 1000 px)
6. Try copying the whole Row or Stack block along with the Heading blocks
7. Paste the block
8. Go to Step 3-4
9. See that the Dimensions resets to Fit

**Expected behavior**
Dimensions should not reset upon copy pasting block.
See example for Native Heading blocks:

https://github.com/gambitph/Stackable/assets/103395655/a2eedd90-aa89-46cb-beec-4a216d649e84

**Screenshots**
_**Row:**_

https://github.com/gambitph/Stackable/assets/103395655/88f12b78-01e0-4303-bb58-2164e7597878

_**Stack:**_

https://github.com/gambitph/Stackable/assets/103395655/f753c01f-4cb1-4bb5-8d29-36afcb0369d9

Issue #2773, opened 2023-07-24
[WP 6.3] Group > Row or Stack - When setting Dimension to Fill or Fixed, the preview is not same for editor vs. frontend


andeng1106:
**Describe the bug**
- Issue in frontend only
- Issue applies to Dimensions Width (Row) or Height (Stack)
- Issue happens when using Dimensions Fill or Fixed
- Issue happens when using Stackable blocks but when using Native blocks, it displays as expected

**To Reproduce**
1. Add a Group block and change it to Row or Stack
2. Inside the block, add two Stackable Heading blocks
3. Add background (optional but issue will be more noticeable)
4. Make the Row or Stack block width to Full Width (optional but issue will be more noticeable)
5. In List View, select Heading 1
6. Go to Styles tab
<img width="279" alt="Screenshot 2023-07-24 at 4 45 26 PM" src="https://github.com/gambitph/Stackable/assets/103395655/d47d83a4-eac0-490f-8c91-feae007e0c18">

7. Change Dimension Width (Row) or Height (Stack) to Fill or Fixed (try 1000 px)
8. Save changes
9. Compare how the blocks looks like editor vs. frontend

**Expected behavior**
Frontend display/preview should be same with editor.
See example when using Native Heading blocks:

https://github.com/gambitph/Stackable/assets/103395655/06f89950-d6d0-412e-a2db-1097eb84032d

**Screenshots**
**_Row (changing Dimension Width to Fill):_**

https://github.com/gambitph/Stackable/assets/103395655/1eedd835-0de8-43e6-a028-893105f7ed73

_**Stack (changing Dimension Height to Fixed with 1000px):**_

https://github.com/gambitph/Stackable/assets/103395655/1273b5a0-fe49-48b6-b664-d64636af0246

PR #2811, opened 2023-08-07
fix: group > row or stack - dimensions

mxkae:
temporary fix for #2773 and #2774

### DIFF
--- a/src/block/icon-label/index.js
+++ b/src/block/icon-label/index.js
@@ -23,7 +23,8 @@ export const settings = {
 	supports: {
 		anchor: true,
 		align: [ 'center', 'wide', 'full' ],
-		stkBlockLinking: false, // Disable linking on the columns in this block.
+		stkBlockLinking: false,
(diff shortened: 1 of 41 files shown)


src/block/icon-label/index.js
```diff
@@ -23,7 +23,8 @@ export const settings = {
 	supports: {
 		anchor: true,
 		align: [ 'center', 'wide', 'full' ],
-		stkBlockLinking: false, // Disable linking on the columns in this block.
+		stkBlockLinking: false,
```
bfintal, line 26:
```suggestion
		stkBlockLinking: false, // Disable linking on the columns in this block.
```

andeng1106:
@mxkae @bfintal Remaining issue:
- Fix for this PR is not applied for all blocks except for Posts block:

https://github.com/gambitph/Stackable/assets/103395655/5c550a64-680d-4284-8ed0-ee0aaf287848

PR #2814, opened 2023-08-07
fix: add has polyfill for alignment

### LINKED ISSUE #2704: Hero block - `.stk-container:has(>.stk--column-flex)` is missing for Firefox browser
**Describe the bug**
- Issue for Firefox browser only
- Fix might affect other blocks

**To Reproduce**
1. Add a new page
2. Add a Hero block
3. Save page
4. Go to frontend
5. Right click and click Inspect
6. In Styles, search "has" in Filter field
7. See all results and check if the class mentioned above is visible (for Chrome)
8. Repeat steps 4-7 for Firefox
9. See that the class is missing

**Expected behavior**
- See screenshot (Google Chrome): the class is visible

https://github.com/gambitph/Stackable/assets/103395655/24361081-f50c-409c-bf32-f08af8d7a1f6

**Screenshots**
- Firefox: the class is missing

https://github.com/gambitph/Stackable/assets/103395655/68a843ec-7708-4f9f-8e1a-6fc3624a6316

**Desktop:**
 - Browser: Firefox
 - Browser Version: 114.0.1 (64-bit)

### LINKED ISSUE #2784: Alignment: Add compatibility for browsers that do not support :has
The CSS in here uses `:has`, this pseudo selector isn't supported in Firefox and Safari <= 15.3:

https://github.com/gambitph/Stackable/blob/ebc09fa4b8a46e8b0cc1931e06353dffb7628e52/src/block-components/alignment/style.scss#L8-L11

I suggest to create some JavaScript for the frontend that will add a custom class name to the element so that the necessary style are included, and new CSS rules to support it. The JS file should only be loaded if the browser is Firefox or Safari <= 15.3

For loading the JS conditionally depending on the browser, the user agent can be used to check similar to https://github.com/gambitph/Stackable/blob/7a7a556f46dea10a98eec8b4dbfe4f088bd0ea8d/src/block/accordion/index.php#L41

In the JS, be sure to support dynamically added DOM elements, use MutationObserver to check for this.

Create a JS file `src/block-components/alignment/frontend-has-polyfill.js` and add your JS polyfill there. Then to compile the script into the `dist/` folder, refer to:
- https://github.com/gambitph/Stackable/blob/develop/.config/webpack.config.dev.js#L172
- https://github.com/gambitph/Stackable/blob/develop/.config/webpack.config.prod.js#L157

mxkae:
fixes #2784, fixes #2704

### DIFF
--- a/src/block-components/alignment/frontend-has-polyfill.js
+++ b/src/block-components/alignment/frontend-has-polyfill.js
@@ -8,11 +8,9 @@ class StackableAlignmentPolyfill {
 		const modifyCSS = () => {
 			const containers = document.querySelectorAll( '.stk-container' )
 			containers.forEach( container => {
-				const columnFlex = container.querySelector( '.stk--column-flex' )
+				const columnFlex = container.querySelector( ':scope > .stk--column-flex' )
 				if ( columnFlex ) {
-					container.classList.add( 'stk-container--has-child-column-flex' )
-				} else if ( ! columnFlex && container.classList.contains( 'stk-container--has-child-column-flex' ) ) {
-					container.classList.remove( 'stk-container--has-child-column-flex' )
+					container.classList.add( 'stk-container--has-child-column-flex-polyfill' )
 				}
 			} )
 
@@ -20,9 +18,7 @@ class StackableAlignmentPolyfill {
 			blocks.forEach( block => {
 				const hasMargin = block.querySelector( ':scope > .stk--block-margin-top-auto, :scope > .stk--block-margin-bottom-auto' )
 				if ( hasMargin ) {
-					block.classList.add( 'stk--height-100' )
-				} else if ( ! hasMargin && block.classList.contains( 'stk--height-100' ) ) {
-					block.classList.remove( 'stk--height-100' )
+					block.classList.add( 'stk--height-100-polyfill' )
 				}
 			} )
 		}
--- a/src/block-components/alignment/frontend-has-polyfill.js
+++ b/src/block-components/alignment/frontend-has-polyfill.js
@@ -3,7 +3,7 @@
  */
 import domReady from '@wordpress/dom-ready'
 
-class StackableAlignment {
+class StackableAlignmentPolyfill {
 	init = () => {
 		const modifyCSS = () => {
 			const containers = document.querySelectorAll( '.stk-container' )
@@ -38,5 +38,5 @@ class StackableAlignment {
 	}
 }
 
-window.stackableAlignment = new StackableAlignment()
-domReady( window.stackableAlignment.init )
+window.stackableAlignmentPolyfill = new StackableAlignmentPolyfill()
+domReady( window.stackableAlignmentPolyfill.init )
--- a/src/block-components/alignment/index.php
+++ b/src/block-components/alignment/index.php
@@ -21,7 +21,7 @@ if ( ! function_exists( 'stackable_load_alignment_frontend_polyfill_script' ) ) 
 		$load_polyfill = false;
 
 		// Safari <= 15.3
-		if ( stripos( $user_agent, 'Safari/' ) !== false ) {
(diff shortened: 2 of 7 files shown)


src/block-components/alignment/frontend-has-polyfill.js
```diff
@@ -0,0 +1,42 @@
...
+ * WordPress dependencies
+ */
+import domReady from '@wordpress/dom-ready'
+
+class StackableAlignmentPolyfill {
+	init = () => {
+		const modifyCSS = () => {
+			const containers = document.querySelectorAll( '.stk-container' )
+			containers.forEach( container => {
+				const columnFlex = container.querySelector( '.stk--column-flex' )
```
bfintal, line 11: Based on the selector we're trying to mimic, we need this to be for the direct descendent
```suggestion
				const columnFlex = container.querySelector( ':scope > .stk--column-flex' )
```

src/block-components/alignment/frontend-has-polyfill.js
```diff
@@ -0,0 +1,42 @@
...
+				if ( columnFlex ) {
+					container.classList.add( 'stk-container--has-child-column-flex' )
+				} else if ( ! columnFlex && container.classList.contains( 'stk-container--has-child-column-flex' ) ) {
+					container.classList.remove( 'stk-container--has-child-column-flex' )
```
bfintal, line 15: Do you still need this?

bfintal (changes requested):
I took the liberty of changing some file/script/class names to better reflect the purpose of the newly introduced script.

I have some remaining suggestions / improvements:

1. We can have better names for the new classes that can signify that they are polyfill classes:
- `stk--height-100` ➡️ `stk--height-100-polyfill`
- `stk-container--has-child-column-flex` ➡️ `stk-container--has-child-column-flex-polyfill`

2. Bring back the specificity of the original selectors you're trying to polyfill, because if not the styles might not be applied correctly.

You're mainly trying to add a class that would have the same effect as the following selectors. But these 2 sets of selectors have different specificities (you can check here: https://specificity.keegan.st/)

- `:is(.stk-block-content, .stk-inner-blocks):not(.stk--column-flex):has(> :is(.stk--block-margin-top-auto, .stk--block-margin-bottom-auto))` ➡️ 3
- `.stk-container:has(> .stk--column-flex)` ➡️ 2

The new selectors are just:

- `.stk--height-100` ➡️ 1
- `.stk-container--has-child-column-flex` ➡️ 1

That means the styles might get easily overridden by other selectors. You'll need to mimic / bump up the specificity to the same level.

andeng1106:
@mxkae @bfintal Fixes are not working in editor for Firefox browser only

some features & scenarios tried:
* innerBlockJustify setting
* innerBlockAlign setting (Inner Column)
* in the inner column block, align all blocks to the left, then the last child block make margin left auto - the last child block should go to the right

bfintal (changes requested):
I feel it's odd that `Chrome/` is part of the detection for Safari. It should be `Safari/` and `Version/` and the number after version should be <= `15.3`

src/block-components/alignment/index.php
```diff
@@ -0,0 +1,59 @@
...
+		$user_agent = ! empty( $_SERVER['HTTP_USER_AGENT'] ) ? $_SERVER['HTTP_USER_AGENT'] : '';
+
+		if ( ! $user_agent ) {
+			return;
+		}
+
+		$load_polyfill = false;
+
+		// Safari <= 15.3
+		if ( stripos( $user_agent, 'Version/' ) && stripos( $user_agent, 'Safari/' ) !== false ) {
```
bfintal, line 24:
```suggestion
		if ( stripos( $user_agent, 'Version/' ) !== false && stripos( $user_agent, 'Safari/' ) !== false ) {
```

PR #2815, opened 2023-08-07
fix: reactDom.render error message

### LINKED ISSUE #2786: Resolve ReactDOM.render warnings
When you edit a post, you'll see these errors in the browser console:

> Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead

If you can't see the error, please use WP 6.2.x

Please check this reference for how to fix the issue: https://make.wordpress.org/core/2023/03/07/upgrading-to-react-18-and-common-pitfalls-of-concurrent-mode/

mxkae:
fixes #2786

### DIFF
--- a/src/components/custom-attributes-control/index.js
+++ b/src/components/custom-attributes-control/index.js
@@ -10,7 +10,7 @@ import { escape as _escape } from 'lodash'
 import {
 	Fragment,
 	useState,
-	render,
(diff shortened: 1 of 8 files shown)


bfintal (commented):
@mxkae please also submit a PR that does this for the premium codebase

PR #2819, opened 2023-08-08
fix (WP6.3): prevent PHP error from appearing

### LINKED ISSUE #2753: [WP 6.3] If `WP_DEBUG` is set to `true`, Deprecated: strpos(): Passing null to parameter #1 ($haystack) of type string… error appears
**Describe the bug**
- Errors shows only in site editor
- Happens only for Twenty Twenty Three Theme

**To Reproduce**
1. In local site folder, go to app > public > wp-config.php
2. Add `define( 'WP_DEBUG', true ); ` (below `/* Add any custom values between this line and the "stop editing" line. */` comment)
3. Save file
4. Go back to site
5. Make sure theme is Twenty Twenty Three
6. View home page
7. See errors

**Screenshots**
- Based on the error, it seems to be related with Dynamic Content

https://github.com/gambitph/Stackable/assets/103395655/d562adc1-3822-47bb-93c2-0d9ba6d6f18d

prconcepcion:
fixes #2753

### DIFF
--- a/src/deprecated/v2/init.php
+++ b/src/deprecated/v2/init.php
@@ -312,6 +312,10 @@ if ( ! function_exists( 'stackable_add_required_block_styles_v2' ) ) {
 if ( ! function_exists( 'load_frontend_scripts_conditionally_v2') ) {
 
 	function load_frontend_scripts_conditionally_v2( $block_content, $block ) {
+		if ( $block_content === null ) {
+			$block_content = '';
+		}
+
 		$block_name = isset( $block['blockName'] ) ? $block['blockName'] : '';
 		if (
 			stripos( $block_name, 'ugb/' ) === 0 ||
@@ -364,6 +368,10 @@ if ( ! function_exists( 'stackable_frontend_v2_try_migration' ) ) {
 	 * @return string The block content
 	 */
 	function stackable_frontend_v2_try_migration( $block_content, $block ) {
+		if ( $block_content === null ) {
+			$block_content = '';
+		}
+
 		$block_name = isset( $block['blockName'] ) ? $block['blockName'] : '';
 		if (
 			stripos( $block_name, 'ugb/' ) === 0 ||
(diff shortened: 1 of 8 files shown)


src/fonts.php
```diff
@@ -19,6 +19,10 @@ function __construct() {
 		}
 
 		public function gather_google_fonts( $block_content, $block ) {
+			if ( $block_content === null ) {
+				return $block_content;
+			}
+
```
bfintal, line 25: Please remove, the block_content isn't actually used here, might cause fonts to stop loading

src/init.php
```diff
@@ -188,6 +188,10 @@ public function load_frontend_scripts_conditionally_head() {
 		 * @return string output block
 		 */
 		public function load_frontend_scripts_conditionally( $block_content, $block ) {
+			if ( $block_content === null ) {
+				return $block_content;
+			}
+
```
bfintal, line 194: Same as above, please just handle when block_content is null because this will skip some processes

bfintal (changes requested):

src/deprecated/v2/init.php
```diff
@@ -364,6 +368,10 @@ function register_frontend_blog_posts_block_compatibility_v2() {
 	 * @return string The block content
 	 */
 	function stackable_frontend_v2_try_migration( $block_content, $block ) {
+		if ( $block_content === null ) {
+			$block_content = '';
```
bfintal, line 372:
```suggestion
			return $block_content;
```

src/deprecated/v2/init.php
```diff
@@ -312,6 +312,10 @@ function stackable_add_required_block_styles_v2() {
 if ( ! function_exists( 'load_frontend_scripts_conditionally_v2') ) {
 
 	function load_frontend_scripts_conditionally_v2( $block_content, $block ) {
+		if ( $block_content === null ) {
+			$block_content = '';
```
bfintal, line 316:
```suggestion
			return $block_content;
```

src/deprecated/v2/optimization-settings.php
```diff
@@ -85,6 +85,10 @@ public function disable_frontend_scripts() {
 		 * @since 2.17.0
 		 */
 		public function load_frontend_scripts_conditionally( $block_content, $block ) {
+			if ( $block_content === null ) {
+				$block_content = '';
```
bfintal, line 89:
```suggestion
				return $block_content;
```

bfintal (changes requested):

PR #2823, opened 2023-08-09
fix (block-background): prevent links from being only clickable within the content of a block

### LINKED ISSUE #2743: When a Block Background image is assigned for blocks with link, link is only clickable in the inner part/container
**Describe the bug**
- Applies to:
   - Inner Column
   - Call to Action
   - Card
   - Icon Box
   - Pricing Box
   - Team Member
   - Testimonial

**To Reproduce**
1. Add a new page
2. Add any block mentioned above
3. Add a link for the block (Advanced tab > Link)
4. Add a block background image
6. Move mouse pointer/hover over the block in the background with image
7. Observe that link is not clickable
8. Hover over the inner container
9. See that it is now clickable

Additional steps:
10. Remove block background image
11. Hover mouse over the block with background
12. See that it is clickable

**Expected behavior**
If there's a Block Background image for blocks with link, link should be clickable in the surrounding area (background image part)

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/59516b12-8bba-48dc-abbc-da192056704f

prconcepcion:
fixes #2743

### DIFF
--- a/src/components/div/style.scss
+++ b/src/components/div/style.scss
@@ -14,7 +14,11 @@
 		bottom: 0;
 		z-index: 2;
 	}
-	> * {
+
+	// Prevents blocks with links from being contained within the content
+	> *
+	:not(.stk-block-column__content, .stk-block-team-member__content, .stk-block-card .stk-container, .stk-block-testimonial__content,
+	.stk-block-pricing-box__content, .stk-block-call-to-action__content) {
 		position: relative;
 		z-index: 3;
 	}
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
I feel that this might not be the ideal fix for this because the issue is related to links (in the advanced tab), if the fix is CSS related, I think the fix should include links / `.stk-block-link`

bfintal:
Closed in favor of #2851

PR #2848, opened 2023-08-17
fix: remove spacing support if wp < 6.3

### LINKED ISSUE #2828: Native Styles settings should only be visible for Stackable blocks if inside Group block
**Describe the bug**
- Applies to all Stackable blocks (whether nested or not)
- Issue happens on WP 6.2
- This setting should not be visible if Stackable block is not in a Group block
Screenshot 2023-08-10 at 2 44 14 PM

**To Reproduce**
1. Add a new page
2. Add any Stackable block (e.g. Columns block)
3. Go to Inspector and see that the Native Styles settings is present

**Expected behavior**
Remove Native Style settings for Stackable blocks if not inside a Group block

**Screenshots**
image

### LINKED ISSUE #2837: Remove native spacing support when WordPress version is < 6.3
In WP 6.2, if you add a block (e.g. heading block), you'll see the native spacings tabs in the inspector:

Image

This panel should not be here in WP 6.2 and below. To fix this, the `spacing: true` entry in `supports`  should be removed for WP <= 6.2 across all Stackable blocks.

We have a JS filter `stackable.block.metadata` which can be used to override things in the block metadata before it's registered.

mxkae:
fixes #2837 
fixes #2828

### DIFF
--- a/src/compatibility/index.js
+++ b/src/compatibility/index.js
@@ -1,1 +1,2 @@
 import './kadence-theme'
+import './wp-6-2'
--- a/src/compatibility/wp-6-2/index.js
+++ b/src/compatibility/wp-6-2/index.js
@@ -1,0 +1,18 @@
+
+/**
+ * Internal dependncies
+ */
+import { semverCompare } from '~stackable/util'
+import { wpVersion } from 'stackable'
+
+/**
+ * WordPress dependencies
+ */
+import { addFilter } from '@wordpress/hooks'
+
+addFilter( 'stackable.block.metadata', 'stackable/wp-6-2', settings => {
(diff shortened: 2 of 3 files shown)


bfintal (changes requested):
I recommend moving this code into a new directory & file in `src/compatibility/wp-6-2/index.js` (note that you'll need to import it in `src/compatibility/index.js` for it to be included in the build.

The new file can have:

```js
addFilter( 'stackable.block.metadata', 'stackable/wp-6-2', settings => {
	if ( wpVersion && semverCompare( wpVersion, '<', '6.3' ) && settings?.supports?.spacing ) {
		delete settings.supports.spacing
	}
	return settings
} )
```

It would be better in my opinion to split it off because this is some old compatibility code that isn't part of the main functionality.

PR #2850, opened 2023-08-17
fix (link/light-box): if user enters a google map like link, show a hint.

### LINKED ISSUE #2807: Help the user if they're trying to add a Google Map in a lightbox wrong
Right now users are failing to embed a Google Map in a lightbox because the wrong URL is inputted. This issue will help them successfully embed a Google Map.

We need a way to detect whether a user is inputting a Google Map link (not an embed link) in a lightbox opened via link:

Image

Detect whether the user is inputting a Google Map URL (you get this by visiting maps.google.com, browsing, and getting the current URL). Samples of urls to detect:

```
https://www.google.com/maps/place/Teatro+Petruzzelli/@41.123542,16.872706,16z/data=!4m6!3m5!1s0x1347e8614094a4df:0xdb54262eb24c2715!8m2!3d41.1235424!4d16.8727057!16s%2Fm%2F0c414xg

https://www.google.com/maps/@14.6243308,121.0842807,16z?entry=ttu
```

When detected, let's display this note near the link input:

> Displaying a Google Map? Use the embed iframe URL instead. Need help finding it?

The last sentence should link to our docs that contains instructions on where to find the correct URL: https://docs.wpstackable.com/article/528-how-to-add-a-google-map-in-a-lightbox?utm_source=inspector&utm_campaign=learnmore&utm_medium=gutenberg

We can follow the look of hints here: [`src/components/inspector-tabs/hints.js`](https://github.com/gambitph/Stackable/blob/ebc09fa4b8a46e8b0cc1931e06353dffb7628e52/src/components/inspector-tabs/hints.js)

prconcepcion:
fixes #2807

### DIFF
--- a/src/block-components/helpers/link/edit.js
+++ b/src/block-components/helpers/link/edit.js
@@ -11,6 +11,10 @@ import {
  */
 import { __ } from '@wordpress/i18n'
 import { useAttributeEditHandlers } from '~stackable/hooks'
+import { useEffect, useState } from '@wordpress/element'
+
+// @see https://stackoverflow.com/questions/14729191/google-maps-link-validation-using-regex
+const GoogleMapRegex = /^https?\:\/\/(www\.)?google\.(com|[a-z]{1,3})\/maps\b/gm
 
 export const LinkControls = props => {
 	const {
@@ -24,6 +28,21 @@ export const LinkControls = props => {
 		getAttribute,
 		updateAttributeHandler,
 	} = useAttributeEditHandlers( props.attrNameTemplate )
+
+	const url = getAttribute( 'url' )
+	const lightbox = getAttribute( 'hasLightbox' )
+
+	const [ showGoogleMapHint, setShowGoogleMapHint ] = useState( false )
+
+	useEffect( () => {
+		if ( hasLightbox ) {
+			if ( GoogleMapRegex.test( url ) && lightbox ) {
+				setShowGoogleMapHint( true )
+			} else {
+				setShowGoogleMapHint( false )
+			}
+		}
+	}, [ url, lightbox ] )
 
 	return (
 		<>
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):

src/block-components/helpers/link/edit.js
```diff
@@ -25,6 +25,10 @@ export const LinkControls = props => {
 		updateAttributeHandler,
 	} = useAttributeEditHandlers( props.attrNameTemplate )
 
+	const url = getAttribute( 'url' )
```
bfintal, line 28:
```suggestion
	const url = getAttribute( 'url' ) || ''
```

src/block-components/helpers/link/edit.js
```diff
@@ -39,14 +43,26 @@ export const LinkControls = props => {
...
+					<AdvancedToggleControl
+						label={ __( 'Open Link in Lightbox', i18n ) }
+						help={ lightboxHelp }
+						checked={ getAttribute( 'hasLightbox' ) }
+						onChange={ updateAttributeHandler( 'hasLightbox' ) }
+					/>
+					{ showGoogleMapHint && (
+						<div className="stk-inspector-hint stk-inspector-hint__google-map ">
+							<span>
+								{ __( 'Displaying a Google Map? Use the embed iframe URL instead. Need help finding it? ', i18n ) }
```
bfintal, line 57:
```suggestion
								{ __( 'Displaying a Google Map? Use the embed iframe URL instead. Need help finding it?', i18n ) }
```

src/block-components/helpers/link/edit.js
```diff
@@ -39,14 +43,26 @@ export const LinkControls = props => {
...
+					<AdvancedToggleControl
+						label={ __( 'Open Link in Lightbox', i18n ) }
+						help={ lightboxHelp }
+						checked={ getAttribute( 'hasLightbox' ) }
+						onChange={ updateAttributeHandler( 'hasLightbox' ) }
+					/>
+					{ showGoogleMapHint && (
+						<div className="stk-inspector-hint stk-inspector-hint__google-map ">
+							<span>
+								{ __( 'Displaying a Google Map? Use the embed iframe URL instead. Need help finding it?', i18n ) }
```
bfintal, line 57:
```suggestion
								{ __( 'Displaying a Google Map in a Lightbox? Use the embed iframe URL instead. Need help finding it?', i18n ) }
```

src/components/inspector-tabs/editor.scss
```diff
@@ -14,3 +14,11 @@
...
 	}
 }
+
+.stk-inspector-hint__google-map {
+	margin: 10px 0 10px !important;
+	span {
+		font-size: 12px;
+		padding: 6px !important;
+	}
+}
```
bfintal, line 24: I see some unnecessary styling here. Please do not change font sizes and other spacing unless necessary - for example, there is no need to change the padding to `6px` because our UI is using `16px` already, just add the missing padding. And the `font-size` is already provided by the UI, no need to change it. I suggest removing this specific styling for this google map hint and instead handle things for the hint element itself:

```suggestion
```

I see you're trying to add padding to the right because this doesn't have a close button, it's better to handle this generally: add this to the `span`

```css
.stk-inspector-hint {
	// Without a close button.
	span:last-child {
		padding: 16px;
	}
}
```

Since you're changing the margin because the default has a large top margin, then you can specify the margins to be different when placed beside panels vs when inside controls: adjust the css for hints to be

```css
.stk-inspector-hint {
	margin: 16px;
}
.components-panel__body + .stk-inspector-hint {
	margin-top: 32px;
}
```

bfintal (changes requested):
Please check out my comments, afterwards, please set this for QA

andeng1106:
@prconcepcion @bfintal 

Message now displays but I only have a few clarifications and one minor correction:
Clarification:
- If I input the correct embed link, should the display message disappear?
- Is this expected to have a spacing on the left side of the message, like it’s not aligned with the “Support links to images..” text and toggle buttons?
Screenshot 2023-08-18 at 11 37 13 AM


Correction:
- Message - space between "?" and "Check" word is enclosed in the `<a>` tag instead of with the text inside the  `<span>`
Screenshot 2023-08-18 at 11 48 35 AM

PR #2860, opened 2023-08-24
fix (columns): reset responsive columns width values when you reduce the number of inner columns to 1

### LINKED ISSUE #2792: Column Width resizing improvement when removing an Inner Column
Yes. Column width doesn't reset if I remove a column.
Currently if I set column 1 to 25% and column 2 to 75% and if I removed column 2, column 1 width doesn't reset.
- Applies when removing columns through:
   - List View / Content Area
   - Modifying No. of Columns in inspector
- Applies to all responsive views

https://github.com/gambitph/Stackable/assets/103395655/f430e5aa-aef9-4204-a0dc-b9107995bb3e

If you have 2 columns, then you manually set tablet / mobile column widths to be different e.g. 25% & 75%, if you delete the 2nd column, the 1st column's tablet/mobile column widths should reset.

You have to manually change/reset column width first before removing a column.

prconcepcion:
fixes #2792


prconcepcion:
@bfintal would it be better to check if it has responsive values first and then we remove?

bfintal (changes requested):
Please check the `ResizableColumn` component, it already performs the resetting of column widths when a column was added / deleted. It would be better to implement the reset there instead of in the columns. You might have to change the implementation.

PR #2865, opened 2023-08-25
fix: alignment compatibility in editor for firefox browser

### LINKED ISSUE #2824: Flexbox columns settings not working as expected in editor for Firefox browser
- Related to https://github.com/gambitph/Stackable/issues/2784
- Fix is not working in editor **for Firefox browser only**
  - some features & scenarios tried:
    - innerBlockJustify setting
    - innerBlockAlign setting (Inner Column)
    - in the inner column block, align all blocks to the left, then the last child block make margin left auto - the last child block should go to the right

mxkae:
fixes #2824


src/block-components/alignment/edit.js
```diff
@@ -346,6 +348,18 @@ export const Edit = props => {
 	)
 }
 
+// Add additional classes when browser is Firefox to fix alignment
+addFilter( 'stackable.block-components.block-div.classnames', 'alignment-editor-has-polyfill', classes => {
+	const userAgent = navigator.userAgent
+	if ( userAgent.indexOf( 'Firefox' ) !== -1 ) {
```
bfintal, line 354: You can move this outside the filter so you can do the `addFilter` only when we're using Firefox. Also make sure that this doesn't error when `navigator` is not present

src/block-components/alignment/editor-has-polyfill.js
```diff
@@ -0,0 +1,46 @@
+/**
+ * External dependencies
+ */
+import { useBlockContext } from '~stackable/hooks'
+
+import { select } from '@wordpress/data'
+import { useBlockEditContext } from '@wordpress/block-editor'
+
+export const EditorHasPolyfill = () => {
```
bfintal, line 9: The case of this one is usually used for Components, use `editorHasPolyfill` instead

src/block-components/alignment/edit.js
```diff
@@ -346,6 +348,18 @@ export const Edit = props => {
 	)
 }
 
+// Add additional classes when browser is Firefox to fix alignment
```
bfintal, line 351: You can move this code into `src/block-components/alignment/editor-has-polyfill.js` and then in `edit.js` you can just do an

```js
import './editor-has-polyfill'
```

src/block-components/alignment/edit.js
```diff
@@ -346,6 +348,18 @@ export const Edit = props => {
...
+addFilter( 'stackable.block-components.block-div.classnames', 'alignment-editor-has-polyfill', classes => {
+	const userAgent = navigator.userAgent
+	if ( userAgent.indexOf( 'Firefox' ) !== -1 ) {
+		const alignmentHasPolyfill = EditorHasPolyfill()
+		const classesToAdd = Object.keys( alignmentHasPolyfill ).filter( classname => alignmentHasPolyfill[ classname ] )
+		const newClasses = classes.concat( classesToAdd )
+		return newClasses
```
bfintal, line 358: If you want to add classes returned by this filter, you can just append things in the argument array and then return that:

```js
return [
    ...classes,
    ...newClasses,
]
```

or 

```js
classes.push( newClass )
return classes
```

bfintal (changes requested):
1. I feel that splitting the code into 2 (1 that adds the filter, and 1 that creates what classes to add) gives us more complex code.

You can do this instead (no need for `classesToAdd`):

```js
addFilter( 'stackable.block-components.block-div.classnames', 'alignment-editor-has-polyfill', classes => {
    if ( something ) {
        classes.push( 'stk--child-has-margin-top-auto' )
    }
    return classes
} )
```

2. Another thing, why do we have a lot of different added editor styles? don't we need to just add another selector near the top of `src/block-components/alignment/editor.scss`?

src/block-components/alignment/editor.scss
```diff
@@ -1,9 +1,48 @@
...
+
+
+// this needs to be on a separate css block because firefox ignores any CSS blocks that include the :has selector
+// reference: https://connect.mozilla.org/t5/ideas/when-is-has-css-selector-going-to-be-fully-implemented-in/idc-p/26839/highlight/true#M19086
 // If an direct inner block has a top/bottom margin-auto, then we need to set
 // the height to full (similar to .stk--column-flex) so the margin auto would
 // look like it took effect..
-:is(.stk-inner-blocks, .stk-block-content):not(.stk--column-flex):has(> .block-editor-inner-blocks > .block-editor-block-list__layout > .wp-block > :is(.stk--block-margin-top-auto, .stk--block-margin-bottom-auto)),
-.stk--align-last-block-to-bottom {
+:is(.stk-inner-blocks, .stk-block-content):not(.stk--column-flex):has(> .block-editor-inner-blocks > .block-editor-block-list__layout > .wp-block > :is(.stk--block-margin-top-auto, .stk--block-margin-bottom-auto)) {
```
bfintal, line 45:
```suggestion
:is(.stk-inner-blocks, .stk-block-content):not(.stk--column-flex):has(> .block-editor-inner-blocks > .block-editor-block-list__layout > .wp-block > :is(.stk--block-margin-top-auto,
.stk--block-margin-bottom-auto)) {
```

src/block-components/alignment/editor.scss
```diff
@@ -1,9 +1,49 @@
...
+
+// this needs to be on a separate css block because firefox ignores any CSS blocks that include the :has selector
+// reference: https://connect.mozilla.org/t5/ideas/when-is-has-css-selector-going-to-be-fully-implemented-in/idc-p/26839/highlight/true#M19086
 // If an direct inner block has a top/bottom margin-auto, then we need to set
 // the height to full (similar to .stk--column-flex) so the margin auto would
 // look like it took effect..
-:is(.stk-inner-blocks, .stk-block-content):not(.stk--column-flex):has(> .block-editor-inner-blocks > .block-editor-block-list__layout > .wp-block > :is(.stk--block-margin-top-auto, .stk--block-margin-bottom-auto)),
-.stk--align-last-block-to-bottom {
+:is(.stk-inner-blocks, .stk-block-content):not(.stk--column-flex):has(> .block-editor-inner-blocks > .block-editor-block-list__layout > .wp-block > :is(.stk--block-margin-top-auto,
+.stk--block-margin-bottom-auto)) {
```
bfintal, line 46:
```suggestion
:is(.stk-inner-blocks, .stk-block-content):not(.stk--column-flex):has(> .block-editor-inner-blocks > .block-editor-block-list__layout > .wp-block > :is(.stk--block-margin-top-auto, .stk--block-margin-bottom-auto)) {
```

src/block-components/alignment/editor.scss
```diff
@@ -1,9 +1,48 @@
 .stk--column-flex,
+.stk--align-last-block-to-bottom,
+// for firefox :has polyfill
+.stk--alignment-polyfill > :is(.stk-inner-blocks, .stk-block-content):not(.stk--column-flex) {
```
bfintal, line 4: The two sets of CSS seem identical, I think you can just add the Firefox polyfill selector to the original CSS we had:

```css
.stk--column-flex,
// If an direct inner block has a top/bottom margin-auto, then we need to set
// the height to full (similar to .stk--column-flex) so the margin auto would
// look like it took effect..
:is(.stk-inner-blocks, .stk-block-content):not(.stk--column-flex):has(> .block-editor-inner-blocks > .block-editor-block-list__layout > .wp-block > :is(.stk--block-margin-top-auto, .stk--block-margin-bottom-auto)),
.stk--align-last-block-to-bottom,
// for firefox :has polyfill
.stk--alignment-polyfill > :is(.stk-inner-blocks, .stk-block-content):not(.stk--column-flex) {
	height: auto; // Reverse the 100% height from style.scss
	// ...other css
}
```

src/block-components/alignment/index.php
```diff
@@ -9,51 +9,29 @@
...
-	function stackable_load_alignment_frontend_polyfill_script() {
-
-		$user_agent = ! empty( $_SERVER['HTTP_USER_AGENT'] ) ? $_SERVER['HTTP_USER_AGENT'] : '';
-
-		if ( ! $user_agent ) {
-			return;
+$user_agent = ! empty( $_SERVER['HTTP_USER_AGENT'] ) ? $_SERVER['HTTP_USER_AGENT'] : '';
+if ( ! empty( $user_agent ) && stripos( $user_agent, 'Firefox/' ) !== false ) {
+	add_filter( 'render_block', function( $block_content, $block ) {
+		$tag = new WP_HTML_Tag_Processor( $block_content );
```
bfintal, line 15: You'll have to check for the existence of `WP_HTML_Tag_Processor` because it was just recently introduced in WP 6.2

```suggestion
		if ( ! class_exists( 'WP_HTML_Tag_Processor' ) ) {
			return $block_content;
		}
		$tag = new WP_HTML_Tag_Processor( $block_content );
```

src/block-components/alignment/index.php
```diff
@@ -9,51 +9,29 @@
...
-if ( ! function_exists( 'stackable_load_alignment_frontend_polyfill_script' ) ) {
-	function stackable_load_alignment_frontend_polyfill_script() {
-
-		$user_agent = ! empty( $_SERVER['HTTP_USER_AGENT'] ) ? $_SERVER['HTTP_USER_AGENT'] : '';
-
-		if ( ! $user_agent ) {
-			return;
+$user_agent = ! empty( $_SERVER['HTTP_USER_AGENT'] ) ? $_SERVER['HTTP_USER_AGENT'] : '';
+if ( ! empty( $user_agent ) && stripos( $user_agent, 'Firefox/' ) !== false ) {
+	add_filter( 'render_block', function( $block_content, $block ) {
```
bfintal, line 14: Can you turn this into a named function instead of an anonymous one and also check for existence? This is so that just incase you have 2 Stackable plugins active (sometimes happens to people), we won't get some odd behaviors.

```suggestion
	if ( ! function_exists( 'stackable_render_block_alignment_frontend_polyfill' ) ) {
		add_filter( 'render_block', 'stackable_render_block_alignment_frontend_polyfill', 10, 2 );
```

src/block-components/block-div/firefox-has-polyfill.js
```diff
@@ -0,0 +1,33 @@
+
+export const firefoxHasPolyfill = ( clientId, attributes ) => {
+	if ( ! clientId ) {
+		return <></>
```
bfintal, line 4: It's better to return null here so that it doesn't have to go through `Fragment`

```suggestion
		return null
```

src/block-components/block-div/firefox-has-polyfill.js
```diff
@@ -0,0 +1,33 @@
...
+			}` }
+			{ attributes.blockMargin?.left === 'auto' && `[data-block="${ clientId }"] {
+				margin-left: auto !important;
+				margin-right: 0;
+				width: fit-content;
+			}` }
+		</style>
+	}
+
+	return <></>
```
bfintal, line 31: Here too
```suggestion
	return null
```

bfintal (changes requested):
Almost there!

bfintal (changes requested):

andeng1106:
@mxkae @bfintal Issue still happens for Firefox in editor
https://github.com/gambitph/Stackable/assets/103395655/94c9e5fe-4991-41bd-b66d-8bd02bf36285

PR #2871, opened 2023-08-29
fix (timeline block): align dot if background is enabled in mobile

### LINKED ISSUE #2863: Timeline block - If background is enabled, dot is not aligned with line on frontend mobile
**Describe the bug**
- Issue only happens on frontend mobile view
- Also add this new setting: Add an option to move the dot up/down on mobile

**To Reproduce**
1. Add a new page
2. Add at least two Timeline blocks
3. Enable block background for both blocks
4. Switch to mobile view in editor
5. See that dot is connected/aligned to the line (expected)
6. Save page and go to frontend
7. Switch to mobile view in frontend
8. See that the dot is not connected/aligned to the line (actual)

**Expected behavior**
- If background is enabled, dot should be connected/aligned to the line on frontend mobile view
- [New feature] Should be able to add an option to move the dot up/down on mobile

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/70a00bf9-9821-40bc-a18a-eb61e67f1185

mxkae:
fixes #2863

### DIFF
--- a/src/block/timeline/style.scss
+++ b/src/block/timeline/style.scss
@@ -125,9 +125,6 @@
 	.stk-block-timeline > .stk-inner-blocks::after {
 		inset-inline-start: calc(var(--line-dot-size, 16px) / 2 - var(--line-bg-width, 3px) / 2 + 16px);
 	}
-	.stk-block-timeline.stk-block-background .stk-block-timeline__middle {
-		margin-left: -16px;
-	}
 	.stk-block-timeline .stk-block-timeline__middle {
 		position: absolute;
 		inset-inline-start: 16px;
--- a/src/block/timeline/style.js
+++ b/src/block/timeline/style.js
@@ -171,7 +171,7 @@ const Styles = props => {
 					const getInherited = true
 					const padding = getAttribute( 'blockPadding', device, 'normal', getInherited )
 					if ( device === 'mobile' ) {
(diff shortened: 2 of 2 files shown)


src/block/timeline/style.scss
```diff
@@ -125,9 +125,6 @@
 	.stk-block-timeline > .stk-inner-blocks::after {
 		inset-inline-start: calc(var(--line-dot-size, 16px) / 2 - var(--line-bg-width, 3px) / 2 + 16px);
 	}
-	.stk-block-timeline.stk-block-background .stk-block-timeline__middle {
-		margin-left: -16px;
-	}
```
bfintal, line 130: This might cause previously existing timeline blocks to break in mobile.

bfintal (changes requested):
Ideally, the fix should not alter existing timeline blocks (unless they're broken and this fixes it). But from the looks of the code, the removal of the style in `style/scss` which is fixed by the generated style in `style.js` might break pages with already existing timeline blocks.

mxkae:
added timeline blocks (branch w/o fix):

https://github.com/gambitph/Stackable/assets/98727316/ea800d61-8426-4742-9f6c-7f4b358f9eba

switched to branch w/ fix:

https://github.com/gambitph/Stackable/assets/98727316/fc444641-102a-4b5f-8d81-90fe53246aa8

Issue #2874, opened 2023-08-30
Include FAQ schema toggle in Accordion block


bfintal:
Add a new panel in the advanced tab of an accordion block:

Panel name: FAQ Schema
Option: (toggle) Enable FAQ Schema

### Implementation

The implementation should be In PHP. If the attribute is set, the accordion contents should be parsed using the `render_block` filter to extract the "answer" title and content. All FAQ Schema-enabled accordions should be parsed and then the schema should be outputted at the footer `wp_footer` as a single entry.

The FAQ Schema: https://developers.google.com/search/docs/appearance/structured-data/faqpage#examples

Only one schema `script` tag should be outputted.

If no blocks have FAQ Schema enabled, then no additional tag should be outputted.

PR #2885, opened 2023-09-04
Feat (Accordion Block): FAQ Schema

mxkae:
#2874

### DIFF
--- a/src/block/accordion/edit.js
+++ b/src/block/accordion/edit.js
@@ -145,9 +145,6 @@ const Edit = props => {
 						</PanelAdvancedSettings>
 					</InspectorStyleControls>
 
-					<Alignment.InspectorControls />
-					<BlockDiv.InspectorControls />
-					<Advanced.InspectorControls />
 					<InspectorAdvancedControls>
 						<PanelAdvancedSettings
 							title={ __( 'FAQ Schema', i18n ) }
@@ -159,6 +156,10 @@ const Edit = props => {
 							/>
 						</PanelAdvancedSettings>
 					</InspectorAdvancedControls>
+
+					<Alignment.InspectorControls />
+					<BlockDiv.InspectorControls />
+					<Advanced.InspectorControls />
 					<Transform.InspectorControls />
 					<EffectsAnimations.InspectorControls />
 					<CustomAttributes.InspectorControls />
--- a/src/block/accordion/edit.js
+++ b/src/block/accordion/edit.js
@@ -142,20 +142,12 @@ const Edit = props => {
 									description: __( 'Automatically closes adjacent accordion panels when clicked.', i18n ),
 								} }
 							/>
-						</PanelAdvancedSettings>
-					</InspectorStyleControls>
-
-					<InspectorAdvancedControls>
-						<PanelAdvancedSettings
-							title={ __( 'FAQ Schema', i18n ) }
-							id="faq-schema"
-						>
 							<AdvancedToggleControl
 								label={ __( 'Enable FAQ Schema', i18n ) }
 								attribute="enableFAQ"
 							/>
 						</PanelAdvancedSettings>
-					</InspectorAdvancedControls>
+					</InspectorStyleControls>
 
 					<Alignment.InspectorControls />
 					<BlockDiv.InspectorControls />
--- a/src/block/accordion/edit.js
+++ b/src/block/accordion/edit.js
@@ -15,6 +15,7 @@ import {
 	InspectorStyleControls,
 	InspectorTabs,
 	PanelAdvancedSettings,
+	InspectorAdvancedControls,
 } from '~stackable/components'
 import {
 	BlockDiv,
@@ -147,6 +148,17 @@ const Edit = props => {
 					<Alignment.InspectorControls />
 					<BlockDiv.InspectorControls />
 					<Advanced.InspectorControls />
+					<InspectorAdvancedControls>
+						<PanelAdvancedSettings
+							title={ __( 'FAQ Schema', i18n ) }
+							id="faq-schema"
+						>
+							<AdvancedToggleControl
+								label={ __( 'Enable FAQ Schema', i18n ) }
+								attribute="enableFAQ"
+							/>
+						</PanelAdvancedSettings>
+					</InspectorAdvancedControls>
 					<Transform.InspectorControls />
 					<EffectsAnimations.InspectorControls />
 					<CustomAttributes.InspectorControls />
--- a/src/block/accordion/index.php
+++ b/src/block/accordion/index.php
@@ -114,7 +114,13 @@ if ( ! class_exists( 'Stackable_Accordion_FAQ_Schema') ) {
 			if ( isset( $attributes[ 'enableFAQ' ] ) && $attributes[ 'enableFAQ' ] ) {
 				// innerBlocks[0] is for the title
 				// retrieve stackable/column -> stackable/icon-label -> stackable/heading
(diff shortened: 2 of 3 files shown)


src/block/accordion/index.php
```diff
@@ -66,3 +66,64 @@ function stackable_load_accordion_frontend_polyfill_script() {
 	}
 	add_action( 'stackable/accordion/enqueue_scripts', 'stackable_load_accordion_frontend_polyfill_script' );
 }
+
+
+if ( ! function_exists( 'stackable_render_block_accordion_FAQ_schema' ) ) {
+	function get_answer( $block, $answer ) {
```
bfintal, line 72: The `get_answer` function name is globally available. To prevent conflicts and overrides, use a namespaced function name like `stackable_faq_schema_get_answer`

```suggestion
	function get_answer( $block, $answer ) {
```

src/block/accordion/index.php
```diff
@@ -66,3 +66,64 @@ function stackable_load_accordion_frontend_polyfill_script() {
...
+		$count_inner_blocks = count( $block[ 'innerBlocks' ] );
+		if ( $count_inner_blocks == 0 ) {
+			// check if it contains a text
+			if ( trim( strip_tags( $block[ 'innerHTML' ] ) ) != '' ) {
+				// return the text with tags
+				return $answer . trim( strip_tags( $block[ 'innerHTML'], [
+					"<h1>", "<h2>", "<h3>", "<h4>", "<h5>", "<h6>",
+					"<p>", "<a>", "<span>", "<strong>", "<em>", "<s>",
+					"<kbd>", "<code>", "<sup>", "<sub>"] ) );
```
bfintal, line 81: Would you be able to use `wp_strip_all_tags` here? I'm thinking you want to strip (for example) this:

`Hello <strong>World</strong>` to `Hello World`

```suggestion
				return $answer . wp_strip_all_tags( $block[ 'innerHTML'], true );
```

Maybe you can also use this in the `if` statement above

src/block/accordion/index.php
```diff
@@ -66,3 +66,64 @@ function stackable_load_accordion_frontend_polyfill_script() {
...
+						"@type": "Question",
+						"name": "' . $question . '",
+						"acceptedAnswer": {
+							"@type": "Answer",
+							"text": \'' . $answer .'\'
+							}
+						}
+					]
+				}
+				</script>';
```
bfintal, line 122: Kindly check the spec in https://github.com/gambitph/Stackable/issues/2874

Optimally we should only have one `FAQPage` in our rendered page, so the different blocks via `render_block` should only gather all the answers, then a separate `wp_footer` call would print out all the answers in a single `FAQPage` schema.

Currently, this is printing out an `FAQPage` schema per accordion block

src/block/accordion/edit.js
```diff
@@ -147,6 +148,17 @@ const Edit = props => {
 					<Alignment.InspectorControls />
 					<BlockDiv.InspectorControls />
 					<Advanced.InspectorControls />
+					<InspectorAdvancedControls>
```
bfintal, line 151: Move this up so that the **FAQ Schema** panel would be the first entry in the Advanced Tab if possible.

bfintal (changes requested):

src/block/accordion/index.php
```diff
@@ -66,3 +66,76 @@ function stackable_load_accordion_frontend_polyfill_script() {
...
+			add_filter( 'render_block_stackable/accordion', array( $this, 'stackable_render_block_accordion_FAQ_schema' ), 10, 2 );
+		}
+
+		public static function stackable_accordion_get_faq_schema() {
+			echo '<script type="application/ld+json">' . implode( ', ', self::$accordion_faq_schema ) . '</script>';
+		}
+
+		public static function stackable_add_schema_to_footer() {
+			add_filter( 'wp_footer', array( 'Stackable_Accordion_FAQ_Schema', 'stackable_accordion_get_faq_schema' ) );
+		}
```
bfintal, line 85: Looks like this is called by `self::stackable_add_schema_to_footer();` multiple times - while this should only be done once. Instead of doing this, just add the filter for `wp_footer` in the `__construct()`. Something like:

```php
function __construct() {
	add_filter( 'wp_footer', array( $this, 'print_faq_schema' ) );
}

public function print_faq_schema() {
	if ( count( $this->faq_answers ) ) {
		// echo...
	}
}
```

src/block/accordion/index.php
```diff
@@ -66,3 +66,76 @@ function stackable_load_accordion_frontend_polyfill_script() {
...
+						"name": "' . $question . '",
+						"acceptedAnswer": {
+							"@type": "Answer",
+							"text": \'' . $answer .'\'
+							}
+						}
+					]
+				}';
+
+				self::stackable_add_schema_to_footer();
```
bfintal, line 133: Delete this based on the suggestion earlier to add the filter from the constructor
```suggestion
```

bfintal (changes requested):
Please see my comments :)

src/block/accordion/index.php
```diff
@@ -66,3 +66,74 @@ function stackable_load_accordion_frontend_polyfill_script() {
...
+				$answer = $partial_answer;
+			}
+
+			return $answer;
+		}
+
+		public function render_block_accordion_faq_schema( $block_content, $block ) {
+			$attributes = $block[ 'attrs' ];
+
+			if ( isset($attributes[ 'enableFAQ' ] ) && $attributes[ 'enableFAQ' ] ) {
```
bfintal, line 114:
```suggestion
			if ( isset( $attributes['enableFAQ'] ) && $attributes['enableFAQ'] ) {
```

src/block/accordion/index.php
```diff
@@ -66,3 +66,74 @@ function stackable_load_accordion_frontend_polyfill_script() {
...
+				// content may have multiple blocks so we need to retrieve all texts from each block
+				$answer = '';
+				$answer = $this->get_faq_answer( $block[ 'innerBlocks' ][1], $answer );
+
+				$this->faq_entities[] = '{
+						"@type": "Question",
+						"name": "' . $question . '",
+						"acceptedAnswer": {
+							"@type": "Answer",
+							"text": \'' . $answer .'\'
```
bfintal, line 129: Since you are adding the `$question` and `$answer` here from inputted text, you may encounter invalid JSON formatted strings here if the question or answer contains a quotation mark. You will need to escape both variables using the `json_encode( $str )` function (I this function adds a a quotation mark at the start and end of the output, so you may need to account for that)

e.g.
```diff
- "name": "' . $question . '",
+ "name": ' . json_encode( $question ) . ',
```

bfintal (changes requested):
Almost there!

andeng1106:
@bfintal @mxkae I have a question. 
I've tried some scenarios that might be an issue when displaying the FAQ Schema:
- When adding quotation mark ("), an additional backslash (\) is added
- When adding other characters (e.g. Korean, Chinese, Arabic, etc.) it displays as "\uc44c" instead of the actual character
Here's a screenshot with the scenarios:
Screenshot 2023-10-03 at 12 14 16 PM

mxkae:
Additional backslash when adding quotation (") is expected. 

New commit now displays:
- unicode characters as is.
- `<br>`
> @bfintal @mxkae I have a question. I've tried some scenarios that might be an issue when displaying the FAQ Schema. Please confirm if this scenarios are normal or not:
> 
> * When adding quotation mark ("), an additional backslash (`\`) is added
> * When adding other characters (e.g. Korean, Chinese, Arabic, etc.) it displays as "\uc44c" instead of the actual character
> * Breaks were not recognized (no `<br>` when new line)
>   Here's a screenshot with the scenarios:
> 
> Screenshot 2023-10-03 at 12 14 16 PM

PR #2886, opened 2023-09-04
Feat (Video Popup): add video controls options

mxkae:


src/block/video-popup/edit.js
```diff
@@ -81,6 +82,11 @@ const Edit = props => {
 		'stk-hover-parent',
 	] )
 
+	const isVideoFile = link => {
+		const regexp = /(?:^.*\.(mp4|avi|wmv|mov|flv|mkv|webm|vob|ogv|m4v|3gp|3g2|mpeg|mpg|m2v|m4v|svi|3gpp|3gpp2|mxf|roq|nsv|flv|f4v|f4p|f4a|f4b)$)/im
```
bfintal, line 86:
```suggestion
		if ( ! link ) {
			return false
		}
		const regexp = /(?:^.*\.(mp4|avi|wmv|mov|flv|mkv|webm|vob|ogv|m4v|3gp|3g2|mpeg|mpg|m2v|m4v|svi|3gpp|3gpp2|mxf|roq|nsv|flv|f4v|f4p|f4a|f4b)$)/im
```

bfintal (changes requested):

src/block/video-popup/save.js
```diff
@@ -54,6 +54,11 @@ export const Save = props => {
 			attributes={ attributes }
 			data-video={ attributes.videoLink }
 			version={ props.version }
+			{ ...{
+				'data-nofullscreen': attributes.videoFullscreen ? null : '',
+				'data-nodownload': attributes.videoDownload ? null : '',
+				'data-loop': attributes.videoLoop ? '' : null,
+			} }
```
bfintal, line 61: You can place the props directly instead of adding them to an object and spreading it
```suggestion
			data-nofullscreen={ attributes.videoFullscreen ? null : '' }
			data-nodownload={ attributes.videoDownload ? null : '' }
			data-loop={ attributes.videoLoop ? '' : null }
```

PR #2889, opened 2023-09-05
feat (image): add image through url and add captions to an image

### LINKED ISSUE #2873: Image select control - Add a text control to add an image via URL
Add a text control under the image upload option - this would need to add another attribute to hold the image.

When an image is picked from the existing picker, the image url should be emptied.. then when the image url is typed, it should remove the picked image.

Add this in:
- image block
- try if we can add this for all blocks that can have a background image

### LINKED ISSUE #2875: Add option to add caption <figcaption> in image block
Add a new text option in the image block, to add a caption to the image. This is different from the image alt, the caption option should be above the image alt.

Adding a caption should create a `figcaption` tag below the `figure > img` tag.

figcaption docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption

prconcepcion:
fixes #2873, fixes #2875


src/block-components/image/index.js
```diff
@@ -112,6 +115,10 @@ export const Image = props => {
 		defaultWidth={ props.defaultWidth }
 		defaultHeight={ props.defaultHeight }
 
+		figcaption={ attributes.figcaption }
+		figcaptionShow={ attributes.figcaptionShow }
+		figcaptionClasses={ props.figcaptionClasses }
```
bfintal, line 120: I think you mean `figcaptionClassnames`
```suggestion
		figcaptionClassnames={ props.figcaptionClasses }
```

src/block-components/image/index.js
```diff
@@ -161,6 +168,10 @@ Image.Content = props => {
 		hasGradientOverlay={ hasHoverOverlay }
 		hasLightbox={ attributes.imageHasLightbox }
 
+		figcaption={ attributes.figcaptionText }
+		figcaptionShow={ attributes.figcaptionShow }
+		figcaptionClasses={ props.figcaptionClasses }
```
bfintal, line 173: I think you mean `figcaptionClassnames`
```suggestion
		figcaptionClassnames={ props.figcaptionClasses }
```

src/block-components/image/image.js
```diff
@@ -444,27 +454,53 @@ const ImageContent = props => {
...
+			src={ props.src || undefined }
+			width={ width || undefined }
+			height={ height || undefined }
+			{ ...propsToPass }
+		/>
+	)
+
+	return (
+		applyFilters( 'stackable.image.save.wrapper',
+			( <Wrapper className={ props.blockName === 'stackable/image' ? undefined : imageWrapperClasses }>
```
bfintal, line 490: you might need to change this also

src/block-components/image/image.js
```diff
@@ -444,27 +454,53 @@ const ImageContent = props => {
...
+			</Wrapper> ), imageWrapperClasses, props?.version, notImageBlock, Wrapper, props.blockName
+		)
 	)
 }
 
 ImageContent.defaultProps = {
 	imageId: '',
 
+	blockName: '',
+
```
bfintal, line 503: Don't forget to remove this

src/block/image/deprecated.js
```diff
@@ -3,6 +3,25 @@ import { attributes } from './schema'
 
 import { withVersion } from '~stackable/higher-order'
 
+import { addFilter } from '@wordpress/hooks'
+import { semverCompare } from '~stackable/util'
+
+addFilter( 'stackable.image.save.wrapper', 'stackable/newWrapper', ( output, imageWrapperClasses, version, notImageBlock, Wrapper, blockName ) => {
```
bfintal, line 9: It's odd that you're padding `notImageBlock` here. `notImageBlock` should be removed from the block component

src/block/image/edit.js
```diff
@@ -102,6 +116,7 @@ const Edit = props => {
 				className={ blockClassNames }
 			>
 				<Image
+					figcaptionClasses={ figcaptionClasses }
```
bfintal, line 119:
```suggestion
					figcaptionClassnames={ figcaptionClasses }
```

bfintal (changes requested):

bfintal (changes requested):
Some recommendations and comments:
- The background image url should be responsive, they can set a different value for tablet or mobile (so they can mix things e.g. uploaded image for desktop, then typed in url for mobile). For non-background images, it's okay for it to be non-responsive
- If you don't have an image, the caption do not show up
- Add a placeholder for the caption
- Caption is not showing in frontend

bfintal (changes requested):
Some findings:
- Adding an image URL in a background doesn't show the background image options. If you pick an image from the media manager from the upload button, it will show additional background image options - entering an image url should also show the same.
- Adding a background image URL doesn't work if you have a background gradient. Looks like the styles are added differently - if you try adding a background image using the upload button, it shows the gradient correctly.

andeng1106:
Additional issues found:
- #2875
  - When Image block is inside an Accordion block and I enabled Container for the accordion content area, added a text color for Parent Hovered state, it doesn't show the color upon hovering. (issue in editor only)
  Here's a vid, comparing Image caption with a Text block:

https://github.com/gambitph/Stackable/assets/103395655/bc4548b4-7025-421e-af91-302cf87f6d2d

andeng1106:
Remaining issues for https://github.com/gambitph/Stackable/issues/2873:

- [ ]  Image block - No preview in Focal Point:

https://github.com/gambitph/Stackable/assets/103395655/cb950128-b831-4767-9c22-49c2b0601659

- [ ] When uploading Background image on all responsive views and use both (uploaded image on desktop > url image on desktop & uploaded image on tablet & mobile), it doesn't show properly on **editor only**
    - Example scenario: Added an image url on desktop & uploaded image on tablet & mobile. It doesn't show the uploaded image on tablet & mobile.
  

https://github.com/gambitph/Stackable/assets/103395655/00e10f1e-e784-4075-9819-9def0c67a6c2

- [ ] When adding Background Image URL, Background panel doesn't automatically enable
   - Related to: https://github.com/gambitph/Stackable/issues/2673
 
@prconcepcion

prconcepcion:
@bfintal the last 2 commits are related to what we have talked about yesterday

src/components/block-css/index.js
```diff
@@ -219,15 +221,32 @@ const BlockCss = props => {
...
+		valueDesktop = getValue( attrName, 'desktop', 'normal' )
+		if ( hasHover ) {
+			valueDesktopHover = getValue( attrName, 'desktop', 'hover' )
+		}
+		if ( hasParentHover ) {
+			valueDesktopParentHover = getValue( attrName, 'desktop', 'parent-hover' )
+		}
+		if ( hasCollapsed ) {
+			valueDesktopCollapsed = getValue( attrName, 'desktop', 'collapsed' )
+		}
```
bfintal, line 236: This looks unnecessarily repeated. I would recommend just having `hasDesktop` to be true if `responsive === false`

bfintal (changes requested):

andeng1106:
Remaining issue 1 & 2 has been fixed. This one hasn't been included in the fix yet but might be a good suggestion. Should we not fix this for now? @bfintal 
> * [ ]  When adding Background Image URL, Background panel doesn't automatically enable
>   
>   * Related to: [When adding a background image or background color, auto-enable hasBackground #2673](https://github.com/gambitph/Stackable/issues/2673)
> 
>

bfintal:
> Remaining issue 1 & 2 has been fixed. This one hasn't been included in the fix yet but might be a good suggestion. Should we not fix this for now? @bfintal 
> 
> > * [ ]  When adding Background Image URL, Background panel doesn't automatically enable
> 
> >   
> 
> >   * Related to: [When adding a background image or background color, auto-enable hasBackground #2673](https://github.com/gambitph/Stackable/issues/2673)
> 
> > 
> 
> > 
> 
> 
> 
> 

Good idea, let's add this

src/block-components/image/image.js
```diff
@@ -312,6 +313,7 @@ const Image = memo( props => {
 			) }
 			{ props.children }
 		</ResizableBox>
+
```
bfintal, line 316:
```suggestion
```

PR #2891, opened 2023-09-11
feat (border-radius): added four range border control

### LINKED ISSUE #2877: Turn border radius option into four range controls
Do this for block, background and container borders

prconcepcion:
fixes #2877


src/block-components/helpers/borders/style.js
```diff
@@ -33,6 +33,58 @@ const Styles = props => {
 				hover="all"
```
prconcepcion, line 33: @bfintal left this become the attribute only changes when the block was clicked

src/block-components/helpers/borders/deprecated.js
```diff
@@ -0,0 +1,43 @@
+import { addFilter } from '@wordpress/hooks'
+import { dispatch } from '@wordpress/data'
+
+addFilter( 'stackable.block-component.helpers', 'borders', ( output, getAttribute, updateAttributes ) => {
+	const borderRadius = getAttribute( 'borderRadius' )
+	const borderRadiusTablet = getAttribute( 'borderRadiusTablet' )
+	const borderRadiusMobile = getAttribute( 'borderRadiusMobile' )
+	dispatch( 'core/block-editor' ).__unstableMarkNextChangeAsNotPersistent()
```
bfintal, line 8: `__unstableMarkNextChangeAsNotPersistent` applies to the **next** `updateAttributes` call. Right now this is called with the possibility that 0-3 `updateAttributes` would be called.

I recommend combining all of the attributes to be changed into just one object (all the changes in just one call) also.

src/block-components/helpers/borders/edit.js
```diff
@@ -40,10 +40,13 @@ export const BorderControls = props => {
 	const {
 		getAttribute,
 		getAttrName,
+		updateAttributes,
 	} = useAttributeEditHandlers( props.attrNameTemplate )
 
 	const borderTypeValue = getAttribute( 'borderType' ) || props.borderTypeValue
 
+	applyFilters( 'stackable.block-component.helpers', null, getAttribute, updateAttributes )
```
bfintal, line 48: The hook name `stackable.block-component.helpers` is too generic and feels that it belong to the entire `helpers` and not just the `border`

I suggest using `stackable.block-component.helpers.borders`

```suggestion
	applyFilters( 'stackable.block-component.helpers', null, getAttribute, updateAttributes )
```

src/block-components/helpers/borders/style.js
bfintal: @prconcepcion please add a comment about this

Also add a comment in the attribute list because you are going to leave `borderRadius` there

bfintal (changes requested):

src/block-components/helpers/borders/deprecated.js
```diff
@@ -0,0 +1,35 @@
+import { addFilter } from '@wordpress/hooks'
+import { dispatch } from '@wordpress/data'
+
+addFilter( 'stackable.block-component.helpers.borders', 'borders', ( output, getAttribute, updateAttributes ) => {
+	const borderRadius = getAttribute( 'borderRadius' )
+	const borderRadiusTablet = getAttribute( 'borderRadiusTablet' )
+	const borderRadiusMobile = getAttribute( 'borderRadiusMobile' )
+	dispatch( 'core/block-editor' ).__unstableMarkNextChangeAsNotPersistent()
+
+	if ( borderRadius || borderRadiusTablet || borderRadiusMobile ) {
```
bfintal, line 10:
```suggestion
	if ( borderRadius || borderRadiusTablet || borderRadiusMobile ) {
		dispatch( 'core/block-editor' ).__unstableMarkNextChangeAsNotPersistent()
```

andeng1106:
@prconcepcion Design Library navigation menu might be affected:

Screenshot 2023-10-04 at 8 13 18 PM

PR #2898, opened 2023-09-19
fix: PHP stripos() error on Firefox

### LINKED ISSUE #2897: PHP `stripos()` error on Firefox if debug mode is enabled
**Describe the bug**
- Errors are only visible on Firefox browser. Make sure to use this browser.
- Related to this issue: https://wordpress.org/support/topic/php8-2-errors/
   - User is using [LibreWolf](https://librewolf.net/) which is a custom version of Firefox
- PHP version/s used: 8.2.8, 8.2.10

Plugins/Theme needed:
- Blocksy Theme
- Query Monitor

**To Reproduce**
1. Activate Blocksy Theme
2. Activate Query Monitor (or enable debug mode)
3. Add a new page
4. Add a Native Heading block
5. Add a Stackable Heading block
6. Publish page
7. Go to Frontend
8. Go to Query Monitor > PHP Errors
9. See the error with this Message: `stripos(): Passing null to parameter #1 ($haystack) of type string is deprecated`

**Expected behavior**
Should not show `stripos()` errors on Firefox browser.

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/4645bc8c-b73d-4dfa-855b-78feca7ce896

**Desktop:**
 - OS: MacOS
 - Browser: Firefox
 - Browser Version: 117.0.1 (64-bit)

mxkae:
fixes #2897

### DIFF
--- a/src/block-components/alignment/index.php
+++ b/src/block-components/alignment/index.php
@@ -42,7 +42,7 @@ if ( ! empty( $user_agent ) && $load_polyfill ) {
 	if ( ! function_exists( 'stackable_render_block_alignment_frontend_polyfill' ) ) {
 
 		function stackable_render_block_alignment_frontend_polyfill ( $block_content, $block ) {
-			if ( stripos( $block[ 'blockName' ], 'stackable/' ) === false ) {
+			if ( stripos( $block[ 'blockName' ] ?? '', 'stackable/' ) === false ) {
 				return $block_content;
 			}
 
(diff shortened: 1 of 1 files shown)


mxkae:
stripos() doesn't allow null to be passed as the haystack parameter; should I use null coalescing operator for all stripos() or only in /src/block-components/alignment/index.php since it is the only indicated location of error?
Screenshot 2023-09-19 at 11 52 51 PM

src/block-components/alignment/index.php
```diff
@@ -42,7 +42,7 @@
 	if ( ! function_exists( 'stackable_render_block_alignment_frontend_polyfill' ) ) {
 
 		function stackable_render_block_alignment_frontend_polyfill ( $block_content, $block ) {
-			if ( stripos( $block[ 'blockName' ], 'stackable/' ) === false ) {
+			if ( stripos( $block[ 'blockName' ] ?? '', 'stackable/' ) === false ) {
```
bfintal, line 45: We cannot use `??` because this is was introduced in PHP 7.0, we need to support lower versions of PHP

bfintal (changes requested):

PR #2909, opened 2023-10-04
fix: display theme font family in global typography

### LINKED ISSUE #2858: Global Typography panel does not show the actual font family used by the current block theme
**Describe the bug**
- Global Typography settings doesn't show the default font family for the current block theme
- Issue happens on:
   - site editor
   - post/page editor (when switching to tablet and mobile view)
- Issue only in Global Typography settings, it displays correctly in content area

**To Reproduce**
1. Go to site editor
2. Go to Global Settings
3. Go to Global Typography
4. See that it doesn't use the default font family for the current theme but instead uses **serif** font

**Expected behavior**
- Global Typography settings should display the default font family used by current theme

**Screenshots**
_**Post / Page Editor**_

https://github.com/gambitph/Stackable/assets/103395655/46052e1a-43a3-41ac-b4ff-41350bacf1f9

_**Site Editor**_

https://github.com/gambitph/Stackable/assets/103395655/21832f24-9350-4b15-a8cc-2c44f0f377de

mxkae:
fixes #2858

### DIFF
--- a/src/plugins/global-settings/typography/get-theme-styles.js
+++ b/src/plugins/global-settings/typography/get-theme-styles.js
@@ -1,0 +1,22 @@
+export const getThemeStyles = () => {
+	const iframe = document.querySelector( 'iframe[name="editor-canvas"]' )
+	let themeStyles = ''
+
+	if ( iframe ) {
+		const doc = iframe.contentDocument || iframe.contentWindow.document
+
+		// theme styles are stored in <style> tags that are direct children of body
+		let styleList = doc.querySelectorAll( 'body style' )
+
+		// converts node list to array
+		styleList = Array.from( styleList )
+
+		styleList.forEach( style => {
+			// retrieves css rules with .editor-styles-wrapper only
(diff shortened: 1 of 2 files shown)


andeng1106:
@mxkae Issue still happens in Site Editor for WP versions 6.1 & 6.2

bfintal:
> @mxkae Issue still happens in Site Editor for WP versions 6.1 & 6.2

Does this happen in Site Editor 6.3? If it works in 6.3, then let's not support this in 6.2 & 6.1 unless there are errors.

src/plugins/global-settings/typography/get-theme-styles.js
```diff
@@ -0,0 +1,27 @@
+export const getThemeStyles = () => {
```
bfintal, line 1:
```suggestion
// This function collects all the styles inside the editor iframe that contain the selector .editor-styles-wrapper.
export const getThemeStyles = () => {
```

PR #2926, opened 2023-10-16
feat (tabs): added option to make height equal throughout the tab's contents

### LINKED ISSUE #2844: Tabs block - Container spacing for Tabs should follow spacing for blocks that are added
**Describe the bug**
- An example scenario would be: When you added 3 accordion blocks in Tab 1 then added 6 accordion blocks in Tab 2, Tab 1 has additional spacing due to Tab 2 having lots of blocks. This shouldn't happen and Tab 1 should not be equal with Tab 2's height.

**To Reproduce**
1. Add a Tabs block
2. In Tab 1, add 3 accordion blocks
3. In Tab 2, add 6 accordion blocks
4. Enable Block or Container Background (issue is more noticeable)
5. Switch between Tab 1 and Tab 2
6. See additional spacing for Tab 1

**Expected behavior**
- Should have a toggle button for Tabs block whether they want all container spacing for tabs to be equal or container spacing would be based on how many blocks are added inside the container

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/cb7bc52f-26f1-483b-8ad0-1c3abba73468

prconcepcion:
fixes #2844


bfintal (changes requested):
A few comments:

1. When you already have a tabs block and you update, in the backend, your tab content will resize smaller when changing tabs (it should not)

2. When you already have a tabs block and you update, the tabs became not equal height. The option should be pre-enabled for existing tabs. (although if I hit preview, the frontend still showed equal height)

3. I couldn't find the option to change the height, please move it to the main tabs block and not the tab content block.

4. Change the option label to "Equal tab height"

prconcepcion:
@bfintal i temporarily changed the stackable version

src/block/tabs/style.js
```diff
@@ -43,6 +43,18 @@ const Styles = props => {
 				format="%spx"
 				responsive="all"
 			/>
+			<BlockCss
+				{ ...propsToPass }
+				selector=".%s .stk-block-tab-content > .stk-block-content > .stk-block-column[hidden]"
```
bfintal, line 48: We cannot have `>` signs in the selector for the **save** render. WP has a bug that it will throw an error in multisite.

Please change the implementation so we do not have `>` in the selector. Note that this is okay for **edit**

src/block/tabs/deprecated.js
```diff
@@ -5,6 +5,16 @@ import { attributes } from './schema'
...
 
 const deprecated = [
+	{
+		attributes: attributes( '3.12.3' ),
+		migrate: attributes => {
+			return {
+				...attributes,
+				equalTabHeight: true,
+			}
+		},
```
bfintal, line 15: As discussed, you will need to include the things in the older **migrate** method for 3.11.9:

```js
let newAttributes = { ...attributes }
newAttributes = deprecateContainerBackgroundColorOpacity.migrate( newAttributes )
newAttributes = deprecateBlockBackgroundColorOpacity.migrate( newAttributes )
return newAttributes
```

This is because deprecation only stops when it encounters the first satisfied deprecation, so all users with an old tabs block will never get migration to the new combined opacity and color.

bfintal (changes requested):
Please see my comment about the `>` in the selector and about migration.

Also, kindly clean up the PR, there is an imported `i18n` that's no longer used and a stray `console.log`

PR #2927, opened 2023-10-16
fix (border-radius): prevent border from misaligning with the new border radius control

prconcepcion:


src/block-components/helpers/borders/style.js
```diff
@@ -109,6 +110,27 @@ const Styles = props => {
 				valueCallback={ () => 'hidden' }
 				dependencies={ [ 'overflow' ] }
 			/>
+			<BlockCss
+				{ ...propsToPass }
+				// borderRadius is deprecated, so we had to duplicate the style
+				// generated by the borderRadius to be used by borderRadius2.
```
bfintal, line 116:
```suggestion
				// borderRadius is deprecated, so we had to duplicate the style
				// generated by the borderRadius to be used by borderRadius2.
				// Adding a border radius should append `overflow: hidden`.
				// This is to prevent gradient background from overflowing.
```

Issue #2822, opened 2023-08-09
If Stackable blocks are inside synced patterns and duplicated, blocks has the same id which makes settings apply to both original and duplicated block


andeng1106:
**Describe the bug**
- If Stackable blocks are inside synced pattern and duplicated, blocks has the same id which makes settings apply to both original and duplicated block. (e.g. an icon list is inside the reusable block and when I duplicate the icon list and change settings (e.g. icon, color settings, layout settings, etc.)
- Issue only on editor
     - Remaining issue [see comment below](https://github.com/gambitph/Stackable/issues/2822#issuecomment-1783026763)
- Applies to all blocks
- Reference: https://secure.helpscout.net/conversation/2323309108/21090?folderId=2526706

**To Reproduce**
1. Add a new page
2. Add any Stackable block (e.g. Heading)
3. Set the Stackable block as a synced pattern (Options > Create pattern/reusable block > set as "Synced")
4. Inside the reusable block, duplicate the Stackable block
5. In the duplicated block, change any setting (color, layout, icon, etc.)
6. See that the settings are also applied on the original block

Optional steps to see same id for both blocks:
7. Save page and go to frontend
8. Right click block and inspect
9. See that both blocks has the same id

**Expected behavior**
When duplicating Stackable block inside a reusable block and changed settings for the duplicated block, it should not:
- Apply setting for the original block
- Should not have same id with original block

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/b1dccb5e-d5ae-4534-9c12-9342739ed779

https://github.com/gambitph/Stackable/assets/103395655/ec118260-30c5-4cb3-a774-0f0efeb74fd3

andeng1106:
Update 10/27/23 - Fixed in frontend, remaining issue on editor

https://github.com/gambitph/Stackable/assets/103395655/349031b8-c348-436d-93d7-5d42471384d8

PR #2935, opened 2023-10-18
fix (unique-ids): prevent unique ids from duplicating in the frontend

### LINKED ISSUE #2890: Ensure unique IDs are all unique in the frontend
We are already doing this in the editor, but we also need to handle this for the frontend because people might be duplicating reusable blocks or content that we cannot detect.

## How

Use `render_block` in PHP to gather the `uniqueId` attributes as the blocks are getting rendered, when we encounter a duplicate `uniqueId`, we generate a new `uniqueId` and replace all the instances of the id.

This should be **high** priority and should happen before inlined CSS is optimized.

prconcepcion:
fixes #2890

### DIFF
--- a/src/unique-id.php
+++ b/src/unique-id.php
@@ -29,7 +29,8 @@ $stackable_unique_ids = array();
 if ( ! function_exists( 'stackable_prevent_duplicate_unique_ids' ) ) {
 	function stackable_prevent_duplicate_unique_ids( $block_content, $block ) {
 
-		if ( strpos( $block['blockName'], 'stackable/' ) === false ) {
+		$block_name = isset( $block['blockName'] ) ? $block['blockName'] : '';
+		if ( stripos( $block_name, 'stackable/' ) === false ) {
 			return $block_content;
 		}
 
--- a/src/unique-id.php
+++ b/src/unique-id.php
@@ -23,7 +23,7 @@ if ( ! function_exists( 'stackableGetRandomString' ) ) {
 	}
 }
 
-function stackable_generate_unique_id( $block_content, $block, $instance ) {
+function stackable_generate_unique_id( $block_content, $block ) {
 
 	$test = wp_cache_get('stackable_unique_ids');
 	if ( ! $test ) {
@@ -43,4 +43,4 @@ function stackable_generate_unique_id( $block_content, $block, $instance ) {
 	return $block_content;
 }
 
-add_filter( 'render_block', 'stackable_generate_unique_id', 9, 3 );
+add_filter( 'render_block', 'stackable_generate_unique_id', 9, 2 );
--- a/src/unique-id.php
+++ b/src/unique-id.php
@@ -15,8 +15,8 @@ if ( ! function_exists( 'stackableGetRandomString' ) ) {
 		$characters = '0123456789abcdefghijklmnopqrstuvwxyz';
 		$string = '';
 
-		for ($i = 0; $i < $length; $i++) {
-			$string .= $characters[mt_rand(0, strlen($characters) - 1)];
+		for ( $i = 0; $i < $length; $i++ ) {
+			$string .= $characters[ mt_rand(0, strlen( $characters ) - 1) ];
 		}
 
 		return $string;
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
A few things:

1. I recommend not using `wp_cache_set` and `wp_cache_get` here. That function is for caching things server-side and across multiple page loads.

Instead, you can just use a global PHP variable that holds all the unique ids

2. Please fix spacing issues:

``diff
- for ($i = 0; $i < $length; $i++) {
+for ( $i = 0; $i < $length; $i++ ) {
```

3. Please also adjust variable names - do not use `$test`

File: src/unique-id.php [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+	exit;
+}
+
+if ( ! function_exists( 'stackableGetRandomString' ) ) {
+	function stackableGetRandomString( $length ) {
+		$characters = '0123456789abcdefghijklmnopqrstuvwxyz';
+		$string = '';
+
+		for ( $i = 0; $i < $length; $i++ ) {
+			$string .= $characters[ mt_rand(0, strlen( $characters ) - 1) ];
```
bfintal, line 19: `mt_rand` is not available across all web servers, please use `wp_rand` instead

src/unique-id.php
```diff
@@ -0,0 +1,49 @@
...
+		}
+
+		return $string;
+	}
+}
+
+function stackable_generate_unique_id( $block_content, $block ) {
+
+	global $stackable_unique_ids;
+	error_log(print_r($stackable_unique_ids, true));
```
bfintal, line 29: Please remove debugging stuff
```suggestion
```

src/unique-id.php
```diff
@@ -0,0 +1,49 @@
...
+	if ( ! $stackable_unique_ids ) {
+		if ( isset( $block['attrs'] ) && array_key_exists( 'uniqueId', $block['attrs'] ) ) {
+			$GLOBALS['stackable_unique_ids'] =  array( $block['attrs']['uniqueId'] );
+		}
+	}
+
+	if ( $stackable_unique_ids ) {
+		$random_unique_id = stackableGetRandomString(7);
```
bfintal, line 37:
```suggestion
		$random_unique_id = stackableGetRandomString( 7 );
```

bfintal (changes requested):
Some comments:

1. Looks like you're checking whether `$stackable_unique_ids` exists inside the function, you don't need to do this. You can define the variable outside the function right away:

```
$stackable_unique_ids = array();
```

2. You can access the global variable without using `$GLOBALS`

3. To make things more readable, I recommend assigning the unique id into a variable:

```php
$unique_id = isset( $block['attrs']['uniqueId'] ) ? $block['attrs']['uniqueId'] : '';
if ( empty( $unique_id ) ) {
	return $block_content;
}
```

4. Only do this for stackable blocks, so check first before if the block name starts with `stackable/` in the function

src/unique-id.php
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	global $stackable_unique_ids;
+	$unique_id = isset( $block['attrs']['uniqueId'] ) ? $block['attrs']['uniqueId'] : '';
+
+	if ( empty( $unique_id ) ) {
+		return $block_content;
+	}
+
+	if ( ! is_array( $stackable_unique_ids ) ) {
```
bfintal, line 39: Instead of checking if it's an array, it's simpler to just define it as an array outside the function.

$stackable_unique_ids = array();

src/unique-id.php
```diff
@@ -0,0 +1,55 @@
...
+		$stackable_unique_ids = array( $unique_id );
+		return $block_content;
+	}
+
+	if ( in_array( $unique_id, $stackable_unique_ids ) ) {
+		$random_unique_id = stackable_get_random_string( 7 );
+		$stackable_unique_ids = array_merge( $stackable_unique_ids, array( $random_unique_id )  );
```
bfintal, line 46: I recommend just pushing the new value here instead of merging for readability

src/unique-id.php
```diff
@@ -0,0 +1,55 @@
...
+ * @package Stackable
+ */
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( ! function_exists( 'stackable_get_random_string' ) ) {
+	function stackable_get_random_string( $length ) {
```
bfintal, line 14: Instead of random string, can you change this function to be stackable_generate_unique_id without the need to pass a length anymore?

bfintal (changes requested):
Minor tweaks

bfintal (changes requested):
Last change, please wrap `stackable_prevent_duplicate_unique_ids` in a `function_exists` check

src/unique-id.php
```diff
@@ -0,0 +1,56 @@
...
+			$block_content = str_replace( $unique_id, $random_unique_id, $block_content );
+		} else {
+			array_push( $stackable_unique_ids, $unique_id );
+		}
+
+		return $block_content;
+	}
+}
+
+add_filter( 'render_block', 'stackable_prevent_duplicate_unique_ids', 9, 2 );
```
bfintal, line 56:
```suggestion
add_filter( 'render_block', 'stackable_prevent_duplicate_unique_ids', 9, 2 );
}
```

andeng1106:
@prconcepcion @bfintal 

Remaining issue:

- If debug mode is on and you duplicate a block (stackable or native) / reusable block, a `strpos()` code error displays:

https://github.com/gambitph/Stackable/assets/103395655/47c210e8-90ef-4858-a2c6-78f222dfe924

- Please include the fix for this since it's related: https://github.com/gambitph/Stackable/issues/2822

src/unique-id.php
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+$stackable_unique_ids = array();
+
+
+if ( ! function_exists( 'stackable_prevent_duplicate_unique_ids' ) ) {
+	function stackable_prevent_duplicate_unique_ids( $block_content, $block ) {
+
+		if ( strpos( $block['blockName'], 'stackable/' ) === false ) {
```
bfintal, line 32:
```suggestion
		$block_name = isset( $block['blockName'] ) ? $block['blockName'] : '';
		if ( stripos( $block_name, 'stackable/' ) === false ) {
```

andeng1106:
@prconcepcion Remaining issue for https://github.com/gambitph/Stackable/issues/2822:
- Fix is now ok for frontend but issue still exist in editor:

https://github.com/gambitph/Stackable/assets/103395655/349031b8-c348-436d-93d7-5d42471384d8

bfintal:
> @prconcepcion Remaining issue for https://github.com/gambitph/Stackable/issues/2822:
> 
> - Fix is now ok for frontend but issue still exist in editor:
> 
> 
> 
> https://github.com/gambitph/Stackable/assets/103395655/349031b8-c348-436d-93d7-5d42471384d8
> 
> 
> 
> 

Please create another issue for this new one. This PR is only for the frontend and is mainly addresses dynamically inserted blocks that can get unique id conflicts

PR #2950, opened 2023-10-26
fix (WP 6.4): render block list in nav panel only if it is open

### LINKED ISSUE #2934: [WP 6.4.1] When clicking Options from list view and inspector is open, the Options encounters a glitch
**Describe the bug**
- When clicking Options from list view, the options just glitches and closes
- Other findings:
   - Make sure list view and Inspector are both open
   - It happens for blocks with nested blocks (Columns, Tabs, Carousel, etc.)

**To Reproduce**
1. Add a new page
2. Add a Tabs block
3. Open list view
4. Open Inspector settings for Tabs block
5. Click Options
6. See if it glitches and closes
7. In the inspector, switch tab from 'Block' to 'Page'
8. Repeat step 5
9. See if it doesn't glitch anymore

**Expected behavior**
When clicking Options from List view, it should open as expected and not glitch

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/103feb05-1652-40da-af84-4b3d05a8ea6c

prconcepcion:
fixes #2934

### DIFF
--- a/src/block-components/navigation-view/edit.js
+++ b/src/block-components/navigation-view/edit.js
@@ -99,7 +99,7 @@ export const Edit = props => {
 					onToggle={ () => dispatch( 'stackable/navigation-view' ).updateIsOpen( ! isOpen ) }
 				>
 					<div className="stk-panel--navigation-view__wrapper">
-						<ListView
+						{ isOpen && <ListView
 							blocks={ blocks }
 							showOnlyCurrentHierarchy
 							showAppender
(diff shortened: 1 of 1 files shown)


bfintal:
@andeng1106 the fix @prconcepcion did was to hide the navigation contents when it's collapsed. Apparently if you have both the navigation panel and the list view, the 3-dots stop working. Let's just assume that people would not want to have both nav panel and list view open at the same time.

PR #2958, opened 2023-11-06
Fix: Horizontal Scrollbar Styling in Firefox

### LINKED ISSUE #2618: Horizontal Scroller - Horizontal scrollbar styling doesn't work for Firefox browser
**Describe the bug**
- It seems like the default scrollbar for the Firefox browser is overlapping the stylings.
- Happens both in editor and frontend

**To Reproduce**
1. Use Firefox browser
2. Add a new page
3. Add a Horizontal Scroller block
4. In Inspector, enable Scrollbar
5. Add scrollbar stylings
6. See editor if the stylings are not applied
7. Publish page
8. See frontend if stylings are also not applied

**Expected behavior**
Scrollbar styling should work for Firefox browsers in both editor and frontend.

**Screenshots**

https://user-images.githubusercontent.com/103395655/225526284-460e6b3f-8869-463e-9bc8-6e75d329edf6.mp4

**Desktop:**
 - OS: MacOS, Windows
 - Browser: Firefox
 - Browser Version: 111.0 (64-bit)

kaeizen:
fixes #2618

### DIFF
--- a/src/block/horizontal-scroller/editor.scss
+++ b/src/block/horizontal-scroller/editor.scss
@@ -10,6 +10,7 @@
 	height: var(--stk-column-height);
 	padding-inline-start: var(--stk-left-offset, 0px);
 	scrollbar-color: var(--stk-scrollbar-thumb-color, #d3d3d3) var(--stk-scrollbar-track-color, #eee); // For Firefox
+	scrollbar-width: var(--stk-scrollbar-height-firefox, 'auto'); // For Firefox
 	> * {
 		scroll-snap-align: var(--stk-snapping, center);
 		width: 100% !important;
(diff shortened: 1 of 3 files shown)


bfintal:
@andeng1106 please verify whether scrollbars in the other browsers are unaffected

andeng1106:
@bfintal Other browsers aren't affected for the fix. 

@mxkae Remaining issue:
- Height settings is still not showing as expected on Firefox (issue on editor & frontend):
   Screenshot 2023-11-08 at 6 17 12 AM
   
    - Scrollbar should look like this:
Screenshot 2023-11-08 at 6 18 28 AM

kaeizen:
Only colors and width of scrollbars can be styled in Firefox.
Specifically, scrollbar width (or height) can only be set to either: `thin` `none`, or `auto`.
For this fix, I made the `scrollbar-width` set to `thin` if the height attribute is set to less than 10, else it will get the default value ( `auto` ).
Reference: 
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scrollbars_styling
https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width

PR #2962, opened 2023-11-08
fix (unique-id): check if variable is not null

### LINKED ISSUE #2961: (unique id fix): PHP Error in PHP 8.1
PHP Fatal error: Uncaught TypeError: in_array(): Argument #2 ($haystack) must be of type array, null given in /xxxxxxx/wp-content/plugins/stackable-ultimate-gutenberg-blocks/src/unique-id.php:45

prconcepcion:
fixes https://github.com/gambitph/Stackable/issues/2961

### DIFF
--- a/src/unique-id.php
+++ b/src/unique-id.php
@@ -49,10 +49,10 @@ if ( ! function_exists( 'stackable_prevent_duplicate_unique_ids' ) ) {
 
 		if ( in_array( $unique_id, $stackable_unique_ids ) ) {
 			$random_unique_id = stackable_generate_unique_id();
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
Please make sure that `$stackable_unique_ids` is always an array:

```php
// Initialize the global variable if it's not set
if ( ! isset( $stackable_unique_ids ) || ! is_array( $stackable_unique_ids ) ) {
	$stackable_unique_ids = array();
}
```

Please push directly into the array instead of calling a function:

```php
$stackable_unique_ids[] = $value;
```

PR #2965, opened 2023-11-08
fix (image block): allow richtext in image captions

### LINKED ISSUE #2964: Image block - Should be able to add a link in the image block caption

bfintal:
fixes #2964 

# TODO

This is an incomplete fix. Now that RichText in the image caption is allowed, it now conflicts with the image link: If you add an image block, then add a caption with a link, then also add a link from the advanced tab > link, the block will produce an error after saving and refreshing the editor.

The cause of the error is that the Adv tab > Link option incorrectly changes the `figure` tag into an `a` tag. The browser then detects that there is an `a` tag within an `a` tag (because you have a link in the figcaption), and tries to split the 2 `a` tags and that messes up the block structure causing the block error.

The proposed fix for this is to change the implementation of `customWrapper` in `src/block-components/image/image.js`. Since this prop is only used to turn the image into the link, we need to change this into `linkUrl` prop instead.. the image should **always** be a `figure`, then if the image needs to be made into a link, then it should just wrap the `img` tag with a link.

### DIFF
--- a/src/block-components/link/attributes.js
+++ b/src/block-components/link/attributes.js
@@ -3,7 +3,7 @@ import { addLinkAttributes } from '../helpers/link'
 export const addAttributes = ( attrObject, options = {} ) => {
 	const {
 		attrNameTemplate = 'link%s',
-		selector = 'a',
+		selector = 'a.stk-link',
 	} = options
 
 	addLinkAttributes( attrObject, attrNameTemplate, selector )
--- a/src/block-components/link/attributes.js
+++ b/src/block-components/link/attributes.js
@@ -3,7 +3,7 @@ import { addLinkAttributes } from '../helpers/link'
 export const addAttributes = ( attrObject, options = {} ) => {
 	const {
 		attrNameTemplate = 'link%s',
-		selector = 'a.stk-link',
+		selector = 'a',
 	} = options
 
 	addLinkAttributes( attrObject, attrNameTemplate, selector )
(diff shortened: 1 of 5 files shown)


bfintal (commented):
Can you try these scenarios, all should not have any block errors:
1. Create an image block in < 3.11.8, then update to this PR
2. Create an image block in < 3.11.8, then update to 3.12.3, save the page, then update to this PR
2. Create an image block in < 3.11.8 add a link in adv tab > link, then update to 3.12.3, save the page, then update to this PR
3. Create an image block in 3.12.3, then update to this PR
4. Create an image block in 3.12.3, add a caption, then update to this PR
5. Create an image block in 3.12.3, add a link in adv tab > link, then update to this PR

prconcepcion:
- [x] Create an image block in < 3.11.8, then update to this PR
- [x] Create an image block in < 3.11.8, then update to 3.12.3, save the page, then update to this PR
- [x] Create an image block in < 3.11.8 add a link in adv tab > link, then update to 3.12.3, save the page, then update to this PR
- [x] Create an image block in 3.12.3, then update to this PR
- [x] Create an image block in 3.12.3, add a link in adv tab > link, then update to this PR
- [x] Create an image block in 3.12.4, add a caption, then update to this PR
- [x] Create an image block in 3.12.4, then update to this PR
- [x] Create an image block in 3.12.4, add a link in adv tab > link, then update to this PR

src/block-components/link/attributes.js
```diff
@@ -3,7 +3,7 @@ import { addLinkAttributes } from '../helpers/link'
 export const addAttributes = ( attrObject, options = {} ) => {
 	const {
 		attrNameTemplate = 'link%s',
-		selector = 'a',
+		selector = 'a.stk-link',
```
bfintal: Instead of changing the default value for everyone that uses the `selector`, can you instead provide `a.stk-link` to the block/component that needs this?
```suggestion
		selector = 'a',
```

PR #2973, opened 2023-11-14
Feat: New option to use Font Awesome Icons v6

### LINKED ISSUE #2525: How to Add Font Awesome 6.2.1 for icon
hi，I need to use the latest 6.2.1. How do I operate it?

kaeizen:
fixes #2525

### DIFF
--- a/src/components/font-awesome-icon/index.js
+++ b/src/components/font-awesome-icon/index.js
@@ -1,9 +1,7 @@
 /**
  * External dependencies
  */
-import {
-	faAPILoaded, faIsAPILoaded, faGetSVGIcon, faIconLoaded, createElementFromHTMLString,
-} from '~stackable/util'
+import { faGetIcon, createElementFromHTMLString } from '~stackable/util'
 import { pick } from 'lodash'
 
 /**
@@ -80,14 +78,38 @@ const addSVGAttributes = ( svgHTML, attributesToAdd = {}, attributesToRemove = [
 }
 
 const FontAwesomeIcon = memo( props => {
-	const [ forceUpdateCount, setForceUpdateCount ] = useState( 0 )
-	const forceUpdate = () => {
-		setForceUpdateCount( forceUpdateCount + 1 )
-	}
+	const [ iconHTML, setIconHTML ] = useState( '' )
+	const [ isLoading, setIsLoading ] = useState( false )
 
-	// Wait for the FA API to load.
+	const prefix = props.value ? props.value.replace( /-.*$/, '' ) : props.prefix
+	const iconName = props.value ? props.value.replace( /^.*?-/, '' ) : props.iconName
+
 	useEffect( () => {
-		faAPILoaded().then( forceUpdate )
+		const loadIconHTML = async () => {
+			setIsLoading( true )
+
+			if ( typeof props.value === 'string' && props.value.match( /^<svg/ ) ) {
+				setIsLoading( false )
+				return
+			}
+
+			// Wait for the FA icon to load.
+			let _iconHTML
+			if ( prefix && iconName ) {
+				await faGetIcon( prefix, iconName ).then( svg => {
+					_iconHTML = svg
+				} )
+			} else {
+				// If no value, just display a smiley placeholder.
+				await faGetIcon( 'far', 'smile' ).then( svg => {
+					_iconHTML = svg
+				} )
+			}
+			setIconHTML( _iconHTML )
+			setIsLoading( false )
+		}
+
+		loadIconHTML()
 	}, [] )
 
 	const propsToPass = pick( props, [ 'className', 'color', 'fill', 'style' ] )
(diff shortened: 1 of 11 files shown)


src/deprecated/font-awesome-version.php
```diff
@@ -0,0 +1,30 @@
...
+	 */
+	function stackable_font_awesome_icons_version_set_default( $old_version, $new_version ) {
+		if ( ! empty( $old_version ) && version_compare( $old_version, "3.12.6", "<" ) ) {
+			// Use FontAwesome icons v5.15.4 for users upgrading from v3.12.6 and below.
+			if ( ! get_option( 'stackable_fa_icons_version' ) || get_option( 'stackable_fa_icons_version' ) === '5.15.4' ) {
+				update_option( 'stackable_fa_icons_version', '5.15.4' );
+			}
+		} else if ( empty( $old_version ) ) {
+			// Use FontAwesome icons v6.4.2 for new installations.
+			update_option( 'stackable_fa_icons_version', '6.4.2' );
```
bfintal: I would recommend removing having to save the icon version for new installs. It would be better for new installs to have `stackable_fa_icons_version` blank (which translates to `false`), so that it means that if `stackable_fa_icons_version` is false, just use 6.4.2 
```suggestion
```

src/components/font-awesome-icon/index.js
```diff
@@ -152,7 +160,10 @@ FontAwesomeIcon.Content = props => {
 	const prefix = props.value ? props.value.replace( /-.*$/, '' ) : props.prefix
 	const iconName = props.value ? props.value.replace( /^.*?-/, '' ) : props.iconName
 
-	const iconHTML = faGetSVGIcon( prefix, iconName )
+	let iconHTML
+	faGetIcon( prefix, iconName ).then( svg => {
+		iconHTML = svg
+	} )
```
bfintal: We might run into problems here because `feGetIcon` is a promise that would need to be resolved before getting the value. Since this is called in `.Content` which is called during the saving process, we cannot have any promises here because we aren't assured that we will be able to wait for this value to resolve.

You will need to change this into a synchronous way to get the icon
```suggestion
	faGetIcon( prefix, iconName ).then( svg => {
		iconHTML = svg
	} )
```

src/components/icon-search-popover/index.js
```diff
@@ -58,6 +58,14 @@ export const cleanSvgString = svgString => {
...
 		.replace( /(<\/svg>)([\s\S]*)/g, '$1' )
 
+	if ( newSvg.indexOf( '<!-- Font Awesome' ) !== -1 ) {
+		// Remove defs
+		newSvg = newSvg.replace( /<defs[\s\S]*?<\/defs>/gm, '' )
+
+		// Remove comments
+		newSvg = newSvg.replace( /<!--[\s\S]*?-->/gm, '' )
+	}
+
```
bfintal: Would this work for all comments? Removing "Font Awesome" future proofs this
```suggestion
	if ( newSvg.indexOf( '<!--' ) !== -1 ) {
		// Remove defs
		newSvg = newSvg.replace( /<defs[\s\S]*?<\/defs>/gm, '' )

		// Remove comments
		newSvg = newSvg.replace( /<!--[\s\S]*?-->/gm, '' )
	}

```

bfintal (changes requested):

bfintal (changes requested):
Some findings from my code review:

### 1. Using a v2 icon block and trying to pick an icon results in a block error

This may be because the deprecated font awesome component is still using the old method that has some methods removed. Try adding a v2 icon block and change the icon.

I would suggest keeping the old method, but moving it near in the deprecated file. I think the deprecated file that's having an error is at `src/deprecated/v2/components/font-awesome-icon/index.js` (please double check this)

### 2. When using a Font Awesome Kit, the kit's setting is ignored and the icon version used is the one in our settings.

I think the behavior should be that if there's a valid Font Awesome Kit used, the version setting should become hidden and there should be a note "Using the icon library version included in your Font Awesome Pro Kit"

### 3. Remove the "Save changes" button and make it auto-save instead

Since we now have 2 fields, the save changes button can now be confusing. Please remove the save changes button and instead, let's autosave whenever the font awesome pro kit field validates. So it should autosave if:

- the pro kit field changes
- the pro kit field value validates successfully

bfintal (changes requested):
I found that if I add a Pro Kit, then remove it, I can't anymore search for free icons

andeng1106:
@mxkae The kit codes works as expected but there are some remaining issues left:

- [ ] IF FontAwesome is deactivated, Pro kit icons should not be used and loaded if there's no pro kit code present.
Findings:  To check if Font Awesome plugin is activated, check if `defined( 'FONTAWESOME_PLUGIN_FILE' )` is `true`

- [ ] When validating pro kit codes and there's an error, it should not show `You are using the version set in your Font Awesome Pro Kit.`
    - **Expected behavior:** It should still show the dropdown since there's no registered pro kit code yet
    - 2 scenarios:
      - If empty field (click spacebar or add empty spaces)

Screenshot 2023-12-05 at 2 32 14 PM

      - If wrong/invalid pro kit code

Screenshot 2023-12-05 at 2 33 11 PM



- [ ] When pasting the [script tag as indicated in the docs](https://docs.wpstackable.com/article/358-how-to-use-your-font-awesome-pro-icons?utm_source=wp-settings-icons&utm_campaign=learnmore&utm_medium=wp-dashboard#:~:text=The%20code%20that%20we%20need%20is%20the%20one%20that%20looks%20like%3A), it doesn't fully save it then upon refresh, it resets to an empty field

https://github.com/bfintal/Stackable-Premium/assets/103395655/e3b849dc-381f-4eeb-a587-f0419b4a2c18

andeng1106:
@mxkae moved the remaining issues list on https://github.com/bfintal/Stackable-Premium/pull/321#issuecomment-1840108709

PR #2995, opened 2023-12-04
fix (inner-column): prevent Inner Block Gap from becoming 0 upon refresh

### LINKED ISSUE #2990: Inner Block Gap resets to 0 upon refresh
**Describe the bug**
- Issue appens if Inner Block Direction is set to **Horizontal**
image

- All blocks with Inner Column Gap are affected
- Happens only on editor
- Happens only on Desktop but please double check tablet and mobile

**To Reproduce**
1. Add a new page
2. Add any block with Inner Block Gap (e.g. Columns > Inner Column)
3. In Layout set Inner Block Direction to "Horizontal"
4. Adjust Inner Block Gap settings
5. Save page
6. Refresh page
7. See that Inner Block Gap setting resets to 0

**Expected behavior**
Inner Block Gap should not reset to 0 upon refreshing of page.

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/4ade0bfc-4db6-4a75-b614-fc504edced16

prconcepcion:
fixes #2990

### DIFF
--- a/src/block/call-to-action/deprecated.js
+++ b/src/block/call-to-action/deprecated.js
@@ -44,27 +44,33 @@ const deprecated = [
 			return hasContainerOpacity || hasBlockOpacity || isNotV4
 		},
 		migrate: ( attributes, innerBlocks ) => {
+			const isNotV4 = attributes.version < 2 || typeof attributes.version === 'undefined'
+
 			let newAttributes = {
 				...attributes,
-				version: 2,
 			}
 
-			// Update the vertical align into flexbox
-			const hasOldVerticalAlign = !! attributes.containerVerticalAlign // Column only, this was changed to flexbox
-
-			if ( hasOldVerticalAlign ) {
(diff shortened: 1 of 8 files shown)


bfintal (changes requested):
Please double check if there are any side-effects. I do not think you can just remove this because it's part of the deprecation process. Removing this would mean that the block update process would fail, most likely you will need to add some conditions on when to correctly apply this `innerBlockColumnGap: 0`

andeng1106:
@prconcepcion For confirmation:

- Created 2 Columns block on the current pr fix with different scenarios but upon refreshing page, there's a margin left on inner blocks. Question: Did the deprecation get triggered incorrectly?
         - Scenario 1: Inner block is 0, no assigned inner block gap
         - Scenario 2: Has an inner block gap value
  - **Expected:** should have no margin left on the inner blocks (with or without inner block gap)
  
  cc: @bfintal

PR #3004, opened 2023-12-11
fix (button): prevent block error upon adding an anchor

### LINKED ISSUE #2998: Button block encounters a block error upon refresh when adding an HTML anchor
**Describe the bug**
Findings:
- When adding an HTML anchor it assigns an ID. But it assigned an ID to the `<a>` tag:
Screenshot 2023-12-07 at 12 55 22 PM

**To Reproduce**
1. Add a new page
2. Add a Button block
3. In the Button block, add an HTML anchor (Advanced tab > Advanced > HTML Anchor)
4. Save page
5. Refresh page
6. See that the button block encounters a block error

**Expected behavior**
Should not encounter block error

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/748ca2aa-46d1-4975-b5b0-ad8915531350

prconcepcion:
fixes #2998


bfintal (changes requested):
I don't think this is the proper fix for this, because if you check all the other blocks, they do not declare `anchor` in their attribute list

src/block-components/advanced/edit.js
```diff
@@ -80,6 +81,10 @@ export const Edit = () => {
 							{ value: 'none', label: __( 'None', i18n ) },
 						] }
 					/>
+					{ props.enableAnchorId && <AdvancedTextControl
+						label={ __( 'Anchor Link', i18n ) }
```
bfintal:
```suggestion
						label={ __( 'Anchor ID', i18n ) }
```

bfintal (changes requested):
It doesn't feel right that the option for the Anchor ID is inside the Advanced block component because only the buttons are going to use it.

I suggest adding the option inside inside the button's **Style tab > Link panel** since it's related to it

bfintal (changes requested):
When I activate the plugin, I get a space on the sidebar: 
image

andeng1106:
> When I activate the plugin, I get a space on the sidebar: image

@bfintal I'm also seeing this when using 3.12.8 with debug mode enabled.

Screenshot 2024-01-04 at 2 09 02 PM

PR #3019, opened 2023-12-20
performance: add frontend variables only when there is a block that needs them

### LINKED ISSUE #3007: Revisit frontend js variables if still needed
Stackable loads the following in the frontend:

```html
<script id="ugb-block-frontend-js-js-extra">
var stackable = {"restUrl":"http:\/\/local.local\/wp-json\/","i18n":[]};
</script>
```

This is loaded in the filter `stackable_localize_frontend_script`, we need to assess...
- Do we still need `restUrl`? If it's only used by the load more block, then only add it if the block/functionality is used
- Do we still use `i18n` for the frontend? only add if it needed

Ideally in a blank page or a page with just simple blocks, we should not have an inline script tag unless needed

## QA Testing

1. The added frontend script should only be present if the load more button block is used. Otherwise, the script should not be present in the frontend
2. V2 blog posts block load more functionality should still work (no need to check if frontend script is loaded or not in v2)

prconcepcion:
fixes #3007

### DIFF
--- a/src/init.php
+++ b/src/init.php
@@ -101,10 +101,7 @@ if ( ! class_exists( 'Stackable_Init' ) ) {
 				wp_register_script( 'ugb-block-frontend-js', null, [], STACKABLE_VERSION );
 			}
 
-			$args = apply_filters( 'stackable_localize_frontend_script', array(
-				'restUrl' => get_rest_url(),
-				'i18n' => array(), // Translatable labels used in the frontend should go here.
-			) );
+			$args = apply_filters( 'stackable_localize_frontend_script', array() );
 			wp_localize_script( 'ugb-block-frontend-js', 'stackable', $args );
 
 			// Register inline frontend styles, these are always loaded.
--- a/src/init.php
+++ b/src/init.php
@@ -453,8 +453,8 @@ if ( ! class_exists( 'Stackable_Init' ) ) {
 
 		public function init_stackable_vars() {
 			$args = apply_filters( 'stackable_localize_frontend_script', array() );
-			if ( array_key_exists( 'restUrl', $args ) ) {
-				echo '<script> stackable = {"restUrl": "' . $args["restUrl"] .'"} </script>';
+			if ( ! empty( $args ) ) {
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
Do not perform the `wp_localize_script` if `$args` is empty. It still prints out in the frontend even if you pass an empty array

src/init.php
```diff
@@ -455,6 +450,14 @@ public function load_plugin_textdomain() {
 		public function init_animation() {
 			echo '<script>requestAnimationFrame(() => document.body.classList.add( "stk--anim-init" ))</script>';
 		}
+
+		// TODO: make this dynamic
```
bfintal:
```suggestion
```

src/init.php
```diff
@@ -455,6 +450,13 @@ public function load_plugin_textdomain() {
 		public function init_animation() {
 			echo '<script>requestAnimationFrame(() => document.body.classList.add( "stk--anim-init" ))</script>';
 		}
+
+		public function init_stackable_vars() {
+			$args = apply_filters( 'stackable_localize_frontend_script', array() );
+			if ( array_key_exists( 'restUrl', $args ) ) {
+				echo '<script> stackable = {"restUrl": "' . $args["restUrl"] .'"} </script>';
```
bfintal: instead of writing the `restUrl` manually, please convert `$args` to JSON instead to print out the entire `$args` if it has contents

bfintal (changes requested):

src/init.php
```diff
@@ -455,6 +450,13 @@ public function load_plugin_textdomain() {
 		public function init_animation() {
 			echo '<script>requestAnimationFrame(() => document.body.classList.add( "stk--anim-init" ))</script>';
 		}
+
+		public function init_stackable_vars() {
+			$args = apply_filters( 'stackable_localize_frontend_script', array() );
+			if ( ! empty( $args ) ) {
+				echo '<script> stackable = ' . json_encode( $args ) . ' </script> ';
```
bfintal:
```suggestion
				echo '<script>stackable = ' . json_encode( $args ) . '</script>';
```

PR #3020, opened 2023-12-21
create temporary new icon list block

### LINKED ISSUE #2818: Revamp how Icon List block is structured
This issue takes over #2763 #2764 #2025 #2365 #2980 #2970 #2971

We have a warning that the `multiline` prop will no longer be supported in WP, this is used by the Icon List block so it can exist as a single block. We'll need to change our implementation of the Icon List block and turn it into the following block structure:

- Icon List block
  - Icon List Item Block
  - Icon List Item Block
  - Icon List Item Block

We also need to change our implementation of the SVG icon to resolve some alignment issues, the new output should be:

```html
<ul>
  <li>
    <span aria-hidden="true">icon</span>
    <span>text</span>
  </li>
</ul>
```

kaeizen:
fixes #2818 

Remaining Issues:
- [x] Open icon search popover on click even when block is not selected
- [x] Move cursor to adjacent blocks without double pressing arrow keys
- [x] Clicking on backspace key should transfer the content to preceding icon list item block
- [x] Nested icon list (Add inner blocks in icon list item)
- [x] Copy paste multiple lines (transform)
- [x] inconsistent merge blocks
- [x] Deprecation


kaeizen:
Fixed: allow nested icon lists
Note: Currently, you can drag icon lists to icon list item to create nested icon lists

Additional Issue:
- [x] Indention controls on icon list item

kaeizen:
Additional issue:
- [x] If item has sub-items, splitting the text shouldn't remove its sub-items

https://github.com/gambitph/Stackable/assets/98727316/42dc2f61-62b5-4ab7-a5cc-f3094a02b116

bfintal (changes requested):
Some initial comments:
- When you update an existing icon list that a large icon size, the updated icon list doesn't look the same in the backend (expected behavior: it should remain the same)
- When you update an existing icon list that a large icon size, the spacing in the frontend looks different (expected behavior: it should remain the same)
- If you increase the size of the icon, it doesn't reflect in the frontend.

src/block-components/icon/index.js
```diff
@@ -63,6 +63,7 @@ export const Icon = props => {
 		hasLinearGradient = true,
 		value = '',
 		onChange = NOOP,
+		openOnClick = false,
```
bfintal: `openOnClick` is a bit misleading because it will always open regardless of the value. I suggest changing this to `openEvenIfUnselected` (or something better and more descriptive like that)

src/block/icon-list-item/block.json
```diff
@@ -0,0 +1,17 @@
+{
+	"apiVersion": 3,
+	"name": "stackable/icon-list-item",
+	"title": "Icon List Item",
+	"description": "A new block.",
```
bfintal: Change this:
```suggestion
	"description": "A single list entry in the Icon List block",
```

src/block/icon-list-item/block.json
```diff
@@ -0,0 +1,17 @@
...
+	"category": "stackable",
+	"usesContext": [ "postId", "postType", "queryId" , "stackable/innerBlockOrientation", "stackable/ordered", "stackable/uniqueId", "stackable/isIndented" ],
+	"keywords": [
+		"Keywords that are not in the title"
+	],
```
bfintal:
```suggestion
	"keywords": [],
```

src/block/icon-list-item/block.json
```diff
@@ -0,0 +1,17 @@
...
+	"parent": [
+		"stackable/icon-list"
+	],
+	"textdomain": "stackable-ultimate-gutenberg-blocks",
+	"stk-type": "special",
```
bfintal: This should not be visible in the settings - this should not be allowed to be disabled
```suggestion
	"stk-type": "hidden",
```

src/block/icon-list-item/edit.js
```diff
@@ -0,0 +1,212 @@
...
+		blockIndex,
+	} = blockContext
+
+	useEffect( () => {
+		setAttributes( { ordered } )
+	}, [ ordered ] )
+
+	useEffect( () => {
+		setAttributes( { parentUniqueId } )
+	}, [ parentUniqueId ] )
```
bfintal: We cannot set attributes inside effects, we should only set attribute when the user changes something. Since this is called in `useEffect`, the attributes will be set when the block is first rendered and the undo button will get a lot of new entries.

If you really need to do this, I recommend you doing it this way so that the undo won't get affected.

```js
useEffect( () => {
	setAttributes( { ordered } )
}, [ ordered ] )
```

src/block/icon-list-item/edit.js
```diff
@@ -0,0 +1,212 @@
...
+
+	const blockProps = useBlockProps( {
+		ref: useCopy( clientId ),
+		blockHoverClass: props.blockHoverClass,
+		clientId: props.clientId,
+		attributes: props.attributes,
+		className: blockClassNames,
+		blockTag: 'li',
+		renderHtmlTag: false,
+		tabIndex: '-1',
```
bfintal:
```suggestion
		tabIndex: '-1', // We need this since navigating up/down selects the wrapper.
```

src/block/icon-list-item/edit.js
```diff
@@ -0,0 +1,212 @@
...
+								disabled={ ! isIndented }
+								onClick={ outdentListItem }
+							/>
+							<ToolbarButton
+								label={ __( 'Indent', i18n ) }
+								icon="editor-indent"
+								disabled={ blockIndex === 0 }
+								onClick={ indentListItem }
+							/>
+						</ToolbarGroup>
```
bfintal: I feel that using indent/outdent still gives a wonky experience (e.g. you lose your cursor when you indent), I recommend removing indent/outdent completely for our icon list for now

src/block/icon-list/style.scss
```diff
@@ -1,43 +1,37 @@
 // Important, we can't use flex since it causes issues with columns.
-.stk-block-icon-list {
+.stk-block-icon-list.stk-block {
```
bfintal: If you have an existing icon list block, then you upgrade to the new PR, the frontend looks of the icon list block changes - the styling of the blocks have changed. we need to keep this working

bfintal (changes requested):
Please see comments.

Also, we need to optimize the SVGs in PHP. In the frontend, we end up with a LOT of repeated `<svg>` definitions that will clutter the DOM and warnings of excessive DOM usage.

In PHP, maybe in the `render_block_stackable/icon-list-item` filter, use this method for reusing `svg` elements: https://dev.to/bmstefanski/reusing-svg-elements-in-html-without-copy-pasting-it-3hfo basically we need to gather the same `<svg>`s then turn the 1st one into a definition, then output `<use>` for the rest.

bfintal (changes requested):
Some feedback:

1. Saw this markup in the frontend, do we need this? I think this might be needed for gradient colors, but since the Icon List block doesn't support gradient colors, I think we can remove this markup in the save:

Screenshot 2024-01-13 at 9 48 23 AM

2. I noticed that when you upgrade, the icon sizes become different:

Here's before you update:

image

Here's after updating (when you edit the page and save):

image

Expectation: 

- the icon sizes should still remain the same.
- (try) the indentation should be the same (maybe allow negative values, then deprecate to make it negative to mimic the old behavior)
- the gap between icon and text looks small (but I think it may be because of the icon size)

bfintal (changes requested):
After updating, the spacing is different:

https://github.com/gambitph/Stackable/assets/1033611/a02d84ed-4d21-4eb1-836a-852948f65f1f

Here's what it looks like after previewing after the update (didn't change anything except I edited the post and hit preview)


https://github.com/gambitph/Stackable/assets/1033611/75156492-2274-4b84-b0fe-a229ecf754e5

Expectation:
❌ spacing more or less the same
✅ Icons the same size
✅ Icon list now aligned all to the left

andeng1106:
@mxkae Here are the link for the issues found (including Jami's comment above): https://www.notion.so/Revamped-Icon-List-Block-Monitor-Issues-Found-34ca8570790948178b2f98a1912d8c34?pvs=4

bfintal (changes requested):
Some change requests:
- [x] Please add a default column gap of 16px (make sure this is only applicable for new icon lists)
  - [x] Also adjust the placeholder in the option
- [x] If 2+ columns, the columns should align to the entire width (right now it's collapsed to the left)
  - [x] If the layout > content alignment is left/center/right, then do not apply the full width
- [x] Icon list item borders > border width should default to 1px (it's currently thick)
  - [x] control should have placeholder of `1`
- [x] Icon list item borders > border color should default to `#00000066`
- [x] Add new option under Icon list item borders "Full Width Borders" (toggle), the borders should go fullwidth (I think this will just make the width `auto`)
- [x] Add new **Icon Vertical Alignment** option for the icon, add option in Icons & Numbers. it should be a button group like top, center, bottom like the other options.
  - [x] Help description: "Visible if you have long text in your list"
- [x] Add a new **Icon Vertical Offset** option under the Icon Vertical Alignment option. Values are from -50 to 50 (default at 0), this should add `margin-top` to the icon

src/block/icon-list/schema.js
```diff
@@ -86,6 +96,25 @@ export const iconListAttributes = {
...
+		default: '',
+	},
+	listItemBorderWidth: {
+		stkResponsive: true,
+		type: 'number',
+		default: '',
+	},
+	listItemBorderColor: {
+		type: 'string',
+		default: '#00000066',
```
bfintal: The default should be `''` here. Please set the border color CSS to `#00000066` instead of setting the default value here. Adding the default here will make the block generate extra css styles

bfintal (changes requested):

PR #3030, opened 2024-01-18
fix (posts): random order of posts functions without enabling v2 compatibility

### LINKED ISSUE #2975: Posts block - If Post Type is set to Products or Posts/Pages and order by is set to Random, it endlessly loads on the editor
**Describe the bug**
- When selecting a Post Type (Posts, Pages, or others) and set Order By to Random for Posts block, it endlessly loads and doesn't show the results in editor
- Issue only in editor
- Issue happens if "Load version 2 blocks in the editor" is checked
- A console error with 400 Bad Request shows:
Screenshot 2023-11-23 at 1 44 01 PM

- More info from user:
> According to WordPress documentation orderby only accept following parameters: author, date, id, include, modified, parent, relevance, slug, include_slugs, title and menu_order

**To Reproduce**
1. Add a new page
2. Add a Posts block
3. In Inspector, go to Style > Query > Post Type and change selection to Posts (or "Products" if you have WooCommerce)
4. Go to Query > Order by and change selection to "Random"
5. See that it loads endlessly and doesn't show the results

**Expected behavior**
If I set Order by to Random, it should show the results on editor

**Workaround**
Disable **Load version 2 blocks in the editor** in Stackable settings

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/2abc3e10-c207-4600-83a7-1946504ebbb1

prconcepcion:
fixes #2975


src/block/posts/index.php
```diff
@@ -159,7 +159,7 @@ function generate_post_query_from_stackable_posts_block( $block_or_attribute, $q
 				'post_type' => $context['type'],
 				'post_status' => 'publish',
 				'order' => $context['order'],
-				'orderby' => $context['orderBy'],
+				'orderby' => 'rand',
```
bfintal: Why is this hardcoded to `rand`?

bfintal (changes requested):

src/block/posts/index.php
prconcepcion: Oops. That is also another mistake.

PR #3031, opened 2024-01-18
fix (frontend-scripting): prevent stackable frontend scripts from initializing multiple times.

### LINKED ISSUE #3026: Expand / Show more block - create a way to only run script one time upon clicking expand
**Describe the bug**
- Unable to replicate this issue but the issue happens here: https://sove.no/hvordan-velge-riktig-fallunderlag/
- It seems like the script runs 4 times

Other notes (based on the meeting w/ Jami):
- Apply fix to all frontend blocks js
- Not sure if this is a issue with Blocksy or not
- if init is applied multiple times, it should be okay
- 4 click events happen upon clicking the link
- It creates another click handler and it will load initialize again

**Expected behavior**
- Create a safeguard to our script to only run it one time

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/d3a8f287-3889-4151-a122-a26e67d9fc72

prconcepcion:
Blocksy loads frontend scripts multiple times to enable functionalities from dynamic content to work normally. This pr is to prevent our scripts from initializing multiple times. 

Fixes #3026


src/block/expand/frontend-expand.js
```diff
@@ -45,12 +45,18 @@ class StackableExpand {
...
 		} )
 
 		blocks.forEach( block => {
-			fixAriaAttributes( block )
-			addAriaAttributes( block )
+			if ( ! block._StackableHasInitExpand ) {
+				fixAriaAttributes( block )
+				addAriaAttributes( block )
+				block._StackableHasInitExpand = true
+			}
```
bfintal:
```suggestion
			if ( ! block._StackableHasInitExpandFix ) {
				fixAriaAttributes( block )
				addAriaAttributes( block )
				block._StackableHasInitExpandFix = true
			}
```

PR #3033, opened 2024-01-22
fixes (button): button now fills space if button full width option is enabled when button group is full width

### LINKED ISSUE #2991: Button/Button Group block - When setting block to full width, it doesn't max to full width in frontend
**Describe the bug**
- If I set the block to "full width" and set content width alignment to full width on toolbar, the button block is not maxed to full width in frontend
- Issue on frontend only
- Happens on any theme

**To Reproduce**
1. Add a new page
2. Add a Button block
3. Select Button Group block in list view
4. In the toolbar, set content width alignment to **Full width**
5. In Inspector > Layout, enable "Full Width Buttons"
6. Save page
7. See frontend page
8. See that the button block is not max to full width

**Expected behavior**
- When setting Button block to full width, it should max to full width in frontend

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/ff1f7afd-1291-43f4-8a85-01c1bac4326e

**Desktop:**
 - OS: MacOS Sonoma
 - Browser: Chrome
 - Browser Version: 119.0.6045.199

prconcepcion:
fixes #2991

### DIFF
--- a/src/block/button-group/style.js
+++ b/src/block/button-group/style.js
@@ -196,17 +196,6 @@ const Styles = props => {
 					return value ? '1' : undefined
 				} }
 			/>
-			<BlockCss
-				{ ...propsToPass }
-				// This is to make icon buttons stretch.
-				selector=".%s, .stk-block-icon-button .stk-button"
-				styleRule="width"
-				attrName="buttonFullWidth"
-				key="buttonFullWidth-icon-button"
-				valueCallback={ value => {
-					return value ? '100%' : undefined
-				} }
-			/>
 		</>
 	)
 }


bfintal:
@andeng1106 for QA, please see whether older Stackable versions also had this issue.

andeng1106:
> @andeng1106 for QA, please see whether older Stackable versions also had this issue.

@bfintal Doesn't happen on versions 3.7.2 or older. Issue started to happen after New Inspector UI (3.8.0 version and newer).

bfintal:
@andeng1106 Can we have a test whether the full width option fix applies correctly if the button group has different widths (center, wide, full)

andeng1106:
> @andeng1106 Can we have a test whether the full width option fix applies correctly if the button group has different widths (center, wide, full)

@bfintal Checked this one out and it applies correctly.

Screenshot 2024-03-22 at 10 58 53 AM

PR #3037, opened 2024-01-30
Fix: Play Unlisted Vimeo Videos on Video Popup Block

### LINKED ISSUE #3025: Video Popup block - Should be able to play an unlisted Vimeo video
Yes. In our Video Popup block, a user is unable to play an unlisted Vimeo video. Sample unlisted Vimeo video to try: https://vimeo.com/581740669/8cec4c50da

Based on the user:
> The videos my customer wants to present need to work both on the website and from a direct link. The unlisted videos work without problems in the basic Gutenberg video block, so I don't quite see why they won't work in the Stackable pop-up block. The fact that the the block tries to open the same URL without the second number when given the Unlisted URL with two numbers tells me that there is too much sanitization going on. Maybe if the block accepted URLs with two numbers, then the Unlisted videos would work, too?

Should be able to enter and play a URL for an unlisted Vimeo video.

N/A

kaeizen:
fixes #3025

### DIFF
--- a/src/block/video-popup/frontend-video-popup.js
+++ b/src/block/video-popup/frontend-video-popup.js
@@ -32,7 +32,11 @@ const getVideoProviderFromURL = url => {
 	}
 
 	// Check for Vimeo.
-	id = ( url.match( /vimeo\.com\/(\w*\/)*(\d+)/i ) || [] )[ 2 ]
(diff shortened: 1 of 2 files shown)


andeng1106:
@mxkae The unlisted video link provided in the issue description is working properly. But the user has more unlisted videos which still doesn't work :( Here's a Notion page that includes the video links: https://www.notion.so/Robin-Pfeifer-5b03102dddd64366ab0fa9315ec987ed

I tried the first link on the screen recording below:

https://github.com/gambitph/Stackable/assets/103395655/c634364b-bb4f-4f29-88f5-228e4891a9c4

PR #3040, opened 2024-02-01
feat: add thumbnail option for video background

### LINKED ISSUE #3032: Add a poster/thumbnail option for video background
Yes. This can fix the issue with no thumbnail showing.
Here's a summary from user:

> Google Search Console has alerted me that this home page is showing an error because the video does not have a thumbnail. Here is their documentation on video SEO (https://developers.google.com/search/docs/appearance/video).

Add a poster/thumbnail option if video background is uploaded as a background.

N/A

kaeizen:
fixes #3032

### DIFF
--- a/src/block-components/helpers/backgrounds/edit.js
+++ b/src/block-components/helpers/backgrounds/edit.js
@@ -173,6 +173,33 @@ export const BackgroundControls = props => {
 				/>
 			}
 
+			{ hasBackgroundMedia && isBackgroundVideo() &&
+				<ImageControl2
+					label={ __( 'Background Video Thumbnail', i18n ) }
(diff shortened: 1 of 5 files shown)


src/block-components/helpers/backgrounds/edit.js
```diff
@@ -173,6 +173,33 @@ export const BackgroundControls = props => {
 				/>
 			}
 
+			{ hasBackgroundMedia && isBackgroundVideo() &&
```
bfintal: I see that we're now calling `isBackgroundVideo()` multiple times, can we create a const that runs this function once and use that to check?

bfintal (changes requested):

PR #3046, opened 2024-02-08
fix (icon list block alignment): try ui change

bfintal:


bfintal:
Closed in favor of #3047 - that already has the fixes here

PR #3047, opened 2024-02-08
Fix (icon list block): alignment and borders

kaeizen:


src/block/icon-list/edit.js
```diff
@@ -19,7 +19,7 @@ import {
 	AdvancedToolbarControl,
 	AdvancedSelectControl,
 	AlignButtonsControl,
-	InspectorLayoutControls,
+	// InspectorLayoutControls,
```
bfintal:
```suggestion
```

PR #3060, opened 2024-02-14
fix (carousel block): infinite scroll

### LINKED ISSUE #3028: Carousel block - Make the carousel to slide smoothly (infinite scroll) when reaching end then going back to start
Yes. A user wants the Carousel block to slide infinitely once it reaches the end and goes back to the first slide.

As per user:
> Carousel setting, is there a way to make it slide indefinitely (infinite), so always right? Because when it reaches the end it jumps back?

Add a setting to make Carousel slide smoothly.

N/A

kaeizen:
fixes #3028

### DIFF
--- a/src/block/carousel/edit.js
+++ b/src/block/carousel/edit.js
@@ -88,7 +88,6 @@ const Edit = props => {
 	const [ columnProviderValue, columnTooltipClass ] = ColumnInnerBlocks.useContext()
 
 	const carouselType = attributes.carouselType === '' ? 'slide' : attributes.carouselType
-	const infiniteScroll = carouselType === 'slide' && attributes.infiniteScroll
 
 	const blockClassNames = classnames( [
 		className,
@@ -99,7 +98,6 @@ const Edit = props => {
 		{
 			'stk--is-slide': carouselType === 'slide',
 			'stk--is-fade': carouselType === 'fade',
-			'stk--infinite-scroll': infiniteScroll,
 			'stk--hide-mobile-arrows': attributes.showArrowsOnMobile === false,
 			'stk--hide-mobile-dots': attributes.showDotsOnMobile === false,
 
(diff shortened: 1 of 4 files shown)


src/block/carousel/edit.js
```diff
@@ -98,6 +98,7 @@ const Edit = props => {
 		{
 			'stk--is-slide': carouselType === 'slide',
 			'stk--is-fade': carouselType === 'fade',
+			'stk--infinite-scroll': attributes.carouselType === '' && attributes.infiniteScroll,
```
bfintal:
```suggestion
			'stk--infinite-scroll': carouselType === 'slide' && attributes.infiniteScroll,
```

src/block/carousel/frontend-carousel.js
```diff
@@ -166,20 +191,56 @@ class _StackableCarousel {
...
 
 	prevSlide = () => {
 		let newSlide = this.currentSlide - 1
+
+		if ( this.infiniteScroll && newSlide < this.slideOffset ) {
+			this.swapSlides( 'P' )
+			return
+		}
+
 		if ( newSlide < this.slideOffset ) {
```
bfintal: Seeing the `if` condition in the `nextSlide`, would you also need `! this.infiniteScroll` in here?
```suggestion
		if ( ! this.infiniteScroll && newSlide < this.slideOffset ) {
```

kaeizen: Technically, the `! this.infiniteScroll` check won’t be needed here because it is already handled by the first if statement in line 209. Both if statements (in line 209 and 214) have the condition`newSlide < this.slideOffset`, and the first if statement checks `this.infiniteScroll` so reaching the second if statement would imply that `this.infiniteScroll` is false.

In contrast to `nextSlide()`, the second condition for both if statements differs (`newSlide > this.slideEls.length` and `newSlide > this.maxSlides`) so it is necessary to add the `! this.infiniteScroll` to the second if statement.

But, I will add `! this.infiniteScroll` just to be sure.

andeng1106:
@mxkae Remaining issue/s:
- When I turn on infinite scrolling, I had additional dots below. Is this expected?

https://github.com/gambitph/Stackable/assets/103395655/7cab34c3-36d1-448a-9e4a-97ff442db048

- Infinite scrolling effect doesn't apply for Safari & Firefox browsers

https://github.com/gambitph/Stackable/assets/103395655/bbb8be93-9c8f-4661-8ea3-59f3fd9f696f

kaeizen:
> When I turn on infinite scrolling, I had additional dots below. Is this expected?

Yes, this is expected. The additional dots correspond to the new slide "groups" for the infinite scrolling. For example, if we have 4 slides and 3 as the number of slides to show, then:

- first dot displays slides 1,2,3
- second dot displays slides 2,3,4 (Note: shows only up to this point if infinite scrolling is disabled)
- third dot displays slides 3,4,1 (new slide group)
- and fourth dot displays slides 4,1,2 (new slide group)

PR #3072, opened 2024-02-21
feat (columns): allow column wrapping in desktop

bfintal:
Added a new option called **Allow Column Wrapping** in columns: 
image

Enabling it will change the column width controls to allow the columns to wrap below one another similar to how columns work in tablet & mobile.

We added a new attribute `columnWrapDesktop` for the columns block to allow wrapping in desktops.

#### TODO

- [ ] When Allow Column Wrapping is enabled, the column resizers should behave similar to how the resizers work in tablet & mobile (they shouldn't auto-compute to 100% total column widths) - check the ResizableColumn component
- [ ] Unsure: When changing individual columns, if you do not supply a value, the widths look broken - we do not need to really handle this scenario since we cannot assume what the user is trying to do, but maybe there's a way not to make it look broken e.g.
  - 3 columns: 50%, 50%, blank 
image
  - 3 columns: blank, 50%, 50% 
image

### DIFF
--- a/src/block-components/column/attributes.js
+++ b/src/block-components/column/attributes.js
@@ -12,6 +12,10 @@ export const addAttributes = attrObject => {
 				type: 'number',
 				default: '',
 			},
+			columnWrapDesktop: {
+				type: 'boolean',
+				default: false,
+			},
 		},
 		versionAdded: '3.0.0',
 		versionDeprecated: '',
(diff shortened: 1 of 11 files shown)


andeng1106:
Remaining issues found @mxkae:

- [x]  Allow Column Wrapping - When enabling then disabling the option, the Column Width percentage resets.

https://github.com/gambitph/Stackable/assets/103395655/6e0f6a4a-2e13-4813-92d4-9b02d5b77706

- [x] Feature block & Feature Grid block - When adjusting column width, the second/next column doesn't wrap unless I use the slide or input value in the Inspector

https://github.com/gambitph/Stackable/assets/103395655/78924b3a-4658-400e-a3c9-f3b12621759d

PR #3104, opened 2024-03-11
fix: Refactor usage of __unstableGetClientIdsTree

### LINKED ISSUE #2798: Refactor usage of __unstableGetClientIdsTree before WordPress 6.4
`__unstableGetClientIdsTree` of `wp.data.select( 'core/block-editor' ).__unstableGetClientIdsTree()` will be removed from WordPress in 6.5, we need to remove it by 6.4

Currently, we are using `__unstableGetClientIdsTree` in 2 areas:

1. Navigator panel
2. `useBlockContext` hook

We will need to remove our dependency from this experimental function. I think we can do it the following way.

**Navigator panel**

Options:

1. We can re-create this method because we really need to get the entire tree for the Navigator to work.
2. We can discuss next time whether we can remove this Navigator feature because more and more people are using the Block Editor and there might not be any need for this feature now.

**`useBlockContext` hook**

We created the `useBlockContext` hook to increase the speed of looking up the surrounding context of a block (it's siblings, parent block, index, etc). From our performance release before, we found out that using the native methods were slower than expected.

Options:

1. We can re-create the `__unstableGetClientIdsTree` method - not too sure about this because the benefit of using `__unstableGetClientIdsTree` is that it's already using the data in the core store - we might lose efficiency by recreating our own
2. We can revisit the code that use `useBlockContext` and see whether we can refactor the use of the hook out.

kaeizen:
fixes #2798


src/plugins/get-client-id-tree/index.js
```diff
@@ -0,0 +1,33 @@
...
+		if ( count !== blockCount ) {
+			setBlockCount( count )
+		}
+	} )
+
+	useEffect( () => {
+		if ( blockCount ) {
+			dispatch( 'stackable/block-editor' ).updateClientTree()
+		}
+	}, [ blockCount ] )
```
bfintal: Can we see if using `getClientIdsWithDescendants` will help with this?
```suggestion
	const allClientIds = useSelect( select => {
		return select( 'core/block-editor' ).getClientIdsWithDescendants()
	} )

	useEffect( () => {
		console.log('did update' )
		dispatch( 'stackable/block-editor' ).updateClientTree()
	}, [ allClientIds ] )
```

mxkae:
`blocks` in `<ListView>` is deprecated since version 6.3 as shown below so there's no need to get the inner blocks using the client tree anymore.

Screenshot 2024-03-20 at 2 10 14 PM

bfintal (changes requested):
Now we have 2 stores for this purpose:
1. `src/hooks/use-block-context.js`
2. `src/plugins/get-client-id-tree/store.js`

I believe both of these are used solely for updating block context, so we should be able to combine these into just one.

andeng1106:
@mxkae Remaining issue: 
- Related to https://github.com/gambitph/Stackable/issues/2822

https://github.com/gambitph/Stackable/assets/103395655/52930abe-4f43-4e54-b8d6-2e9baac99163

cc @bfintal

PR #3121, opened 2024-03-25
fix (video popup): change button tag with div tag

### LINKED ISSUE #3055: Stackable adds a button HTML element when using the video popup, creates a box-shadow when on mouse over
**_Note: This issue is from the affiliate who encountered bugs when testing out Stackable_**

Reference: https://www.notion.so/Stackable-adds-a-button-HTML-element-when-using-the-video-popup-creates-a-box-shadow-when-on-mouse--cd4fcf6828f046fc8f807481f6b6eaa7

For the video popup, Stackable adds a "button" HTML element for the video poster (which is probably not a very good idea as it can lead to style issues) which resulted in a box-shadow on mouse over (coming from the Kadence free theme). Had to write some CSS to remove this box-shadow.

**Fix**

We need to replace the `button` tag with a `div` tag with the following attributes `tabindex="0" role="button"` and add the necessary deprecation steps for it.

kaeizen:
fixes #3055

### DIFF
--- a/src/block/video-popup/deprecated.js
+++ b/src/block/video-popup/deprecated.js
@@ -2,12 +2,37 @@ import { Save } from './save'
 import { attributes } from './schema'
 
 import { withVersion } from '~stackable/higher-order'
+import { semverCompare } from '~stackable/util'
+import { i18n } from 'stackable'
 import {
 	deprecateBlockBackgroundColorOpacity, deprecateContainerBackgroundColorOpacity,
 	deprecateBlockShadowColor, deprecateContainerShadowColor,
 } from '~stackable/block-components'
 
+/**
+ * WordPress dependencies
+ */
+import { InnerBlocks } from '@wordpress/block-editor'
+import { addFilter } from '@wordpress/hooks'
+import { __ } from '@wordpress/i18n'
+
+// Version 3.0.2 Deprecations
+addFilter( 'stackable.video-popup.save.div.content', 'stackable/3.12.12', ( output, props, contentClassNames ) => {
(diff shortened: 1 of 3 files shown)


src/block/video-popup/deprecated.js
```diff
@@ -2,12 +2,36 @@ import { Save } from './save'
...
+			</button>
+		)
+	}
+	return output
+} )
+
 const deprecated = [
+	{
+		attributes: attributes( '3.12.12' ),
+		save: withVersion( '3.12.12' )( Save ),
```
bfintal:
```suggestion
		attributes: attributes( '3.12.14' ),
		save: withVersion( '3.12.14' )( Save ),
```

src/block/video-popup/deprecated.js
```diff
@@ -2,12 +2,36 @@ import { Save } from './save'
...
 
+/**
+ * WordPress dependencies
+ */
+import { InnerBlocks } from '@wordpress/block-editor'
+import { addFilter } from '@wordpress/hooks'
+import { __ } from '@wordpress/i18n'
+
+addFilter( 'stackable.video-popup.save.div.content', 'stackable/3.12.12', ( output, props, contentClassNames ) => {
+	if ( semverCompare( props.version, '<', '3.12.12' ) ) {
```
bfintal:
```suggestion
addFilter( 'stackable.video-popup.save.div.content', 'stackable/3.12.14', ( output, props, contentClassNames ) => {
	if ( semverCompare( props.version, '<', '3.12.14' ) ) {
```

PR #3124, opened 2024-04-01
feat (tabs block): added tab anchors

### LINKED ISSUE #3067: Allow tab to open depending on the url hash
Add a new option to the **Tabs block** under **Advanced Tab > General** named "**Tab Anchors**" (if not possible in the Tabs block, place it in the **Tab Labels block**)

- The **Tab Anchors** option shows one text input per tab present.
- If present, when the tab is clicked, it will also update the URL adding in the anchor for that tab
- If present, when the URL is changed and the anchor is added, the corresponding tab should open if it isn't yet (and also should be scrolled to if not visible)

kaeizen:
fixes #3067

### DIFF
--- a/src/block/tab-labels/edit.js
+++ b/src/block/tab-labels/edit.js
@@ -625,11 +625,17 @@ const Edit = props => {
 										const updatedLabels = cloneDeep( tabLabels )
 										updatedLabels[ index ].anchor = value
 										setTabLabels( updatedLabels )
+
+										if ( ! value ) {
+											setAttributes( { tabLabels: updatedLabels } )
+										}
 									} }
 									onBlur={ () => {
 										const updatedLabels = tabLabels
-										updatedLabels[ index ].anchor = kebabCase( updatedLabels[ index ].anchor )
-										setAttributes( { tabLabels: updatedLabels } )
+										if ( updatedLabels[ index ].anchor ) {
+											updatedLabels[ index ].anchor = kebabCase( updatedLabels[ index ].anchor )
+											setAttributes( { tabLabels: updatedLabels } )
+										}
 									} }
 								/>
 							) ) }
--- a/src/block/tab-labels/edit.js
+++ b/src/block/tab-labels/edit.js
@@ -603,7 +603,7 @@ const Edit = props => {
 							title={ __( 'Tab Anchors', i18n ) }
 							id="tabAnchors"
 						>
-							<GutBaseControl help={ __( "Note: Assign unique anchor names to each tab so you'll be able to link directly and open each one.", i18n ) } />
+							<GutBaseControl help={ __( "Assign unique anchor names to each tab so you'll be able to link directly and open each one.", i18n ) } />
 							{ props.attributes.tabLabels.map( ( tab, index ) => (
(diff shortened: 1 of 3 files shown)


src/block/tab-labels/edit.js
```diff
@@ -592,6 +617,34 @@ const Edit = props => {
...
+							{ props.attributes.tabLabels.map( ( tab, index ) => (
+								<AdvancedTextControl
+									label={ sprintf(
+									// Translators: %s is the tab label.
+										__( '%s Anchor', i18n ),
+										// eslint-disable-next-line @wordpress/i18n-no-variables
+										__( tab.label, i18n )
+									) }
+									value={ tabLabels[ index ].anchor }
+									placeholder="Tab Anchor"
```
bfintal: This needs to be translatable
```suggestion
									placeholder="Tab Anchor"
```

src/block/tab-labels/edit.js
```diff
@@ -592,6 +617,34 @@ const Edit = props => {
 					</InspectorStyleControls>
 
 					<BlockDiv.InspectorControls />
+
+					<InspectorAdvancedControls>
+						<PanelAdvancedSettings
+							title={ __( 'Tab Anchors', i18n ) }
+							id="tabAnchors"
+						>
+							{ props.attributes.tabLabels.map( ( tab, index ) => (
```
bfintal: Add a help note before the fields saying "Assign unique names to each tab so you'll be able to link directly and open each one."

bfintal (changes requested):
Comments:
- Update the labels to be kebabCase on blur instead of timeout because if you edit the middle of the string, the cursor will jump to the end.
- The tab anchor doesn't work if linking directly from a button on the same page. e.g. if you have a button/link that links to `#tab-two`, the 2nd tab will not open when you click on the button/link

bfintal (changes requested):
I'm setting this for QA but here are 2 comments that need to be addressed:

1. Clicking on a tab that has an anchor should not scroll the browser (to test, add a 1000px spacer block above and below the tabs block, then click on a tab that has an anchor)

2. Refreshing the browser that has a tab anchor link should scroll to the tab. Currently, the tab is open, but it doesn't scroll to it (to test, have a 1000px spacer block before the tabs block, then visit the url with the has tab hash in another window)

andeng1106:
@mxkae some issues found:

- [x] When typing in the tabs anchor fields, it seems laggy if I have nested blocks in the Tabs block (not sure if this is related to the Tabs slowdown?)

https://github.com/gambitph/Stackable/assets/103395655/bb7e5eb2-7ebb-4362-af2f-64c9a22ea1c8


- [x] [Incosistent] Adding / resetting tabs anchor field sometimes doesn't register the change (Update / Save Draft button doesn't highlight)
     - Unable to figure out what's causing it and hard to catch but might be related to the Reset button. Here are some sample videos how it happens:

https://github.com/gambitph/Stackable/assets/103395655/a4b906ee-9de0-49a9-a294-c065d1ab750a


https://github.com/gambitph/Stackable/assets/103395655/8da05cc2-e201-472e-b096-2ab08a26f7a4

PR #3136, opened 2024-04-04
feat (accordion block): added anchor links

### LINKED ISSUE #2888: Accordion block - When navigating to Accordion block through HTML anchor link, it should be automatically opened
Yes. User wants the Accordion block to automatically open when typing the HTML anchor in the URL link upon page load.
Reference: 
- https://www.facebook.com/groups/wpstackable/posts/1279939306227913/
- https://www.facebook.com/groups/wpstackable/permalink/1282148709340306/?mibextid=2JQ9oc&paipv=0&eav=AfalBL-kU4As_LUHzH3yamZ9H4S-FfDMFzjyjVk-bWSlEGOsf5OvBmFmx3O6ulyU-D4&_rdr

Make Accordion block to automatically expanded upon accessing the HTML anchor link

Add HTML anchor for the Accordion block and enable "Open at the Start" setting.

kaeizen:
fixes #2888

### DIFF
--- a/src/block-components/advanced/edit.js
+++ b/src/block-components/advanced/edit.js
@@ -5,19 +5,22 @@ import { i18n } from 'stackable'
 import {
 	AdvancedRangeControl,
 	AdvancedSelectControl,
+	AdvancedTextControl,
 	FourRangeControl,
 	InspectorAdvancedControls,
 	PanelAdvancedSettings,
 } from '~stackable/components'
-import { useBlockAttributesContext } from '~stackable/hooks'
+import { useBlockAttributesContext, useBlockSetAttributesContext } from '~stackable/hooks'
 
 /**
  * WordPress dependencies
  */
+import { dispatch } from '@wordpress/data'
 import { Fragment } from '@wordpress/element'
 import {
 	__, _x, sprintf,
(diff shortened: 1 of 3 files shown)


src/block-components/advanced/edit.js
```diff
@@ -52,6 +58,19 @@ export const Edit = () => {
 					title={ __( 'General', i18n ) }
 					id="general"
 				>
+					{ allowCustomAnchor && (
```
bfintal: Ideally, we should only add new things to block components if they are usable by other blocks, but in this case, this option is specific only to the Accordion block. Structurally, this option should not be present here.

That being said, since this option is just an Anchor to the Accordion block, we do not need to create a new option for this because WP already added one under **Advanced Tab > Advanced Panel** as the `anchor` attribute. Let's use that attribute and option instead.

bfintal (changes requested):
Some comments when I tried out the new feature:
1. When opening/closing the accordion with an anchor, it shouldn't scroll the page
2. When I opened the url with a hash in another window, the accordion didn't open
3. When I opened the url with a hash in another window, the page didn't scroll to the accordion

PR #3178, opened 2024-05-06
fix (performance): improvements based on code profiler part 2

### LINKED ISSUE #3172: Performance Optimizations (part 2)
# Optimizations To Do:

## Remove `stackable_add_custom_orderby_params` and `stackable_add_custom_orderby`

- [ ] Done

These 2 filters are only here because we need to add **random** and **menu order** sorting options in the Stackable Posts block. However, it's slowing down WP because every time posts are used, the filters are also applied.

I suggest removing these 2 filters and creating our own "Post getter" rest api endpoint for the Stackable Posts block. If we create our own, then there will be no need to add filters and extend the post rest api with our own filters. If this is done, this can also fix the **next entry below**.

Here's a screenshot of a profiler session when saving a post

Image

---

## Create a new page and add a tabs block with 5 tabs containing the following: pricing table with icon list of different icons. Profile the editor URL, you'll see that the code for `get_excerpt_by_post_id` is slow:

- [ ] Done

image

Upon further investigation, it appears that in `src/block/posts/index.php` we are registering a bunch of new rest fields for posts in `register_rest_fields`. By registering new rest fields, ALL rest API calls for posts will now compute for these new fields (even when the editor just loads). 

We need to change this behavior because the added data are **only needed** during the generation of the Stackable post block's content **in the editor** here: https://github.com/gambitph/Stackable/blob/ccd91e3d0a4aad5204d701cb3fd1aea8c4ccafc9/src/block/posts/edit.js#L152

Because the field is added in `register_rest_fields`, now whenever posts are queried, it also calls our functions.  We need to change the following:

1. remove the added fields from `register_rest_fields` or only add them when generating Stackable post block content
2. or, the implementation of `generateRenderPostItem` and `usePostsQuery` to query a new rest API endpoint that grabs the necessary post data for displaying in the post block content.

---

## Create a new page and add a posts block. Profile the editor URL, you'll see that the posts block code is quite slow:

- [ ] Done

image

---

## ~Change block registration method. To check, profile the Settings > General page.~ **Do not do this for now**

- [ ] Done

Our blocks are registered using the function `register_block_type_from_metadata` it's reading the `block.json` files and registering it using the properties found in those files. I think this might be slow.

image

We can try placing all the relevant data from `block.json` files into a PHP object (maybe with a script that auto-compiles things if we prove this to be a good optimization strategy) then use `register_block_type` to register the blocks instead. This way we do not need to do any file read/writes and might bring down the execution time.

kaeizen:
fixes [#3172](https://github.com/gambitph/Stackable/issues/3172)


src/blocks.php
```diff
@@ -143,7 +35,9 @@ function stackable_register_blocks() {
 				$metadata
 			);
 
-			register_block_type_from_metadata( $metadata['block_json_file'], $register_options );
+			$block_args = array_merge( $metadata, $register_options );
+			$block_type = new WP_Block_Type( $metadata['name'], $block_args );
+			register_block_type( $block_type );
```
bfintal: Any reason why we can't just register the block right away?

```php
register_block_type( $metadata['name'], $block_args );
```

bfintal (changes requested):
Comments:
- Please also commit the built file `src/stk-block-types.php` so the plugin would work even by just doing a checkout

src/blocks.php
kaeizen: based on this [documentation](https://developer.wordpress.org/reference/functions/register_block_type/) , `register_block_type` accepts a string or an instance of `WP_Block_Type` for the first argument, if it's a string it will consider it as a path to the `block.json` and will try to use `register_block_type_from_metadata()` instead.

PR #3179, opened 2024-05-06
fix: uploaded svg icon not showing

### LINKED ISSUE #3103: SVG icon only shows up on some blocks with icon
**Describe the bug**
- Related to: https://www.notion.so/Karin-Bronwasser-6b47398388f941c1a2b59e85633053bf
- This issue happens for blocks such as:
     - Icon List
     - Carousel
     - Map
- Bullet svg icon: https://drive.google.com/file/d/1PRTecd1hwVE70b72B2fAZQ-cMhPMr4EO/view?usp=sharing
- Make sure to set the Icon Color to darker color to see if it shows or not

**To Reproduce**
1. Add a new page (or skip steps and use this [page](https://gambitsandbox.instawp.xyz/andrea/icon-list-button-block-issue/)
2. Add any of the blocks mentioned above
3. Upload the SVG icon from the link above
4. Try to change the Icon Color to any color (e.g. Black)
5. See that the icon doesn't show up

**Expected behavior**
SVG icons should display correctly for Icon List, Carousel block & Map block

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/06a10fbd-3cf5-452b-aa42-457b61578d37

Arukuen:
fixes #3103

Details of the issue and the solution: [https://www.notion.so/SVG-icon-only-shows-up-on-some-blocks-with-icon-3103-94b9c7925d1f4eaba8ebb2752f4b1660?pvs=4](https://www.notion.so/SVG-icon-only-shows-up-on-some-blocks-with-icon-3103-94b9c7925d1f4eaba8ebb2752f4b1660?pvs=4)

TLDR: The SVG that can be uploaded by the users might have inline style which takes precedence specially since this can be applied to the lowest level which is the `<path>` element. This PR fixes the issue by applying the `fill` directly to `<path>` to have higher precedence.

### DIFF
--- a/src/block/carousel/style.js
+++ b/src/block/carousel/style.js
@@ -109,6 +109,15 @@ const Styles = props => {
 			/>
 			<BlockCss
 				{ ...propsToPass }
+				selector=".stk-block-carousel__button svg.ugb-custom-icon path"
+				hoverSelector=".stk-block-carousel__button svg.ugb-custom-icon path:hover"
+				styleRule="fill"
+				attrName="arrowIconColor"
+				key="arrowIconColor"
+				hover="all"
+			/>
+			<BlockCss
+				{ ...propsToPass }
 				styleRule="--button-width"
 				attrName="arrowWidth"
 				key="arrowWidth"
--- a/src/block/carousel/style.js
+++ b/src/block/carousel/style.js
@@ -109,8 +109,8 @@ const Styles = props => {
 			/>
 			<BlockCss
 				{ ...propsToPass }
-				selector=".stk-block-carousel__button svg.ugb-custom-icon path"
-				hoverSelector=".stk-block-carousel__button svg.ugb-custom-icon path:hover"
+				selector=".stk-block-carousel__button svg.ugb-custom-icon :is(g,path,rect,polygon,ellipse)"
+				hoverSelector=".stk-block-carousel__button svg.ugb-custom-icon :is(g,path,rect,polygon,ellipse):hover"
 				styleRule="fill"
 				attrName="arrowIconColor"
 				key="arrowIconColor"
--- a/src/block/icon-list-item/style.scss
+++ b/src/block/icon-list-item/style.scss
@@ -18,7 +18,7 @@
 	}
 
 	// Apply the color to the SVG path on the defintion
-	svg.ugb-custom-icon path {
+	svg.ugb-custom-icon :is(g,path,rect,polygon,ellipse) {
 		fill: var(--stk-icon-list-marker-color) !important;
 		color: var(--stk-icon-list-marker-color) !important;
 	}
--- a/src/block/icon-list-item/style.scss
+++ b/src/block/icon-list-item/style.scss
@@ -16,6 +16,12 @@
 		border-width: 1px;
 		border-color: rgba(0, 0, 0, 0.4);
 	}
+
+	// Apply the color to the SVG path on the defintion
+	svg.ugb-custom-icon path {
+		fill: var(--stk-icon-list-marker-color);
+		color: var(--stk-icon-list-marker-color);
+	}
 }
 
 .stk-block-icon-list-item__content {
--- a/src/block/icon-list-item/style.scss
+++ b/src/block/icon-list-item/style.scss
@@ -19,8 +19,8 @@
 
 	// Apply the color to the SVG path on the defintion
 	svg.ugb-custom-icon path {
-		fill: var(--stk-icon-list-marker-color);
-		color: var(--stk-icon-list-marker-color);
+		fill: var(--stk-icon-list-marker-color) !important;
+		color: var(--stk-icon-list-marker-color) !important;
 	}
 }
 
--- a/src/block/map/save.js
+++ b/src/block/map/save.js
@@ -29,6 +29,7 @@ import { i18n, version as VERSION } from 'stackable'
 import { __ } from '@wordpress/i18n'
 import { compose } from '@wordpress/compose'
 import { RawHTML } from '@wordpress/element'
+import { applyFilters } from '@wordpress/hooks'
 
 export const Save = props => {
 	const {
@@ -90,7 +91,14 @@ export const Save = props => {
 					<div
 						data-map-options={ JSON.stringify( mapOptions ) }
 						data-marker-options={ JSON.stringify( markerOptions ) }
-						data-icon-options={ JSON.stringify( getIconOptions( attributes ) ) }
+						data-icon-options={ JSON.stringify(
+							applyFilters(
+								'stackable.map.util.applySVGAttributes',
+								getIconOptions( attributes, false ),
+								attributes,
+								props
+							)
+						) }
 						className="stk-block-map__canvas"
 					/>
 				)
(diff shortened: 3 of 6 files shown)


Arukuen:
Will ask API Key to test solution for `Map`

src/block/icon-list-item/style.scss
```diff
@@ -16,6 +16,12 @@
 		border-width: 1px;
 		border-color: rgba(0, 0, 0, 0.4);
 	}
+
+	// Apply the color to the SVG path on the defintion
+	svg.ugb-custom-icon path {
+		fill: var(--stk-icon-list-marker-color);
+		color: var(--stk-icon-list-marker-color);
+	}
```
bfintal: In this same file, we are already defining similar styles to the `.stk--inner-svg svg` and `.stk-block-icon-list-item__marker::before`, is it possible to just add the `.ugb-custom-icon` selector to those so that custom icons would also get the styles?

src/components/icon-search-popover/index.js
```diff
@@ -71,6 +71,10 @@ export const cleanSvgString = svgString => {
 		newSvg = newSvg.replace( /\s*<g\s*>([\s\S]*?)<\/g>\s*/gm, '$1' )
 	}
 
+	// Remove width and height attribute so that we can control the size (default to 100%)
+	newSvg = newSvg.replace( /width="[^"]*"/, '' )
+		.replace( /height="[^"]*"/, '' )
```
bfintal: 1. Can you replace attributes only if they are inside the `<svg>` tag, so we can be sure that other tags won't be affected.
2. Please support cases where single quotes are used e.g. `width='100'`
```suggestion
	newSvg = newSvg.replace( /width="[^"]*"/, '' )
		.replace( /height="[^"]*"/, '' )
```

src/block/icon-list-item/style.scss
```diff
@@ -16,6 +16,12 @@
 		border-width: 1px;
 		border-color: rgba(0, 0, 0, 0.4);
 	}
+
+	// Apply the color to the SVG path on the defintion
+	svg.ugb-custom-icon path {
```
bfintal:
```suggestion
	svg.ugb-custom-icon :is(g,path,rect,polygon,ellipse) {
```

src/block/carousel/style.js
```diff
@@ -107,6 +107,15 @@ const Styles = props => {
 				key="arrowIconColor"
 				hover="all"
 			/>
+			<BlockCss
+				{ ...propsToPass }
+				selector=".stk-block-carousel__button svg.ugb-custom-icon path"
+				hoverSelector=".stk-block-carousel__button svg.ugb-custom-icon path:hover"
```
bfintal:
```suggestion
				selector=".stk-block-carousel__button svg.ugb-custom-icon :is(g,path,rect,polygon,ellipse)"
				hoverSelector=".stk-block-carousel__button svg.ugb-custom-icon :is(g,path,rect,polygon,ellipse):hover"
```

src/block/icon-list-item/style.scss
```diff
@@ -16,6 +16,12 @@
...
+		fill: var(--stk-icon-list-marker-color) !important;
+		color: var(--stk-icon-list-marker-color) !important;
```
bfintal, line 23: Do we need `color`? In the Icon block, `color` isn't used

src/block/map/util.js
```diff
@@ -64,6 +64,12 @@ export const getIconOptions = attributes => {
 	svgEl.setAttribute( 'style', `color: ${ iconColor1 || DEFAULT_ICON_COLOR }; opacity: ${ iconOpacity || iconOpacity === 0 ? parseFloat( iconOpacity, 10 ) : DEFAULT_ICON_OPACITY }` )
 	svgEl.setAttribute( 'transform', `rotate(${ iconRotation || iconRotation === 0 ? parseInt( iconRotation, 10 ) : DEFAULT_ICON_ROTATION })` )
 
+	// Apply the fill directly to the path elements
+	const pathElements = svgEl.querySelectorAll( 'path' )
```
bfintal: Also do other tags: `g,path,rect,polygon,ellipse`

bfintal: Only do this if there is a color assigned

bfintal (changes requested):

andeng1106:
@Arukuen Issue has been fixed on the three blocks affected. However, there's one remaining issue for Map block only:

- Block errors are encountered on Map block when using the pr build. See error in console.
  - Block error happens for existing Map blocks with icons selected from icon library / uploaded an svg icon
  - Steps to replicate:
     1. On 3.12.17, add a Map block
     2. Add a Map Marker
     3. Upload an svg icon or select from icon selection
     4. Save page
     5. Upgrade to pr build
     6. Refresh editor
     7. See block error
  Screenshot 2024-05-16 at 9 46 27 AM

Arukuen:
@andeng1106  Since the SVG is serialized for map block, the `iconURL` of the generated icon will also change when the application of attribute (which this PR did to apply the fill to SVG shapes) is changed. That is why the value of the `url` property for the `data-icon-option` for the production build and the PR build are different which causes the issue.

To address this, we can advise affected users to use the "Attempt Block Recovery" feature, which successfully reformats the URL to match the current build and resolves the block error. However, I would like ask for suggestion if there is a better approach.

bfintal:
@Arukuen since we are encountering a block error that can be recovered, we'll need to add some deprecation code to handle migration of the icon. The expected behavior is for the map block not to encounter an error upon upgrading to this new build

bfintal (changes requested):
It's great that deprecation is already handled, but it's also a good idea to further separate our deprecated code so it doesn't mingle with our up-to-date functionality. Currently, our deprecated code is located inside the `util.js` code:

```js
export const getIconOptions = ( attributes, isDeprecated ) => {
    // ...
    if ( isDeprecated ) {
        // ...
```

I recommend reverting back to the original, then create a function specific for the needs of the deprecation inside `src/block/map/deprecated.js` that uses the original `getIconOptions` instead. This way, all our deprecated code stays out of our main code.

```js
// file: src/block/map/deprecated.js
import { getIconOptions } from './utils'

const getIconOptions_3_13_0 = attributes => {
    const newAttributes = { ...attributes }
    // Do things required for the deprecation...
    return getIconOptions( newAttributes )
}

addFilter( 'stackable.map.util.applySVGAttributes', 'stackable/3.13.0', ( output, attributes, props ) => {
	if ( semverCompare( props.version, '<', '3.13.0' ) ) {
		return getIconOptions_3_13_0( attributes )
	}
	return output
} )
```

src/block/map/save.js
```diff
@@ -90,7 +91,14 @@ export const Save = props => {
 					<div
 						data-map-options={ JSON.stringify( mapOptions ) }
 						data-marker-options={ JSON.stringify( markerOptions ) }
-						data-icon-options={ JSON.stringify( getIconOptions( attributes ) ) }
+						data-icon-options={ JSON.stringify(
+							applyFilters(
+								'stackable.map.util.applySVGAttributes',
```
bfintal: I recommend changing the name of this filter because if you look at where this filter is being used, it's not related to `util.applySVGAttributes`. I feel like `stackable.map.icon-options` makes more sense here:

```suggestion
								'stackable.map.icon-options',
```

Also change the corresponding code that uses this filter

bfintal (changes requested):

PR #3180, opened 2024-05-07
feat: add warning if content-only-editing

### LINKED ISSUE #2911: Role Manager - Add note if role manager is set to "content only editing"
Yes. Related to this issue: https://www.facebook.com/groups/wpstackable/permalink/1312199566335220/?mibextid=2JQ9oc

When the role manager has content only editing for the user, it should display this warning in the browser console: 
`Stackable: Content only editing is enforced on the current user, you can only see text blocks in the inserter.`

N/A

Arukuen:
fixes #2911

### DIFF
--- a/src/plugins/design-library-button/index.js
+++ b/src/plugins/design-library-button/index.js
@@ -18,6 +18,8 @@ import { subscribe } from '@wordpress/data'
 const mountDesignLibrary = () => {
 	// Content only editing mode shouldn't have a button.
 	if ( isContentOnlyMode ) {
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
The warning is good, but this is in the wrong location. Since this is about "content only editing", then it should be in the `pro__premium_only/src/plugins/editor-mode/index.js` file.

Add the warn right after this line in that file:

```js
if ( isContentOnlyMode ) {
    // here
```

PR #3182, opened 2024-05-12
Fix/accordion highlight

### LINKED ISSUE #3096: Accordion block - When expanding accordion, it gets highlighted on frontend
**Describe the bug**
- No certain option/setting where this can be disabled
- This issue happens if Accordion block is:
    - Outside Stackable Columns block
    - Inside Group/Columns block

https://github.com/gambitph/Stackable/assets/103395655/38847133-a4e6-4b2b-9694-0ea98e5bd953

- This issue causes the highlight in the Accordion (Shadow layout) to glitch if inside the Columns block

https://github.com/gambitph/Stackable/assets/103395655/4d2b0132-0ba0-4616-ab0e-b411428c01f7

Arukuen:
fixes #3096

- No outline when clicked, but outline when tabbed for accessibility

### DIFF
--- a/src/block/accordion/frontend-accordion.js
+++ b/src/block/accordion/frontend-accordion.js
@@ -30,6 +30,10 @@ class StackableAccordion {
 					el.doAnimate = false
 					const preHeight = el.dataset.preHeight
 
+					// Prevent text selection while animating
+					const content = el.querySelector( '.stk-block-accordion__content' )
+					content.style.userSelect = 'none'
+
 					// When inside columns, flex prevents the accordion closing animation, this hack fixes it.
 					const doWrapHack = !! el.closest( '.stk-block-columns' )
 					let wrapper = null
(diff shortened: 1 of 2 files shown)


src/block/accordion/frontend-accordion.js
```diff
@@ -30,6 +30,10 @@ class StackableAccordion {
 					el.doAnimate = false
 					const preHeight = el.dataset.preHeight
 
+					// Prevent text selection while animating
+					const content = el.querySelector( '.stk-block-accordion__content' )
+					content.style.userSelect = 'none'
```
bfintal: It seems that doing this on the content would still make the text highlight. I would recommend doing this on the element itself to prevent this:
```suggestion
					// Prevent text selection while animating
					el.style.userSelect = 'none'
```

src/block/accordion/frontend-accordion.js
```diff
@@ -51,6 +55,22 @@ class StackableAccordion {
...
+							// eslint-disable-next-line @wordpress/no-global-get-selection
+							const selection = window.getSelection()
+							if ( selection.removeAllRanges ) {
+								selection.removeAllRanges()
+							}
+						} else if ( document.selection ) {
+							// For older IE browsers
+							document.selection.empty()
+						}
+						content.style.userSelect = 'auto'
```
bfintal: I see that you are clearing the selection because the `userSelect` above was not taking full effect and you can still highlight text when opening an accordion. Applying the style directly on the element fixes things, and now you do not need all of the selection removal code.

Also, to me, if the fix only partially works and you needed to add some selection removal code to compensate is an indicator that there should be a cleaner solution 👍

```suggestion
						el.style.userSelect = 'auto'
```

bfintal (changes requested):
Good solution, just a suggestion for improvement.

PR #3196, opened 2024-05-26
feat (accordion): open accordions when printing

### LINKED ISSUE #3186: Accordion block - Accordion should be expanded upon printing of page
- User wants to print a page with Accordion block that's expanded upon print. See issue [here](https://www.notion.so/Adam-Frazier-349c3f66e9ef4e888c379661e0bf0b30?pvs=4).
- Notes from Jami: We’re using the browser tag “details” for the accordion, but it seems that browsers haven’t yet added support to automatically open them up for printing - looks like it’s still being worked on.

The only solution would be to use some Javascript. See possible solutions [here](https://www.notion.so/Adam-Frazier-349c3f66e9ef4e888c379661e0bf0b30?pvs=4#d7d0e019837f4e66b11ed725785ce9a7)

N/A

Arukuen:
fixes #3186

### DIFF
--- a/src/block/accordion/frontend-accordion.js
+++ b/src/block/accordion/frontend-accordion.js
@@ -199,4 +199,24 @@ class StackableAccordion {
 }
 
 window.stackableAccordion = new StackableAccordion()
+
+// Open closed accordions when printing
+// and close them again after printing
+window.matchMedia( 'print' ).addEventListener( 'change', event => {
+	if ( event.matches ) {
+		const els = document.querySelectorAll( 'details.stk-block-accordion:not([open])' )
+		for ( const el of els ) {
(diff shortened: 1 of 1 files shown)


src/block/accordion/frontend-accordion.js
```diff
@@ -199,4 +199,24 @@ class StackableAccordion {
 }
 
 window.stackableAccordion = new StackableAccordion()
+
+// Open closed accordions when printing
+// and close them again after printing
+window.matchMedia( 'print' ).addEventListener( 'change', event => {
```
bfintal:
```suggestion
window?.matchMedia( 'print' ).addEventListener( 'change', event => {
```

PR #3199, opened 2024-05-30
fix: Ignore invalid CSS that affects the CSS of subsequent blocks

### LINKED ISSUE #3164: Entering invalid Custom CSS can break block styles
**Describe the bug**
- Related to this [issue](http://surl.li/svegj)
- Issue on frontend only

**To Reproduce**
1. Add a new page (or skip steps and use this [test page](https://gambitsandbox.instawp.xyz/andrea/wp-admin/post.php?post=605&action=edit) instead)
2. Add a Heading block
3. Highlight text by using Highlight & Color
4. Add this Custom CSS for the Heading block:
```
.stk-block-heading{
.stk-highlight{
	border-radius:10px;
}
}
```
5. Add a Carousel block with at least 4 slides
6. Set "Slides to Show" to 4
7. Save page as draft and preview in frontend
8. See that the slides are 4 each time
9. Go back to editor and publish page
10. View frontend
11. See that Carousel slide is only 1 each time

**Expected behavior**
- ~~If invalid CSS, it might not work in the frontend but it should not break blocks~~
- ~~If possible, display a warning message in the inspector that css added might be invalid~~
- Using the CSS above should not make the block look broken in the frontend.

**Additional notes**

Note that if you only have a single heading block, using the example Custom CSS above doesn't break the Heading block in the frontend and still displays the block correctly.

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/58ab81eb-6b0e-4b55-8309-e5afbdb50c28

Arukuen:
fixes #3164 

[notes](https://www.notion.so/Entering-invalid-Custom-CSS-can-break-block-styles-3164-7a027fbc7037417ea71988226691b78e?pvs=4)

### DIFF
--- a/src/css-optimize.php
+++ b/src/css-optimize.php
@@ -462,7 +462,17 @@ if ( ! class_exists( 'Stackable_CSS_Optimize' ) ) {
 					// Optimize selectors by combining similar ones.
 					$selector_arr = self::combine_selectors( $selector_arr );
 
-					$css .= implode( ',', $selector_arr ) . $style_rules;
+					$combined_css = implode( ',', $selector_arr ) . $style_rules;
+
+					// Prevent invalid CSS in affecting the rest of the CSS
+					// by closing any open curly braces.
+					$open_curly_brace = substr_count( $combined_css , '{' );
+					$close_curly_brace = substr_count( $combined_css , '}' );
+					if ( $open_curly_brace > $close_curly_brace ) {
+						$combined_css .= str_repeat( '}', $open_curly_brace - $close_curly_brace );
+					}
+
+					$css .= $combined_css;
 				}
 
 				if ( ! empty( $media_query ) ) {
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
I believe the `return ''` here stops the css from working (because it stops prepending the necessary CSS selector). I think the better way to fix this is to still continue prepending the necessary CSS selector even if a nested selector is used.

CSS selector nesting is now allowed in the CSS standard, so we should allow it also.

bfintal:
@andeng1106 we need to check if media queries (and shortcut media queries `@media mobile` still work

andeng1106:
Issue is now fixed. I just have one query:

> @andeng1106 we need to check if media queries (and shortcut media queries `@media mobile` still work

@bfintal media queries are still working but shortcut media queries (`@media mobile)` is not working anymore on frontend since 3.13.1 and older versions

https://github.com/gambitph/Stackable/assets/103395655/6a81a4d9-cd1d-447f-b2dd-67c5b8cbf854

PR #3201, opened 2024-06-05
feat: shared inspector controls

### LINKED ISSUE #2711: If multiple blocks are selected, the inspector should still show the shared controls
Currently, the inspector controls are hidden when multiple **same** blocks are selected. This is mostly because we conditionally display (for performance) inspector controls using `props.isSelected`

But if you remove this `isSelected` condition, the inspector shows the controls and changing options change the attribute correctly.

There is an issue though, multiple instances of the controls/panels are show up in the inspector.

**Findings as to why this happens:**

I think this might have something to do with our Slot/Fill solution. Because if you simply do the `InspectorControls` inside the block's edit function directly, the panel/control doesn't show as double. And if you check our `InspectorTabs` component, it doesn't double either.

kaeizen:
fixes #2711


src/block-components/columns/edit.js
```diff
@@ -53,14 +84,14 @@ export const Controls = props => {
...
+		columnWidthsMobile.push( attributes?.columnWidthMobile )
 
-		if ( attributes.columnWidthTablet ) {
+		if ( attributes?.columnWidthTablet ) {
 			hasTabletColumnWidths = true
 		}
-		if ( attributes.columnWidthMobile ) {
+		if ( attributes?.columnWidthMobile ) {
 			hasMobileColumnWidths = true
 		}
```
bfintal: If the attributes suddenly are not available, this might stop working

bfintal (changes requested):

andeng1106:
@mxkae Blocker: Conflict files. Will return this to Change Requested.

PR #3204, opened 2024-06-10
fix (inner columns): fix flex direction in Firefox polyfill

### LINKED ISSUE #3193: Firefox issue with Inner Columns Margin
**Describe the bug**
- When top and bottom margins from inner column is set to `auto`, it looks different on Firefox browser.
- Issue only on Firefox

**To Reproduce**
1. Add Columns block (2 or more inner columns)
2. Add Heading block to the first inner column, then Card block to the second inner column
3. Make the first inner column have top and bottom margins set to `auto`
4. Compare Editor in both Chrome and Firefox

**Expected behavior**
- Columns should still be in one line:
Screenshot 2024-05-23 at 2 47 48 PM

**Screenshots**
1st part - Chrome (expected)
2nd part - Firefox (has issue)

https://github.com/gambitph/Stackable/assets/103395655/a120a10b-bcfd-4e93-a4ee-aa8c85c7a964

 - OS: [e.g. MacOS, Windows, iOS, Android, Palm OS]
 - Browser: [e.g. Chrome, Firefox, Edge, Safari, IE 11, IE 7]
 - Browser Version: [e.g. 22]

**Smartphone (please complete the following information):**
 - Device: [e.g. iPhoneX]
 - OS: [e.g. iOS 12.0]
 - Browser: [e.g. stock browser, Safari, Chrome]
 - Browser Version: [e.g. 22]

kaeizen:
fixes #3193 

The css rule polyfill has higher specificity so `flex-direction: row` when doesn't get applied even if the block has `.stk-row` class.

Chrome:
chrome

Firefox:
firefox

### DIFF
--- a/src/block-components/alignment/editor.scss
+++ b/src/block-components/alignment/editor.scss
@@ -79,8 +79,3 @@
 	}
 }
 
-.stk-row.stk--alignment-polyfill > :is(.stk-inner-blocks, .stk-block-content):not(.stk--column-flex) {
-	&:not(.stk--block-horizontal-flex) > .block-editor-inner-blocks > .block-editor-block-list__layout {
-		flex-direction: row;
-	}
-}
--- a/src/block-components/alignment/editor.scss
+++ b/src/block-components/alignment/editor.scss
@@ -78,3 +78,9 @@
 		}
 	}
 }
+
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
It feels a incorrect that our fix is to create a really long and high specificity selector.

Not sure the implications currently, but can we add in the an `!important` to the `flex-direction`?

PR #3205, opened 2024-06-10
feat(image): add aspect ratio option

### LINKED ISSUE #3200: Add aspect ratio option to Image Block
Add an **Aspect Ratio** option between the **Image Url** and **Width** option in the image block.

Add the following aspect ratios:

#### Square
- Square - 1:1
#### Landscapes
- Standard - 4:3
- Classic - 3:2
- Wide - 16:9
- Cinematic - 2:1
- Ultra Wide - 3:1
- Panoramic - 4:1
#### Portraits
- Portrait - 3:4
- Classic Portrait - 2:3
- Tall - 9:16

The css rule `aspect-ratio` should be added to the image block's `.stk-img-wrapper` class

## Expected Behavior

- When you change the aspect ratio, the image dimensions should change to reflect it
- Changing either the height or the width attribute of the image should change the size of the image, but the aspect ratio should still be followed
- Changing both height & width would override the aspect ratio (but no need to do anything about it: e.g. prevent the css rule from being applied etc)

Arukuen:
fixes #3200

### DIFF
--- a/src/block-components/image/edit.js
+++ b/src/block-components/image/edit.js
@@ -49,10 +49,25 @@ const IMAGE_SHADOWS = [
 	'70px 60px 60px rgba(72, 73, 121, 0.2) ',
 ]
 
+const ASPECT_RATIOS = [
+	{ label: __( 'Default', i18n ), value: '' },
+	{ label: __( 'Square 1:1', i18n ), value: '1/1' },
+	{ label: __( 'Standard 4:3', i18n ), value: '4/3' },
+	{ label: __( 'Classic 3:2', i18n ), value: '3/2' },
+	{ label: __( 'Wide 16:9', i18n ), value: '16/9' },
+	{ label: __( 'Cinematic 2:1', i18n ), value: '2/1' },
+	{ label: __( 'Ultra Wide 3:1', i18n ), value: '3/1' },
+	{ label: __( 'Panoramic 4:1', i18n ), value: '4/1' },
+	{ label: __( 'Portrait 3:4', i18n ), value: '3/4' },
+	{ label: __( 'Classic Portrait 2:3', i18n ), value: '2/3' },
+	{ label: __( 'Tall 9:16', i18n ), value: '9/16' },
+]
+
 const Controls = props => {
 	const attributes = useBlockAttributesContext( attributes => {
 		return {
 			imageId: attributes.imageId,
+			imageAspectRatio: attributes.imageAspectRatio,
 			imageWidthUnit: attributes.imageWidthUnit,
 			imageHeightUnit: attributes.imageHeightUnit,
 			imageWidth: attributes.imageWidth,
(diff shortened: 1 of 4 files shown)


src/block-components/image/edit.js
```diff
@@ -150,6 +151,28 @@ const Controls = props => {
...
+					{ label: __( 'Ultra Wide 3:1', i18n ), value: '3/1' },
+					{ label: __( 'Panoramic 4:1', i18n ), value: '4/1' },
+					{ label: __( 'Portrait 3:4', i18n ), value: '3/4' },
+					{ label: __( 'Classic Portrait 2:3', i18n ), value: '2/3' },
+					{ label: __( 'Tall 9:16', i18n ), value: '9/16' },
+				] }
+				helpTooltip={ {
+					title: __( 'Aspect Ratio', i18n ),
+					description: __( 'Sets the aspect ratio of the image', i18n ),
+				} }
```
bfintal: No need for this since it's quite self explanatory. We want as least amount of text as possible because every text would need to be translated to other languages.
```suggestion
```

src/block-components/image/edit.js
```diff
@@ -185,7 +204,15 @@ const Controls = props => {
...
 						title: __( 'Image height', i18n ),
 						description: __( 'Adjusts the image height', i18n ),
-					} }s
+					} }
+					// Set the width to auto if the aspect ratio is set and the width is currently empty
+					// to allow the aspect ratio to take effect with height changes.
+					onChange={ value => {
+						setAttributes( { imageHeight: value } )
+						if ( attributes.imageAspectRatio !== '' && attributes.imageWidth === '' ) {
+							setAttributes( { imageWidth: 'auto' } )
```
bfintal: We want to call `setAttributes` only **once** when a value changes. In the block editor, whenever `setAttributes` is called, it creates a new entry in the **changes history**, since it's called twice here, you'll have to undo two times.

bfintal (changes requested):

src/block-components/image/style.js
```diff
@@ -35,9 +35,16 @@ const Styles = props => {
 		heightUnitCallback = null,
 		dependencies = [],
 	} = props
-
 	return (
 		<>
+			<BlockCss
+				{ ...propsToPass }
+				selector={ selector }
+				styleRule="aspect-ratio"
```
bfintal: @Arukuen this should be in camel case
```suggestion
				styleRule="aspectRatio"
```

andeng1106:
@Arukuen Aspect Ratio setting is present for Card block but it doesn't follow any ratio, even when changing height.

https://github.com/gambitph/Stackable/assets/103395655/60181d15-784d-4075-8068-a7655a1e8e79

cc: @bfintal

bfintal:
For the card block, I would expect the Aspect Ratio to be present and should have an effect. So e.g. setting it to square should really make the image square

Arukuen:
Fixed the aspect ratio for card block, specifically `default `and `default-2` variations. 

For variation `horizontal`, I think the aspect ratio option should be removed since the height is already dependent to the content of the card. For variation `full` and `faded`, I think it should also be removed since the dimensions of the image are dependent to the card itself like a background.

src/block-components/image/style.js
```diff
@@ -35,9 +35,29 @@ const Styles = props => {
 		heightUnitCallback = null,
 		dependencies = [],
 	} = props
-
 	return (
 		<>
+			<BlockCss
+				{ ...propsToPass }
+				selector={ [
+					`${ selector }`,
```
bfintal:
```suggestion
					selector,
```

PR #3209, opened 2024-06-13
fix (icon label): change icon gap

### LINKED ISSUE #3126: Icon Label block Icon Gap issue
**Describe the bug**
When I set the Icon Gap to `36` or below, it stops immediately and no gap anymore instead of having no gap at `0`.

**To Reproduce**
1. Add a new page
2. Add an Icon Label block
3. Move slide to set Icon Gap to `36` or below
4. Set Icon Gap to `0`
5. Compare the difference between step 3 and 4

**Expected behavior**
Icon Gap no gap should appear when it's set to `0`

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/314dbd82-62ae-440e-bf45-3f3510f42593

kaeizen:
fixes #3126

### DIFF
--- a/src/block/icon-label/editor.scss
+++ b/src/block/icon-label/editor.scss
@@ -1,9 +1,9 @@
 .stk-block-icon-label [data-type="stackable/icon"] {
-	flex: 0 0 0;
+	flex: 0;
 }
 // Icon label can collapse in mobile in the editor.
 .stk-preview-device-mobile .stk-block-icon-label .block-editor-block-list__layout [data-type="stackable/icon"] {
-	flex: 0 0 0;
+	flex: 0;
 }
 .stk-preview-device-mobile .stk-block-icon-label .block-editor-block-list__layout [data-type="stackable/heading"] {
 	flex: 1 1 0;
--- a/src/block/icon-label/editor.scss
+++ b/src/block/icon-label/editor.scss
@@ -1,9 +1,9 @@
 .stk-block-icon-label [data-type="stackable/icon"] {
-	flex: 0 0 64px;
+	flex: 0 0 0;
 }
 // Icon label can collapse in mobile in the editor.
 .stk-preview-device-mobile .stk-block-icon-label .block-editor-block-list__layout [data-type="stackable/icon"] {
-	flex: 0 0 64px;
+	flex: 0 0 0;
 }
 .stk-preview-device-mobile .stk-block-icon-label .block-editor-block-list__layout [data-type="stackable/heading"] {
 	flex: 1 1 0;
@@ -13,4 +13,8 @@
 	.stk-inner-blocks.stk-block-content {
 		max-width: unset !important;
 	}
+}
+
+.stk-block-icon-label .stk-inner-blocks .block-editor-block-list__layout {
+	gap: 28px;
 }
--- a/src/block/icon-label/index.php
+++ b/src/block/icon-label/index.php
@@ -1,0 +1,24 @@
+<?php
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( ! function_exists( 'stackable_render_block_icon_label' ) ) {
+	/**
+	 * Adds a special class if block uses the old icon gap attribute
+	 */
+	function stackable_render_block_icon_label( $block_content, $block ) {
+		$attributes = $block[ 'attrs' ];
+
+		$hasOldIconGap = isset( $attributes[ 'iconGap' ] ) && ! isset( $attributes[ 'iconGap2' ] );
+
+		if ( $hasOldIconGap ) {
+			return preg_replace( '/stk-block-icon-label/i', 'stk-block-icon-label stk-block-icon-label--use-flex-basis', $block_content );
+		}
+
+		return $block_content;
+	}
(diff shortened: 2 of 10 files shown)


bfintal (changes requested):
While this should work, I want to suggest whether we can try a different (cleaner) approach.

Approach:
- We create a new attribute `iconGap2` that holds the new value for `iconGap`, moving forward `iconGap` will no longer be used, but it will still be kept in the attribute definition
- In PHP, in `render_block` if we encounter the block that has an `iconGap` attribute, but doesn't have an `iconGap2`, we:
  - Calculate for the new icon gap value, then replace the old CSS with the updated CSS
  - Add a special class to say that this block should use the old css `flex: 0 0 64px`
- In the editor JS, we have this deprecation entry: `isEligible` returns true if the block contains a value for `iconGap`, the migrate the value from `iconGap` to `iconGap2`

Pros of approach:
- No need to create a new `version` attribute
- No need for a new class name

src/block/icon-label/index.php
```diff
@@ -0,0 +1,25 @@
...
+
+if ( ! function_exists( 'stackable_render_block_icon_label' ) ) {
+	/**
+	 * Adds a special class if block uses the old icon gap attribute
+	 */
+	function stackable_render_block_icon_label( $block_content, $block ) {
+		$attributes = $block[ 'attrs' ];
+
+		// check if the block uses the old icon gap attribute
+		$use_old_styles = ( isset( $attributes[ 'iconGap' ] ) || isset( $block[ 'innerBlocks' ][0]['attrs']['iconSize'] ) ) && ! isset( $attributes[ 'iconGap2' ] );
```
bfintal: Check first for the presence of the elements in `$block['innerBlocks']` to prevent fatal errors.

src/block/icon-label/index.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+// Exit if accessed directly.
```
bfintal: can we rename this file to deprecated.php or something so we can denote that it's for deprecation purposes?

src/block/icon-label/schema.js
```diff
@@ -37,6 +37,11 @@ export const attributes = ( version = VERSION ) => {
 				stkResponsive: true,
 				default: '',
 			},
+			iconGap2: {
```
bfintal, line 42: Add a comment on `iconGap` that it's deprecated but kept here for migration purposes

src/block/icon-label/style.js
```diff
@@ -26,6 +26,7 @@ const Styles = props => {
 
 	return (
 		<>
+			{ /* For the old icon gap */ }
```
bfintal: I believe you should be able to remove the `BlockCss` entry for the old `iconGap` here and the block should not error out.

src/block/icon-label/style.scss
```diff
@@ -1,8 +1,20 @@
 .stk-block-icon-label {
-	// Default width of the icon width.
-	.stk-block-icon {
-		flex: 0 0 64px;
+	// For old icon gap
```
bfintal: can you move this css to the bottom (isolating it from the rest of the "current" css code) and then label it as "deprecated"

bfintal (changes requested):
Just some code organization suggestions

src/block/icon-label/style.scss
```diff
@@ -25,3 +26,14 @@
 		width: inherit;
 	}
 }
+
+// Deprecated styles
```
bfintal:
```suggestion
// Deprecated styles: we changed how icon gap works. Before it was flex basis, now it's a gap
```

src/block/icon-label/schema.js
```diff
@@ -32,11 +32,17 @@ export const attributes = ( version = VERSION ) => {
 	ConditionalDisplay.addAttributes( attrObject )
 	attrObject.add( {
 		attributes: {
+			// iconGap is deprecated but is kept for migration purposes
 			iconGap: {
 				type: 'number',
 				stkResponsive: true,
 				default: '',
 			},
+			iconGap2: {
```
bfintal, line 42:
```suggestion
			// This replaces iconGap above, the old attribute changes the flex basis, now it's gap
			iconGap2: {
```

src/block/icon-label/editor.scss
```diff
@@ -1,9 +1,9 @@
 .stk-block-icon-label [data-type="stackable/icon"] {
-	flex: 0 0 64px;
+	flex: 0 0 0;
 }
 // Icon label can collapse in mobile in the editor.
 .stk-preview-device-mobile .stk-block-icon-label .block-editor-block-list__layout [data-type="stackable/icon"] {
-	flex: 0 0 64px;
+	flex: 0 0 0;
```
bfintal: Does it work if we remove the flex definition completely since this is for the old `0 0 64px`?

kaeizen: it still needs the flex, but i changed it to `flex: 0` instead.

andeng1106:
@mxkae Issue with Icon Gap has now been fixed. 

One remaining issue:
- Upon updating to the PR 3209 build, Icon Gap settings applied on tablet and mobile is missing/reset 

https://github.com/gambitph/Stackable/assets/103395655/5a473dad-eb6b-4fc9-84fb-460190aa429d

PR #3222, opened 2024-06-27
fix: add admin notice if occurs error in settings

### LINKED ISSUE #2882: Plugin Settings: Display an error message if REST API fails
If the REST API doesn't work, or saving/loading settings in the plugin settings area doesn't work, we should display an **admin notice** with the message we got from the REST fetch request (if there is one) and a message to check the settings of any security plugin or security setting in their website that disables REST API for administrators.

Steps to replicate:
1. install https://wordpress.org/plugins/all-in-one-wp-security-and-firewall/ and enable Intranet option
2. Go to Stackable settings
3. See that all settings are not loading, you also cannot change settings

Arukuen:
fixes #2882
Screenshot 2024-06-27 223818

### DIFF
--- a/src/welcome/admin.js
+++ b/src/welcome/admin.js
@@ -239,7 +239,9 @@ const SettingsNotice = () => {
 	return (
 		<div className="notice notice-error">
 			<p>{ __( 'Error getting Stackable settings. We got the following error. Please contact your administrator.', i18n ) }</p>
-			{ error.responseJSON.message && <p><strong>{ error.responseJSON.message }</strong></p> }
+			{ error.responseJSON &&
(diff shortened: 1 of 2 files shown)


src/welcome/admin.js
```diff
@@ -219,6 +219,31 @@ const pick = ( obj, keys ) => {
...
+	}, [] )
+
+	if ( ! error ) {
+		return null
+	}
+
+	return (
+		<div className="notice notice-error">
+			<p>{ __( 'Error getting Stackable settings. We got the following error. Please contact your administrator.', i18n ) }</p>
+			{ error.responseJSON.message && <p><strong>{ error.responseJSON.message }</strong></p> }
```
bfintal: Please also add the error code. Afterwards please set for QA

bfintal (changes requested):

PR #3227, opened 2024-07-01
fix: display theme fonts in font picker

### LINKED ISSUE #3195: Fonts from the Block Theme & Uploaded from Site Editor should show up in the Font Picker
Expectations: 
- When using a Block Theme, the theme included fonts should be selectable in our font pickers
- When uploading a font in the Site Editor, the font should be selectable in our font picker
- You should be able to use a font or font-stack that's not included in our list (e.g. "Arial" or "Futura, "Trebuchet MS", Arial, sans-serif")

kaeizen:
fixes #3195


src/components/font-family-control/index.js
```diff
@@ -40,11 +61,11 @@ const FontFamilyControl = props => {
 				options: fontOptions,
 			},
 		] )
-	}, [] )
+	}, [ themeFonts ] )
 
 	return (
 		<AdvancedAutosuggestControl
-			options={ options }
+			options={ themeFonts.length ? options : options.filter( option => option.id !== 'theme-fonts' ) }
```
bfintal: Do not filter it out here, instead, the theme fonts should not be added from the start if there aren't any

bfintal (changes requested):
The theme font should not be loaded the same way Stackable loads our Google fonts in the backend & frontend (they are not necessarily Google Fonts). Since they are theme fonts, then they should be loaded from the theme

src/fonts.php
```diff
@@ -22,6 +22,72 @@ function str_ends_with( $haystack, $needle ) {
...
+			if ( is_frontend() ) {
+				add_action( 'init', array( $this, 'gather_theme_fonts' ) );
+			}
+
+			add_action( 'rest_api_init', array( $this, 'register_rest_fields' ) );
+		}
+
+		public function gather_theme_fonts() {
+			if ( ! method_exists( 'WP_Font_Face_Resolver', 'get_fonts_from_theme_json' ) ) {
+				return false;
```
bfintal: No need to return anything in an action.
```suggestion
				return;
```

src/fonts.php
```diff
@@ -22,6 +22,72 @@ function str_ends_with( $haystack, $needle ) {
...
 }
 
+if ( ! class_exists( 'Stackable_Theme_Fonts' ) ) {
+	class Stackable_Theme_Fonts {
+
+		public static $theme_fonts = [];
+
+		function __construct() {
+			if ( is_frontend() ) {
+				add_action( 'init', array( $this, 'gather_theme_fonts' ) );
```
bfintal: We can optimize this area. Checking the code, looks like this `init` action is only needed to load all the fonts once so we can have a local list of what all the fonts are.. then this list is only needed by the `is_theme_font` function.

I recommend deleting this `add_action` since it's just an initializer, and instead call `gather_theme_fonts` directly inside the function `is_theme_font` when it's needed. We can also only do the gather once.

Suggestion:
```js
// Change the default value to `false` so we know when it's not yet initialized
public static $theme_fonts = false;

public static function gather_theme_fonts() {
	if ( self::$theme_fonts !== false ) {
		return self::$theme_fonts;
	}
	self::$theme_fonts = [];
	// Proceed with the original function...
	return self::$theme_fonts;
}

public static function is_theme_font( $font_name ) {
	return in_array( strtolower( $font_name ), self::gather_theme_fonts() );
}
```

src/fonts.php
```diff
@@ -22,6 +22,72 @@ function str_ends_with( $haystack, $needle ) {
...
+			if ( ! method_exists( 'WP_Font_Face_Resolver', 'get_fonts_from_theme_json' ) ) {
+				return false;
+			}
+
+			$theme_fonts = WP_Font_Face_Resolver::get_fonts_from_theme_json();
+
+			$response = array();
+			foreach ( $theme_fonts as $key => $value ) {
+				$response[] = $value[ 0 ][ 'font-family' ];
+			}
```
bfintal: Looks like this is doing the same thing as `gather_theme_fonts` except returning it as a REST response. Maybe we can combine for DRY code

bfintal (changes requested):

PR #3236, opened 2024-07-09
fix: extract custom handler for aspect ratio to allow extra operation…

### LINKED ISSUE #3232: Image block > Aspect Ratio & Height - Tablet & Mobile responsive setting not working
**Describe the bug**
- When I change tablet/mobile setting for the ff. below, it applies even to desktop:
   - Aspect Ratio
   - Height
- Even the orange indicator is not working
- Applies to blocks such as:
   - Image block
   - Card block (any layout)
   - Posts block (any layout)

Notes/Investigation:
- Issue started to happen after the additional fixes was added in this [pr](https://github.com/gambitph/Stackable/pull/3205)
Screenshot 2024-07-05 at 7 28 22 AM

**To Reproduce**
1. Add a new page
2. Add any block mentioned above
3. Change aspect ratio setting in tablet/mobile
4. See that it also applies to desktop/all responsive views

**Expected behavior**
- Aspect Radio responsive setting should work properly

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/200c2643-9348-4b1a-82ca-0341d84bc804

Arukuen:
… while being responsive

fixes #3232

### DIFF
--- a/src/block-components/image/edit.js
+++ b/src/block-components/image/edit.js
@@ -25,8 +25,9 @@ import {
 import {
 	useBlockAttributesContext,
 	useBlockSetAttributesContext,
+	useDeviceType,
 } from '~stackable/hooks'
-import { useControlHandlers } from '~stackable/components/base-control2/hooks'
+import { getAttributeName } from '~stackable/util'
 
 /**
  * WordPress dependencies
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
While this works, using `useControlHandlers` like this here is a bit hard to comprehend. Since the issue is that it's always updating the desktop value, you can use the function `getAttributeName` to grab the desktop/tablet/mobile attribute name and update that:

```js
import { getAttributeName } from '~stackable/util'

// In the component:
const deviceType = useDeviceType()

// In onChange:
const attrImageAspectRatio = getAttributeName( 'imageAspectRatio', deviceType )
const values = { [ attrImageAspectRatio ]: value }
```

On another note, is reseting the desktop value for `imageHeight` to `''` the correct way to go? Or do we also have to reset the tablet and mobile values for this attribute?

PR #3237, opened 2024-07-10
fix (Feature Grid): fix full width on column wrap desktop with column justify 

### LINKED ISSUE #3225: [WP 6.6] Feature Grid block - If there's a Column Justify setting present, the Full Width loses its setting
**Describe the bug**
- Issue both in editor and frontend
- If block width alignment is set to Full Width and there's a Column Justify setting (any), it loses the full width alignment

**To Reproduce**
1. Add a new page
2. Add a Feature Grid block
3. Set block width alignment to **Full Width**
4. Enable "Allow Column Wrapping"
5. Set all Inner Column Width to 100%
6. Set any Column Justify setting
7. See that the full width doesn't apply anymore

**Expected behavior**
If there's a Column Justify setting present and block width alignment is set to Full Width, the block should look full width

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/5a030a35-e8ec-4b04-8b24-95765eb6d084

kaeizen:
fixes #3225

### DIFF
--- a/src/block/feature-grid/deprecated.js
+++ b/src/block/feature-grid/deprecated.js
@@ -20,6 +20,28 @@ import {
  */
 import { addFilter } from '@wordpress/hooks'
 
+// Version 3.13.3 Deprecations
+addFilter( 'stackable.feature-grid.save.blockClassNames', 'stackable/3.13.3', ( output, props ) => {
+	if ( compareVersions( props.version, '3.13.3' ) === 1 ) {
+		return output
+	}
+
+	const separatorClass = getSeparatorClasses( props.attributes )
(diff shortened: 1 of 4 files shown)


bfintal:
@mxkae does the original issue only affect the Feature Grid?

kaeizen:
> @mxkae does the original issue only affect the Feature Grid?

in columns, having:
- enabled desktop column wrap
- all column widths = 100%
- block width `align full`
- justify content `space between`

it looks like this:
Screenshot 2024-07-11 at 6 19 29 AM

is this expected or should the columns still have a 100% width?

Issue #354, opened 2019-11-19
 plain and outlined notification's button color do not match the notification type


mico97:
![notif](https://user-images.githubusercontent.com/50035760/69118416-5f803580-0ace-11ea-8f9a-7bb14e2b7bc0.png)

PR #3243, opened 2024-07-13
fix: corrected variable name that causes empty media queries

### LINKED ISSUE #3214: Custom CSS - media queries not working properly
**Describe the bug**
- When using Shortcut Media Queries for Custom CSS, it's not working properly especially on frontend
- Two issues:
    - 1. In frontend, Media queries added are not showing (whether shortcut media query or not)
    - 2. In editor, shortcut media query `@media mobile` is applying on both tablet and mobile instead of mobile only
          - Shortcut Media Queries such as:
              - `@media tablet`
              - `@media mobile`
              - etc..

Notes:
- Issue on both editor and frontend
- Issue happens on 3.13.1 and older versions

**To Reproduce**
_**Using Shortcut Media Queries**_
1. Add a new page
2. Add a Heading block
3. Add these codes on Custom CSS (replace the **.stk-xxxx** to the assigned id `data-block-id`):
```
/* Media Query will apply to Tablet */
@media tablet {
		  .stk-xxxx .stk-block-heading__text {
		font-size: 50px !important;
    background-color: darkmagenta !important;
  }
}

/* Media Query will apply to Mobile */
@media mobile {
	  .stk-xxxx .stk-block-heading__text {
		font-size: 10px !important;
    background-color: lightgoldenrodyellow !important;
  }
}

```

4. In editor, change responsive mode to **Mobile**
5. See that the `@media mobile` queries are applied correctly
6. Change responsive mode to **Tablet**
7. See if the `@media mobile` queries are applied instead of `@media tablet`
8. Save page and view frontend
9. Go to Inspect Element
10. Use Toggle device toolbar then switch responsive to tablet or mobile
11. See that the shortcut media queries is not applied

====

**_Using Media Queries_**
1. Add a new page
2. Add a Heading block
3. Add these codes on Custom CSS (replace the **.stk-xxxx** to the assigned id `data-block-id`):
```
/* Media Query will apply to Tablet */
@media screen (max-width: 1024px) {
		  .stk-xxxx .stk-block-heading__text {
		font-size: 50px !important;
    background-color: darkmagenta !important;
  }
}

/* Media Query will apply to Mobile */
@media screen (max-width: 768px) {
	  .stk-xxxx .stk-block-heading__text {
		font-size: 23px !important;
    background-color: lightgoldenrodyellow !important;
			color: blue;
  }
}

```
4. In editor, check if css is working on different responsive modes (based on the css breakpoint provided)
6. See that the css media queries added are not showing

**Expected behavior**
- Media queries should work properly on both frontend and editor
- Shortcut media queries should only apply to its respective responsive mode

**Screenshots**
**_Using Shortcut Media Queries_**

https://github.com/gambitph/Stackable/assets/103395655/d9898b20-d216-45f0-bdd9-3fd910cff674

**_Using Media Queries_**

https://github.com/gambitph/Stackable/assets/103395655/756e6c9e-1ce0-4da6-a14e-2bdb8780e604

Arukuen:
fixes #3214


andeng1106:
@Arukuen Media Query is still not working on my end. Is there any steps I need to take to see the fix? 🤔


https://github.com/user-attachments/assets/940cf1a6-4b99-4e81-b2da-06d8c110d89a

Arukuen:
For the Media Query, we should use `@media screen and (max-width: 1024px)` instead of `@media screen (max-width: 1024px)`, since CSS syntax require the `and` keyword.

andeng1106:
> For the Media Query, we should use `@media screen and (max-width: 1024px)` instead of `@media screen (max-width: 1024px)`, since CSS syntax require the `and` keyword.

This is now working 😄

src/dynamic-breakpoints.php
```diff
@@ -32,6 +32,9 @@ function __construct() {
 			add_action( 'rest_api_init', array( $this, 'register_settings' ) );
 
 			if ( is_frontend() && $this->has_custom_breakpoints() ) {
+				// Add a filter for replacing shortcut media queries before the breakpoint adjustment.
+				add_filter( 'stackable_frontend_css', array( $this, 'replace_shortcut_media_queries' ) );
```
bfintal: Filters need to be assigned. 

e.g.
```php
$new_value = add_filter( 'filter_name', 'old_value' );
```

bfintal (changes requested):

src/dynamic-breakpoints.php
Arukuen: @bfintal Isn't this for `apply_filter` function? My goal here is to add a filter that replace the shortcut media queries before the breakpoint adjustment.

andeng1106:
@Arukuen Fix for shortcut media query is not yet working on my end. Is there something I need to do to make it work? 🤔 

https://github.com/user-attachments/assets/12ad5a26-a5c0-4e00-b314-332a14febbff

Arukuen:
@andeng1106 Have you tried also adding the fix from Premium?

andeng1106:
> @andeng1106 Have you tried also adding the fix from Premium?

@Arukuen Apologies for adding the comment here this PR. Yes, I'm using PR #354 in the video.

What I did so far:
- Edited existing block's custom css and saved page
- Added a new Heading block and used the `@media tablet` and `@media mobile`

Arukuen:
@andeng1106 The implementation for the editor comes from the premium PR, while the one for the frontend is from this PR. Based on the video, it seems like it is only working on the editor, and I think it's because the PR are tested separately.

If that's the case, here is a build that contains code for both PR.
[stackable.zip](https://github.com/user-attachments/files/16266621/stackable.zip)

andeng1106:
@Arukuen The build you provided shows the fix now. Thanks!

PR #3257, opened 2024-07-22
fix (carousel block): show inner column margins in frontend

### LINKED ISSUE #3173: Carousel block inner column margin issue
**Describe the bug**
- If I set margin on the Carousel block's Inner Column, it looks different on frontend compared to the editor
- Issue on frontend

**To Reproduce**
1. Add a new page (or skip steps and use [this page](https://gambitsandbox.instawp.xyz/andrea/carousel-block-code-test/) instead)
2. Add a Carousel block with at least 5 slides
3. Set Carousel block to "Full Width"
4. Set **Slides to Show** to 3
5. Add these settings to each Inner Column:
      - Background Color to each Inner Column (not related to the issue but useful) 
      - Margin (e.g. 15px on all sides) (setting that's causing the issue)
6. Add a Heading block on each Inner Column and add long text content (not related to the issue but useful) 
7. Save page and view frontend
8. Compare editor (correct) vs. frontend (issue)
9. See that the third slide overlaps on the page, which shouldn't

**Expected behavior**
- When adding a margin on the Carousel block's inner column, it should not overlap nor look different compared to the one in editor

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/6bf6c395-7cae-410b-850f-d27566e165f1

kaeizen:
fixes #3173

### DIFF
--- a/src/block-components/block-div/edit.js
+++ b/src/block-components/block-div/edit.js
@@ -57,6 +57,7 @@ export const Edit = props => {
 							visualGuide={ {
 								highlight: 'padding',
 							} }
+							{ ...props.sizeControlSpacingProps }
 						/>
 					</PanelAdvancedSettings>
 				) }
--- a/src/block/carousel/edit.js
+++ b/src/block/carousel/edit.js
@@ -252,6 +252,9 @@ const Edit = props => {
 		}
 	}, [ maxSlides, activeSlide ] )
 
+	const sizeControlSpacingProps = {
+		enableMargin: false,
+	}
 	return (
 		<>
 			<>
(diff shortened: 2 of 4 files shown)


bfintal (commented):
Looks like a huge amount of changes that even adds to PHP overhead, just because the left/right margins are not showing in the frontend. Why are they not showing up in the first place?

andeng1106:
@mxkae The fix didn't apply on my end. Is there any setting I need to apply to show the fix? 🤔 


https://github.com/user-attachments/assets/22f4375a-958e-4831-98f0-0954dc44bcb7

bfintal:
@andeng1106 our fix is just to remove the ability to change the margins when inside columns since you can achieve the same results with containers

PR #3264, opened 2024-07-24
fix(carousel): pause autoplay while inline navigation is in progress

### LINKED ISSUE #3098: HTML Anchor page jumping issue with Carousel Block
**Describe the bug**
- Related to: https://www.notion.so/Duro-dddc27d619f4479ebfc227ddd0f71bdf
- This issue happens usually if autoplay is enabled
- Sample test page: https://gambitsandbox.instawp.xyz/andrea/blocksy-carousel-block-html-anchor-issue/
As per investigation:
- Only able to replicate this on Windows. Issue on Mac is inconsistent.
     - This issue shows up when I tried 4k screen resolution, works fine on 2k screens and mobile.

**To Reproduce**
1. Add a new page
2. Add atleast 3 sections (Columns, any Design Library templates, etc.) and one Carousel block
3. Assign HTML anchors to the 3 sections
4. Put the Carousel on top of any section (make sure it's not the last block)
5. Create a Header navigation menu
6. Add the HTML anchors as link in the URL (e.g #contact-us)
7. Go the page created
8. Try clicking the links in the nav menu
9. See that the url didn't jump exactly to the section but to the Carousel

**Expected behavior**
Page should not jump if there's an assigned HTML anchor to a section and a Carousel block in the page

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/27c90c6e-54b8-4da3-adea-b786dfacd6df

https://github.com/user-attachments/assets/dc4a1b89-57de-4841-bec4-5fc380d53d57

 - OS: Windows
 - Browser: Chrome
 - Browser Version: no specific browser

Arukuen:
fixes #3098

### DIFF
--- a/src/block/carousel/frontend-carousel.js
+++ b/src/block/carousel/frontend-carousel.js
@@ -2,6 +2,7 @@
  * WordPress dependencies
  */
 import domReady from '@wordpress/dom-ready'
+import { throttle } from 'lodash'
 
 // Function to check if the user is on an iOS device
 function isiOS() {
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
I like the theory of the fix, but I think the drawback of this fix is it relies on the detection of a successful scroll to the targeted page navigation. Also, we're attaching event listeners to each anchor link to the page once - so this can be a lot, and also will not be able to handle dynamically added links.

I'd like to recommend another approach, since the goal is to detect a scrolling navigation click, then pausing the auto-scroll until the page scroll is finished, we can do exactly this:

```js
// Note: Untested code
let listenerTimeout = null

// After 800ms of non-scrolling, then the navigation scroll is done and let's result autoplay
const scrollListener = e => {
    clearTimeout( listenerTimeout )
    listenerTimeout = setTimeout( () => {
        this.unpauseAutoplay()
        document.removeEventListener( 'scroll', scrollListener )
    }, 800 ) // The max of scroll navigations
}

// Listen to any scroll navigation clicks
document.addEventListener('click', e => {
    if ( e.target.closest( '[href^="#"]' ) ) { // Need this to check if we're inside an anchor
        this.pauseAutoplay()
        // Start listening to scroll events
        document.addEventListener( 'scroll', scrollListener )
    }
} )
```
```

File: src/block/carousel/frontend-carousel.js [hunk 1]
```diff
@@ -2,6 +2,7 @@
  * WordPress dependencies
  */
 import domReady from '@wordpress/dom-ready'
+import { throttle } from 'lodash'
```
bfintal: We cannot use `throttle` here because we do not want to load `lodash` in the frontend, we'll need to just do this the `setTimeout` way

src/block/carousel/frontend-carousel.js
```diff
@@ -588,6 +590,37 @@ class _StackableCarousel {
...
+		let fallbackTimeout = null
+
+		const handleEndScroll = () => {
+			clearTimeout( fallbackTimeout )
+			this.unpauseAutoplay()
+			document.removeEventListener( 'scroll', scrollListener )
+		}
+
+		// After 100ms of non-scrolling, then the navigation scroll is done, restart the autoplay
+		const scrollListener = throttle( () => {
```
bfintal: I believe you do not need `throttle` here since the scroll handler will just be called rapidly, so you're just re-creating the timeout fast until all the scrolls are finished. You may just need to increase your timeout

src/block/carousel/frontend-carousel.js
```diff
@@ -588,6 +590,37 @@ class _StackableCarousel {
...
+		// Listen to any scroll navigation clicks, including dynamically added
+		document.addEventListener( 'click', e => {
+			// Check if we're inside an anchor link
+			// eslint-disable-next-line @wordpress/no-global-event-listener
+			if ( e.target.closest( '[href^="#"]' ) ) {
+				this.pauseAutoplay()
+				document.addEventListener( 'scroll', scrollListener, { passive: true } )
+				// Fallback to autoplay is resumed even if the user manually override
+				// the inline navigation or by clicking the link again
+				fallbackTimeout = setTimeout( handleEndScroll, 1000 )
```
bfintal: I think you can also just do the same `listenerTimeout` here 
```suggestion
				clearTimeout( listenerTimeout )
				listenerTimeout = setTimeout( handleEndScroll, 1000 )
```

bfintal (changes requested):

src/block/carousel/frontend-carousel.js
```diff
@@ -588,6 +589,34 @@ class _StackableCarousel {
...
+		document.addEventListener( 'click', e => {
+			// Check if we're inside an anchor link
+			// eslint-disable-next-line @wordpress/no-global-event-listener
+			if ( e.target.closest( '[href^="#"]' ) ) {
+				this.pauseAutoplay()
+				document.addEventListener( 'scroll', scrollListener, { passive: true } )
+				clearTimeout( listenerTimeout )
+				listenerTimeout = setTimeout( handleEndScroll, 1000 )
+			}
+		}, { passive: true } )
```
bfintal:
```suggestion
		// After 200ms of non-scrolling, then the navigation scroll is done, restart the autoplay
		const scrollListener = () => {
			clearTimeout( listenerTimeout )
			listenerTimeout = setTimeout( handleEndScroll, 200 )
		}

		// Listen to any scroll navigation clicks, including dynamically added
		document.addEventListener( 'click', e => {
			// Check if we're inside an anchor link
			// eslint-disable-next-line @wordpress/no-global-event-listener
			if ( e.target.closest( '[href^="#"]' ) ) {
				this.pauseAutoplay()
				document.addEventListener( 'scroll', scrollListener, { passive: true } )
				clearTimeout( listenerTimeout )
				listenerTimeout = setTimeout( handleEndScroll, 200 )
			}
		}, { passive: true } )
```

PR #3270, opened 2024-07-31
fix/3239/3268 image default width

### LINKED ISSUE #3239: Image block - Default Size issue
Yes. If a user adds a Stackable Image block and uploads an image with a smaller size, it appears bigger upon uploading. Compared to the WordPress Image block, it shows the exact image size upon uploading the image.

Here's a sample video, comparing WordPress Image block vs. Stackable Image block:

https://github.com/user-attachments/assets/df907a85-401f-43db-be50-c71a7341c319

- Add a new option to have another mode for images so that images by default will occupy by 100% so that images by default will occupy its current size
- New installs will have the new method for images

- User need to resize the Width and Height manually to get the exact image size.

### LINKED ISSUE #3268: Image block - When changing image size units for Width, it doesn't apply properly on both editor and frontend
**Describe the bug**
- Issue both on editor and frontend
- Issue applies to Width setting only
- Issue applies if I haven't added any Width setting yet
- Scenarios:
  1. Change to px
      - Editor: Image size is 100px (correct)
      - Frontend: It shows as 100% (wrong)
  2. Change to vw
      - Editor: Image Resizer shows 100vw (correct) but image size is 100px (wrong)
      - Frontend: It shows as 100% (wrong)
- Might also fix https://github.com/gambitph/Stackable/issues/3139

**To Reproduce**
1. Add a new page
2. Add an Image block and select an image
3. Change the unit of Width from `%` to `px`
4. See that the image has width of 100px in editor
5. Check if image in frontend is still 100%
6. Change the unit of Width to vw
7. See that the image resizer shows 100vw, but image size is still 100px
8. Check if Image In frontend is still 100%

**Expected behavior**
When changing Image Width unit to `px` or `vw`, it should show the expected size on both editor and frontend.

**Screenshots**

https://github.com/user-attachments/assets/15b72fea-1ce0-4f49-8705-e028613c5ec8

Arukuen:
fixes #3268
fixes #3239

Note that this PR can satisfy the feature of #3239 since the default image `width` in the units of `px` is chaged from a default of 300px to the original `width` of the image.


bfintal:
Changing the default width doesn't fix the issue since uploading a small image would still stretch them. Closing this in favor of #3403

PR #3274, opened 2024-08-05
fix: remove defs and comments unconditionally

### LINKED ISSUE #3265: Uploaded SVG icon from Icon block causes the Icon List icons to be invisible
**Describe the bug**
- Issue only happens with this [SVG icon](https://www.notion.so/Torsten-Wagner-f5946c334e754fbebc891c936d3fc026?pvs=4#c535189356e5432280d664ad6ef5870b)
- Issue both in editor and frontend
- More details here: https://www.notion.so/Torsten-Wagner-f5946c334e754fbebc891c936d3fc026
- Sample page: https://gambitsandbox.instawp.xyz/andrea/icon-list-svg-icon-issue/

**To Reproduce**
1. Add a new page
2. Add these blocks:
[icon list issue.txt](https://github.com/user-attachments/files/16357693/icon.list.issue.txt)

3. Add an Icon block
4. Use the SVG icon in the description above
5. See that the Icon List icons are not visible
7. Repeat Step 5

**Expected behavior**
When uploading an SVG icon to Icon block, it should not cause the icons from Icon List block to be invisible

**Screenshots**

https://github.com/user-attachments/assets/95add11d-ad28-4a1e-9fa0-6305dee623a6

Arukuen:
fixes #3265

### DIFF
--- a/src/components/icon-search-popover/index.js
+++ b/src/components/icon-search-popover/index.js
@@ -58,13 +58,11 @@ export const cleanSvgString = svgString => {
 	let newSvg = svgString.replace( /(^[\s\S]*?)(<svg)/gm, '$2' )
 		.replace( /(<\/svg>)([\s\S]*)/g, '$1' )
 
-	if ( newSvg.indexOf( '<!--' ) !== -1 ) {
-		// Remove defs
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
I recommend to do the replacements only when necessary (when `<defs` is present and if `<!--` is present)

PR #3275, opened 2024-08-05
fix: set the default block to stackable/text

### LINKED ISSUE #3273: Hitting while editing a Heading & Text block doesn't produce anymore a new Stackable Text block
Maybe a WP 6.6 issue. Hitting enter now places a native text block.

**Target behavior:**
When editing a Stackable Heading and Stackable Text block, pressing enter should add a Stackable text block after it

Arukuen:
fixes #3273 

Is there a better place where can we set the default block aside aside from editor-loader?


bfintal (changes requested):
Our previous behavior was that hitting enter would create a new Stackable text block.. but only when coming from a Stackable Text or Stackable Heading block. Otherwise, it would still just be a native text block.

This isn't an ideal fix because it will change all default blocks to be Stackable text blocks.

PR #3289, opened 2024-08-11
fix(posts): use uniqueId as stkQueryId to ensure uniqueness even when…

### LINKED ISSUE #3189: Posts block - Load More/Pagination issue when duplicating Posts block
**Describe the bug**
- Related to this [link](https://www.notion.so/Posts-block-Pagination-Load-More-button-not-working-properly-6cf6f41e33614921a0a2de97ac1791bd)
- Issue both happens if I **duplicated the Posts block**
- Two issues:
   - **Scenario 1**: Load More - Loading wrong data. I tried changing query to a custom type (e.g. Books) but it’s still showing Posts data
      - Sample page: https://gambitsandbox.instawp.xyz/andrea/posts-block-load-more-pagination-issue/
   - **Scenario 2**: Load More & Pagination - When changing pagination in the other Posts block in the same page with the Load More, the Posts block with the Load More is also affected
      - Sample page: https://gambitsandbox.instawp.xyz/andrea/posts-block-load-more-pagination-issue-2/

**To Reproduce**
_**Scenario 1**_
1. Add a Posts block
2. Duplicate the Posts block
3. In the original block, change Post Type to something else
4. Add Load More button on both Posts block
6. Try clicking the Load More of the original block
7. See that it gets the query from the duplicated block's Post Type

_**Scenario 2**_
1. Add a Posts block
2. Duplicate the Posts block
3. In the original block, change Post Type to something else
4. Add Load More button on the original block (optional)
5. Add a Pagination button for the duplicated block
7. Try clicking any page on the Pagination button
8. See that the original block data/query changes also 

**Expected behavior**
- Scenario 1:
   - The original block (the one that's duplicated) should not follow the duplicated posts block's query when clicking Load More
- Scenario 2:
   - The original block (the one that's duplicated) should not be affected when changing page on the duplicated posts block

**Screenshots**
Scenario 1:

https://github.com/gambitph/Stackable/assets/103395655/e71fc1ce-a6c5-4a0e-a8b3-673e96931b12

Scenario 2:

https://github.com/gambitph/Stackable/assets/103395655/a97ab416-a6f5-40cd-9653-bd87cfd01082

Arukuen:
… duplicated

fixes #3189

The problem arises from the `stkQueryId` attribute being also duplicated when the post block is duplicated. Only unique `stkQueryId` is assigned when new post blocks are created using `useInstanceId`.

Question: Is there a reason why in the previous implementation, `uniqueId`, which is always unique, was not used? If so, I can change the implementation to check all post blocks first before assigning `stkQueryId` to ensure uniqueness.

### DIFF
--- a/src/block/posts/edit.js
+++ b/src/block/posts/edit.js
@@ -68,9 +68,9 @@ import { Placeholder, Spinner } from '@wordpress/components'
 import { __ } from '@wordpress/i18n'
 import { applyFilters, addFilter } from '@wordpress/hooks'
 import { InnerBlocks, useBlockEditContext } from '@wordpress/block-editor'
-import { useMemo } from '@wordpress/element'
+import { useMemo, useEffect } from '@wordpress/element'
 import { useSelect } from '@wordpress/data'
-import { compose } from '@wordpress/compose'
+import { compose, useInstanceId } from '@wordpress/compose'
 
 const ALLOWED_INNER_BLOCKS = [
 	'stackable/load-more',
@@ -98,6 +98,7 @@ const Edit = props => {
 	useGeneratedCss( props.attributes )
 
 	const {
+		stkQueryId,
 		imageSize,
 		type = 'post',
 		orderBy = 'date',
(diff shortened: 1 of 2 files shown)


bfintal:
> Question: Is there a reason why in the previous implementation, `uniqueId`, which is always unique, was not used? If so, I can change the implementation to check all post blocks first before assigning `stkQueryId` to ensure uniqueness.

I believe that we were copying how it was done in the native posts block, where each post block has their own unique "instanceId" that is a marker so that query and pagination can be found when changing pages

PR #3245, opened 2024-07-15
fix: fixes freezing when creating synced patterns

kaeizen:
fixes #3229 #3226

PR #3292, opened 2024-08-14
fix (timeline block): set attribute timelineIsLast

kaeizen:
Issue:
- the line in last timeline block does not end at the middle.
- only happens on the frontend
- caused by #3245 

Screenshot:
Screenshot 2024-08-14 at 11 06 03 PM

Expected:
Screenshot 2024-08-14 at 11 08 45 PM


andeng1106:
@mxkae @bfintal 
When I create a new synced pattern or added an existing synced pattern, the page freezes. Tested on 3.13.5 also but no issue for that version.

**Created synced pattern**

https://github.com/user-attachments/assets/cc146c3b-26e2-48f9-8966-87143ea9345e

**Added existing synced pattern**

https://github.com/user-attachments/assets/a84198ab-6ea9-44de-8646-252e3e878bdd

PR #3293, opened 2024-08-14
fix (accordion block): fix timeline block lines if inside accordion block

### LINKED ISSUE #3152: Timeline block issue - Timeline lines glitch inside Accordion block issue
**Describe the bug**
- Timeline lines is glitchy if Timeline block is inside the Accordion block.
   - Seems that the scrolling effects isn't working either
- Issue on frontend only

**To Reproduce**
1. Add a new page
2. Add an Accordion block
3. In the Accordion's second inner column, add at least two or more Timeline blocks
4. Save page and view frontend
5. Expand the Accordion block
6. See that the Timeline lines are glitchy

**Expected behavior**
Timeline lines should not be glitchy on frontend if the Timeline blocks are inside the Accordion block

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/c3db88e1-b368-4959-97ab-077dbef0ad40

kaeizen:
fixes #3152

### DIFF
--- a/src/block/accordion/frontend-accordion.js
+++ b/src/block/accordion/frontend-accordion.js
@@ -157,9 +157,6 @@ class StackableAccordion {
 					elsAnchors[ el.getAttribute( 'id' ) ] = el
 				}
 
-				if ( el.contentEl.querySelector( '.stk-block-timeline' ) ) {
-					el.contentEl.classList.add( 'has-stk-block-timeline' )
-				}
 				el._StackableHasInitAccordion = true
 			}
 		} )
--- a/src/block/accordion/frontend-accordion.js
+++ b/src/block/accordion/frontend-accordion.js
@@ -156,6 +156,10 @@ class StackableAccordion {
 				if ( el.getAttribute( 'id' ) ) {
 					elsAnchors[ el.getAttribute( 'id' ) ] = el
 				}
+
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
I recommend not removing this, instead let's keep this but make it so that the timeline block will `revert` this or not inherit this value

src/block/accordion/style.scss
```diff
@@ -86,6 +86,13 @@
 		}
 	}
 
+	// the transform property causes the background-attachment: fixed to not work as expected.
+	// remove it to prevent timeline block lines to be glitchy.
+	// see similar issue: https://www.sitepoint.com/community/t/background-attachment-fixed-not-working-as-expected/35374/4
+	.stk-block-accordion__content.has-stk-block-timeline {
+		transform: initial;
+	}
```
bfintal: I believe we can start using `:has()` since it has been now implemented in Firefox!

bfintal (changes requested):

src/block/accordion/frontend-accordion.js
```diff
@@ -156,6 +156,7 @@ class StackableAccordion {
 				if ( el.getAttribute( 'id' ) ) {
 					elsAnchors[ el.getAttribute( 'id' ) ] = el
 				}
+
```
bfintal:
```suggestion
```

PR #3295, opened 2024-08-19
fix: lower specificity for hover transitions

### LINKED ISSUE #3266: Transitions with high specificity added
Yes. Related to high specificity changes due to the fix from this [issue](https://github.com/gambitph/Stackable/issues/3207)
See more information here: https://www.notion.so/Hannes-51fd0ae8db7a48f2b5ee6900e516e1d2

Lower the specificity to make the styles easier to overwrite.

N/A

Arukuen:
fixes #3266 

image

- Lowers specificity to from (0,3,1) to (0,1,0)
- Added the variable `--stk-transition-default` that can be used for overwriting the transition property

### DIFF
--- a/src/styles/block-transitions.scss
+++ b/src/styles/block-transitions.scss
@@ -2,20 +2,20 @@
  * This defines all styles related to making the hover states of blocks animate smoothly.
  */
 // Hover transitions.
-body.stk--anim-init {
-	:is(
+.stk--anim-init {
+	:where(
 	.stk-block,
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
After reviewing this PR and the changes that we will introduce, I am uneasy changing a huge amount of specificity to something that's already working without any problems.

Let's change this PR to just include the new custom property `--stk-transition-default`. Let's revert back to the original selectors.

PR #3298, opened 2024-08-21
fix/3109 query loop class

### LINKED ISSUE #3109: Some Columns block settings aren't working properly if Columns block is inside Query Loop block
**Describe the bug**
- Issue only in editor. Works correctly on frontend
- Happens to query loop blocks such as:
    - WordPress Query Loop block
    - Advanced Posts block (Blocksy)  
- Settings in the Layout tab such as:
    - [ ] Inner Column Spacing - doesn't work
    - [ ] Column Gap - doesn't work
    - [ ] Row Gap - doesn't work
    - [ ] Block / Content Width - doesn't work
    - [ ] Content Max Width - doesn't work
    - [ ] Content Horizontal Align - doesn't work
    - [ ] Column Justify - not working properly

**To Reproduce**
1. Add a new page
2. Add a Query Loop block or Advanced Posts block
3. Add a Columns block inside the **Post Template** block
4. Try changing settings mentioned above
5. See that it doesn't work as expected

**Expected behavior**
Settings should work as expected. 

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/386ea31a-ccb0-49f4-9803-fb68fa9a4baf

Arukuen:
fixes #3109

Note that the issue is also arises in the Horizontal Scroller Block.

- The class of blocks for styling in the editor looks like this: `stk-abc1234`
	- For blocks inside Query Loop, instanceId is added like this: `stk-abc1234-1`, `stk-abc1234-2`, etc.
	- These are targeted properly in the `BlockCSS `(found in style tag) providing styles in the editor
- However, for Columns and Horizontal Scroller Block, the targeted class (found in style tag) for styling looks like this respectively: `stk-abc1234-column` and `stk-abc1234-horizontal-scroller`
	- If inside Query Loop, instanceId is added like this: `stk-abc1234-1-column`, `stk-abc1234-2-column`, etc.
	- However, instanceId is not considered in generating the class (implemented via `getContentAlignmentClasses`).
	- Solution is to add appropriate instanceId to the generated class.

### DIFF
--- a/src/block-components/content-align/use-content-align.js
+++ b/src/block-components/content-align/use-content-align.js
@@ -4,10 +4,15 @@
 import { applyFilters } from '@wordpress/hooks'
 import classnames from 'classnames'
 
-export const getContentAlignmentClasses = ( attributes, blockName = 'column' ) => {
+export const getContentAlignmentClasses = ( attributes, instanceId, blockName = 'column' ) => {
+	let instanceIdString = ''
+	if ( instanceId ) {
+		instanceIdString = `${ instanceId }-`
+	}
+
 	return classnames(
 		'stk-content-align',
-		`stk-${ attributes.uniqueId }-${ blockName }`,
+		`stk-${ attributes.uniqueId }-${ instanceIdString }${ blockName }`,
 		applyFilters( 'stackable.block-components.content-align.getContentAlignmentClasses', {
 			'stk--flex': attributes.columnJustify,
 			alignwide: attributes.innerBlockContentAlign === 'alignwide', // This will align the columns inside.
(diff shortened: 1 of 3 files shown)


bfintal (changes requested):
1. The change to the function argument arrangement will cause issues since the function is currently used. Please use a different arrangement:

```diff
- getContentAlignmentClasses = ( attributes, instanceId, blockName = 'column' )
+ getContentAlignmentClasses = ( attributes,  blockName = 'column', instanceId = '' )
```

2. A question, I see changes only for the editor, does the issue only exist in the editor?

Arukuen:
Apparently, class names in the frontend do not use the same class appended with the `instanceId`, but instead are unique from each other.

Editor:
361630704-cf0c8681-3d77-4b68-982b-7eeb264b89f2
image


Frontend:
Untitled

bfintal:
@andeng1106 Will need to test:
- Block updating / deprecation
- Works in backend & frontend

PR #3299, opened 2024-08-21
fix: editing/adding global colors when blocks are present is slow

bfintal:
Adjustments:
- When adding global colors, fallback colors will no longer be updated
- When using global colors, fallback colors will no longer be added

Here's what this adjustment looks like:
```diff
- color: var(--stk-global-color-1234, #ff00ff);
+ color: var(--stk-global-color-1234);
```


andeng1106:
@bfintal Text Color name is no longer showing when selecting existing global text colors in the palette. It's showing the var() instead:

https://github.com/user-attachments/assets/929c9652-beff-4277-a5e7-3d9a89513937

PR #3317, opened 2024-09-17
Feat: Icon Library

### LINKED ISSUE #3316: New Feature: Icon Library
Design Document: https://www.notion.so/Icon-Library-de83b3adcceb4b48ac280521607b13d6

kaeizen:
resolves #3316


src/components/font-awesome-icon/index.js
```diff
@@ -80,6 +80,7 @@ const addSVGAttributes = ( svgHTML, attributesToAdd = {}, attributesToRemove = [
 }
 
 const FontAwesomeIcon = memo( props => {
+	const { svgAttrsToAdd = { width: '32', height: '32' }, svgAttrsToRemove = [ 'id', 'data-name' ] } = props
```
bfintal:
```suggestion
	const { 
		svgAttrsToAdd = { width: '32', height: '32' },
		svgAttrsToRemove = [ 'id', 'data-name' ],
	} = props
```

bfintal:
## Change Requests:

1. Let's change the "Icon Settings" in the icon picker popup into a "Library" icon, then when clicked, it should open the Icon Library in the sidebar

### In the icon picker in the icon library

image

Screenshot 2024-10-08 at 8 17 20 PM

2. See above image as reference, change the popover to only have:
  - Drop icon similar to the screenshot above
  - Mid aligned text that says: "Drag an SVG icon here or click the button below upload one"
  - Button that says "Upload SVG"

3. Dragging an SVG icon should make the popover look/behave the same way as the normal icon popover when dragging on 

### When clicking the plus button in the icon library

4. Change the popover to the same popover mentioned above

src/components/icon-search-popover/index.js
```diff
@@ -243,13 +249,24 @@ const IconSearchPopover = props => {
 				</div>
 				<div className="ugb-icon-popover__iconlist">
 					{ isBusy && <Spinner /> }
-					{ ! isBusy && results.map( ( { prefix, iconName }, i ) => {
+					{ ! isBusy && applyFilters( 'stackable.global-settings.inspector.icon-library.icons', null, {
```
bfintal: This is a bit hard to read, I would like to suggest moving the `applyFilters` outside this area and change this to a component instead.

For example, somewhere near the top of the component:

```js
const IconLibraryIcons = applyFilters( 'stackable.global-settings.inspector.icon-library.icons', Fragment )
```

Then in this component:

```jsx
{ ! isBusy && (
	<IconLibraryIcons
		icons={ results.iconLibrary }
		onChange={ props.onChange }
		onClose={ props.onClose }
	/>
) }
```

you can add an `isPro` constant check in there as well to make sure it doesn't render if not in the premium version

src/components/icon-search-popover/index.js
```diff
@@ -150,6 +151,11 @@ const IconSearchPopover = props => {
 			fr.onload = function( e ) {
 				setIsDropping( false )
 				const svgString = cleanSvgString( addCustomIconClass( e.target.result ) )
+
+				if ( isPro && props.showPrompt ) {
+					applyFilters( 'stackable.global-settings.inspector.icon-library.prompt', null, svgString )
+				}
+
```
bfintal: Instead of creating a hook specifically for the icon library, it would be best to create a hook for when an icon is loaded in general. Also, use `doAction` here since we are not doing anything to the output value. 

Note the change to `doAction` and hook name:
```suggestion
				doAction( 'stackable.icon-search-popover.svg-upload', svgString )

```

The checking if `isPro` can be done in the premium side:

```js
// Somewhere in the premium code:
if ( isPro ) {
	addAction( 'stackable.icon-search-popover.svg-upload', svgString => {
		// ...
	} )
}
```

bfintal (changes requested):

PR #3319, opened 2024-09-23
feat: granular plugin setting

### LINKED ISSUE #3318: Granular Plugin Settings

### LINKED ISSUE #3354: Granular Plugin Settings

Arukuen:
fixes #3354


bfintal (changes requested):
Some comments:

1. After I activated this PR, all the Stackable blocks disappeared from my editor (steps: activate the PR, open the editor)

2. Adjusting settings should show a "Unsaved changes" near the save button. We can also make the button grayed out if no changes have been made yet (but still clickable)

3. If you search for random characters (e.g. "kwjdnqwkhdnq") you'll see all the headings remaining

4. Remove all emojis from the headings since some have emojis and some don't

5. Looks like a lot of the settings have line separators between them. Lines should only appear between different sections of settings. Only the red lines here should have lines:

Screenshot 2024-11-06 at 9 41 41 AM

Arukuen:
@bfintal The latest commits are for migrating from `disabled_blocks` array to `block_states` object to fix disappearing blocks. Since the previous definition of disabled is hidden, when migrating, should the previous disabled blocks be hidden in the new implementation?

bfintal:
@andeng1106 For the QA, we need to verify the following:
1. A clean install of this build in a new site (never installed Stackable before), all the blocks are present
2. When updating to this build, if you haven't touched the settings before (you installed an older version of Stackable previously but didn't touch the settings), all the blocks should be present
3. When updating to this build, and you have previously saved the settings before (a non-block setting, for example disabling the design library button), after updating to this build, all the blocks should be present
4. When updating to this build, and you have previously saved the settings before (you disabled 1 block in the settings), after updating to this build, all blocks should be present except for the 1 disabled block (it should now show as "hidden")
5. The number 4 above should also work if multiple blocks were previously disabled

PR #3320, opened 2024-09-23
Fix: Performance Cleanup

kaeizen:
- [x] remove useBlockContext
- [x] Remove the useGeneratedCss hook in src/block-components/style/use-generated-css.js


bfintal:
@mxkae kindly resolve the conflicts

bfintal:
@andeng1106 kindly test whether the Icon block inside the accordion block's title has 2 icon pickers? And an icon block outside an accordion block should only have 1 icon picker

bfintal:
@andeng1106 here are the additional things to test to check whether everything still looks fine:

Things to test:
- changing column widths from the columns block (changing and resetting values for desktop/tablet/mobile)
- changing column widths from the inner column block (changing and resetting values for desktop/tablet/mobile)
- drag handlers should show correctly for inner column blocks inside columns block (left-most & right-most column should show 1 drag handler only, others 2 drag handlers)
- image block (2 drag handles should show)
- card block (1 drag image handle show)
- experimental block linking feature: block linking button on the corner should only show up if there are more than 1 inner column inside a columns block
- bottom margin handler should show up: accordion block, blockquote block, button group block
- accordion block, icon block should show 2 icon pickers
- blockquote block: icon label option should show “Pick another icon”
- button group block: “add button” placeholder should show up if there are no child buttons in the button group block
- icon label: heading block inside should NOT have top and bottom line inspector options
- icon list block: you should be able to delete a second list item block by pressing backspace until there are no characters left (and backspace again to delete the block)
- icon list block: if you have multiple icons for different bullet points, clicking on the icon reset button from the main icon list block should reset all the icons
- image box block: the image block inside it should not have a “link” option under the advanced tab
- image block: there should be a “link” option under the advanced tab
- default styles: inserting using a block with default styles should work
- default styles: inserting a block that has variations/layouts, then picking the “default” option should show the default styles
- default styles: inserting a block that has variations/layouts, then picking the non-default option should NOT show the default styles
- block hover states styles should still work

PR #3323, opened 2024-09-24
Fix (Icon List): add use tag on kses

kaeizen:


andeng1106:
Note only: 
Steps taken before fix:
1. Log in as contributor
2. Add a new post
3. Add an Icon List
4. Save page
5. Log in as admin
6. See post from step 2
7. See block errors

PR #3349, opened 2024-10-09
fix(heading): initially follow theme's margin if OceanWP is used

### LINKED ISSUE #3259: [OceanWP] Stackable blocks added to Posts are already aligned as Full Width
**Describe the bug**
- Happens only for Posts (ok for Pages)
- Issue happen on OceanWP theme
- Happens if Page Layout is set to "Full Width" or "100% Full Width" (OceanWP Page Settings & Styling > General > Layout)

**To Reproduce**
1. Activate OceanWP theme
2. Activate [Ocean Extra](https://wordpress.org/plugins/ocean-extra/) plugin
3. Add a new post
4. Add these blocks:
    - WordPress Heading block
    - Stackable Heading block
5. In Editor, set Page Settings & Styling > General > Layout to "Full Width"
6. Save post and go to frontend
7. See that the Stackable Heading block is already in Full Width even if no block width alignment settings are present

**Expected behavior**
- Stackable blocks should not look full width yet if block width alignment is not yet set to "Full Width"

**Workaround**
- Enable "Use theme heading margins" for the Heading block

**Screenshots**

https://github.com/user-attachments/assets/ba0b73a9-34c3-4644-ad21-72c3d1c3dc8d

**Desktop:**
 - OS: MacOS
 - Browser: Chrome
 - Browser Version: Version 126.0.6478.183

Arukuen:
fixes #3259

### DIFF
--- a/src/block-components/typography/edit.js
+++ b/src/block-components/typography/edit.js
@@ -100,6 +100,15 @@ export const Controls = props => {
 
 	const onChangeContent = useCallback( text => setDebouncedText( escapeHTML( text ) ), [] )
 
+	// OceanWP compatibility.
+	// If OceanWP theme is used, initially use the theme's margin to follow its layout
+	useEffect( () => {
+		if ( hasRemoveMargins ) {
+			const value = getAttribute( 'useThemeTextMargins' )
+			updateAttribute( 'useThemeTextMargins', applyFilters( 'stackable.heading.edit.useThemeTextMargins', value ) )
+		}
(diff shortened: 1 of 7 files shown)


bfintal (changes requested):
Since the issue is with OceanWP only, it would be better to have a fix that would be applicable only with OceanWP and only targets the issue. Adjusting all blocks for all scenarios would pose more issues. Also, we have default margins for our blocks, this removes them

PR #3364, opened 2024-11-04
fix (accordion block): enable text selection if accordion is inside columns

### LINKED ISSUE #3350: Accordion block - Text content is not being selected / highlighted
**Describe the bug**
- Only happens for Accordion blocks from the [Accordion page](https://wpstackable.com/accordion-block/) in Stackable website
    - A newly added Accordion doesn't cause any issue
- Upon closing the Accordion then reopening, it doesn't work anymore.
- Issue happens **if Accordion is inside the Columns block**
- Another page to test that has the issue (possibly copied from the Stackable website): https://collegecareerlab.hosting.nyu.edu/exploratory-stage/
- Sandbox test page: https://gambitsandbox.instawp.xyz/andrea/accordion-block-unable-to-select-text-content/

**To Reproduce**
1. Add a new page
2. Add an Accordion block
3. Go to Accordion block page in Stackable website
4. Copy any ready-made Accordion templates
5. Paste it on the page created from step 1
7. Try to highlight text from the newly made Accordion block (see if it works)
8. Try to highlight text from the ready-made Accordion blocks (see if it doesn't work)

**Expected behavior**
- Should be able to highlight Accordion block text content inside when opening and closing the accordion.

**Workaround**
- Remove Accordion block out of the Columns block for now

**Possible solution:** 
- In the accordion block, make sure that after some delay, the “user-select” is back to normal as a fallback.

**Screenshots**

https://github.com/user-attachments/assets/657ae599-c48a-443b-943f-5eb2e0365eff

kaeizen:
fixes #3350

### DIFF
--- a/src/block/accordion/frontend-accordion.js
+++ b/src/block/accordion/frontend-accordion.js
@@ -54,13 +54,12 @@ class StackableAccordion {
 						}, ANIM_OPTS )
 					}
 
-					// When the animation is done, allow text selection again.
+					// When the animation is done, remove wrapper and allow text selection again.
 					el.anim.onfinish = el.anim.oncancel = () => {
+						if ( doWrapHack ) {
+							removeWrapperHack( el, wrapper )
+						}
 						el.style.userSelect = 'auto'
-					}
-
-					if ( doWrapHack ) {
-						removeWrapperHack( el, wrapper )
 					}
 				}
 			} )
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
Please add a fallback just incase the `onfinish` or `oncancel` doesn't fire. Like if after 700ms seconds, make sure that the accordion is selectable

PR #3369, opened 2024-11-19
Fix (Dynamic Content): prevent blurry images when using image optimizer

### LINKED ISSUE #3296: Image block issue with Dynamic Content and EWWW Image Optimizer
**Describe the bug**
- If Easy IO setting is activated for [EWWW Image Optimizer](https://wordpress.org/plugins/ewww-image-optimizer/), dynamic images becomes blurry.
Screenshot 2024-08-20 at 10 49 07 AM

- Issue happens if I use Dynamic Content for Image block
   - Any Dynamic Content field that displays images (e.g. Featured Image, ACF Image field, etc.)
- Issue happens even if I set Image Size to Full Size
- Sample test page: https://gambitsandbox.instawp.xyz/andrea/ewww-image-optimizer-issue/

**To Reproduce**
1. Add a new page
2. Add an Image block that fetches a Dynamic Content Image field
3. Save page and view frontend
4. See that the Image is high resolution and not blurred
5. Activate [EWWW Image Optimizer](https://wordpress.org/plugins/ewww-image-optimizer/)
6. In EWWW settings, go to Essential > Easy IO and activate the setting
7. Go back to step 4 and see see that dynamic images are now blurred

**Expected behavior**
- When using EWWW Image Optimizer > Easy IO setting, it should not show dynamic images as blurred for Image block

**Screenshots**

https://github.com/user-attachments/assets/e3b1ea7e-afa7-4b76-8c80-3684d5cf3321

kaeizen:
fixes #3296

### DIFF
--- a/src/block-components/image/image-optimizer-polyfill.js
+++ b/src/block-components/image/image-optimizer-polyfill.js
@@ -4,6 +4,11 @@
 import domReady from '@wordpress/dom-ready'
 
 class ImageOptimizerPolyfill {
+	/**
+	 * This script is loaded when EWWW Image Optimizer plugin is activated
+	 * If Easy IO setting is activated for EWWW Image Optimizer plugin, dynamic images becomes blurry.
+	 * This script fixes the issue by removing the &fit parameter from the srcset and src attributes
+	 */
 	init = () => {
 		const imgs = document.querySelectorAll( '.stk-block img' )
 		imgs.forEach( img => {
--- a/src/block-components/image/image-optimizer-polyfill.js
+++ b/src/block-components/image/image-optimizer-polyfill.js
@@ -1,0 +1,37 @@
+/**
+ * WordPress dependencies
+ */
+import domReady from '@wordpress/dom-ready'
+
+class ImageOptimizerPolyfill {
+	init = () => {
+		const imgs = document.querySelectorAll( '.stk-block img' )
+		imgs.forEach( img => {
+			if ( img.hasAttribute( 'srcset' ) ) {
+				let srcset = img.getAttribute( 'srcset' )
+				const pattern = /https?:\/\/[^\s,]+/g
+				const urls = srcset.match( pattern )
+
+				urls.forEach( url => {
+					const index = url.indexOf( '&fit' )
+					if ( index !== -1 ) {
+						const newUrl = url.slice( 0, index )
+						srcset = srcset.replace( url, newUrl )
+					}
+				} )
+
+				img.setAttribute( 'srcset', srcset )
+			}
+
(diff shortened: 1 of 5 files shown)


src/block-components/image/index.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( ! function_exists( 'stackable_load_image_optimizer_polyfill_frontend_script' ) ) {
+	function stackable_load_image_optimizer_polyfill_frontend_script( $block_content, $block ) {
+		if ( ! is_admin() ) {
```
bfintal, line 12: Let's only do this only when the EWWWW plugin is present

src/block-components/image/index.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( ! function_exists( 'stackable_load_image_optimizer_polyfill_frontend_script' ) ) {
+	function stackable_load_image_optimizer_polyfill_frontend_script( $block_content, $block ) {
```
bfintal, line 9: Please add some comments here to explain why we need to load this file when EWWWW is present

src/block-components/image/image-optimizer-polyfill.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * WordPress dependencies
+ */
+import domReady from '@wordpress/dom-ready'
+
+class ImageOptimizerPolyfill {
```
bfintal, line 6: Please add a description of what this file is for, what it does, why we need it, and when it is loaded

bfintal (changes requested):
Let's add some documentation to help us in the future.

Also, let's only load the file only when the EWWWW plugin is activated

src/block-components/image/index.php
```diff
@@ -0,0 +1,30 @@
...
+				STACKABLE_VERSION,
+				true
+			);
+
+			// Only do this once.
+			remove_action( 'stackable/enqueue_scripts', 'stackable_load_image_optimizer_polyfill_frontend_script', 10 );
+		}
+	}
+
+	if ( ! is_admin() && is_plugin_active( 'ewww-image-optimizer/ewww-image-optimizer.php' )) {
```
bfintal: This would not work if the EWWW folder is renamed. It's would be better if you can use a checker if a function or constant from EWWW is present

bfintal (changes requested):

PR #3381, opened 2024-12-17
Fix (WooCommerce): remove css styles displayed on shop page

kaeizen:


src/kses.php
```diff
@@ -8,6 +8,29 @@
...
+			// remove CSS before kses strips out <style> tags
+			if ( ! empty( $optimized_css ) ) {
+				$content = str_replace( '<style>' . $optimized_css . '</style>', '', $content );
+			}
+
+		}
+		return $content;
+	}
+
+	add_filter('pre_kses', 'stackable_pre_kses_woocomerce_shop', 10, 3);
```
bfintal: See whether you can check for the existence of the `is_shop` function prior to doing the `add_filter`

src/css-optimize.php
```diff
@@ -209,12 +209,19 @@ public static function parse_block_style( $block, &$style_arr ) {
...
 		public function load_cached_css_for_post() {
-			// DEV NOTE: If we'll also do this for wp_template and
+			// DEV NOTE #1: If we'll also do this for wp_template and
 			// wp_template_part then we might need to use the actions:
 			// render_block_core_template_part_post and
 			// render_block_core_template_part_file
-			if ( is_singular() && ! is_preview() && ! is_attachment() ) {
-				$post_id = get_the_ID();
+			// DEV NOTE #2: Check for WooCommerce Shop Page as well
+			if ( ( is_singular() && ! is_preview() && ! is_attachment() ) ||  ( function_exists('is_shop' ) && is_shop() ) ) {
```
bfintal: This condition is becoming long and hard to read, can we split this to make it more readable? I feel uneasy that we're mingling WooCommerce code inside our normal behavior, let's add a filter so that we can split woocommerce compatibility into it's own file - we can add a `src/compatibility/woocommerce.php` and place the necessary compatibility code there.

```
$optimize_css = is_singular() && ! is_preview() && ! is_attachment();
$optimize_css = apply_filters( 'stackable/load_cached_css_for_post', $optimize_css );
if ( $optimize_css ) {
    // ...
```

src/kses.php
```diff
@@ -8,6 +8,29 @@
 	exit;
 }
 
+/**
+ * In WooCommerce Shop page, <style> tags are stripped out and the CSS styles are displayed in the frontend.
+ * This function removes the <style> tags and CSS styles before they are stripped out.
+ */
+if ( ! function_exists( 'stackable_pre_kses_woocomerce_shop' ) ) {
```
bfintal: Can we move this code from here to the new `src/compatibility/woocommerce.php` file suggested above?

bfintal (changes requested):
I suggest we create a new file specifically for WC compatibility.

PR #3399, opened 2025-01-22
fix: only escape HTML if invalid

### LINKED ISSUE #3392: Text Formatting issue - When editing text in the side editor, text format breaks
Describe the bug

If text is edited on the side editor, the text formatting breaks.

To Reproduce

1. Add a new page
2. Add a Text block
3. Type a text
4. Highlight a word in bold
5. Enter text within the font style in the side editor

Expected behavior

When entering text within a defined font style in the side editor, the text should automatically follow and display the formatting of the specified font style settings.

Screenshot

https://github.com/user-attachments/assets/81200f8c-c683-4266-aad2-fed1ab1eb676

Arukuen:
fixes #3392

### DIFF
--- a/src/block-components/typography/util.js
+++ b/src/block-components/typography/util.js
@@ -1,0 +1,14 @@
+import { escapeHTML } from '@wordpress/escape-html'
+
+export const escapeHTMLIfInvalid = html => {
+	const parser = new DOMParser()
+	const doc = parser.parseFromString( html, 'text/html' )
+	const parseError = doc.querySelector( 'parsererror' )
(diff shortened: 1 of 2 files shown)


src/block-components/typography/util.js
```diff
@@ -0,0 +1,14 @@
+import { escapeHTML } from '@wordpress/escape-html'
+
+export const escapeHTMLIfInvalid = html => {
+	const parser = new DOMParser()
```
bfintal, line 4: Is there any performance issues because we're doing this every keypress?

PR #3416, opened 2025-02-05
fix: use selector with direct descendant, tested old issue

### LINKED ISSUE #3398: Columns Block -  Nested Columns displayed in Reverse Order on Tablet/Mobile When Parent Column arrangement Changes
Describe the bug
When using nested columns (columns within a column), the inner columns are displayed in reversed order on tablet and mobile (only happens when the column arrangement is different from the desktop view)

To Reproduce
1. Add a Two Column Block
2. Add Nested Columns: In the first column, add a two-column block (50-50 structure).
3. Add Content:
- First parent column: Add Text in both columns.
- Second parent column: Add Text in both columns.
4. Change the column arrangement on the tablet/mobile different from the desktop.

Expected behavior
The nested columns (inner columns) should remain in the same order as they are on desktop view, even when the parent column arrangement changes for tablet/mobile views.

Screenshots

https://github.com/user-attachments/assets/4d415700-5ebc-4b9c-91b5-2e001d916f4a

### LINKED ISSUE #3406: Columns Block - Accordion opens upwards when Column arrangement changes on Mobile breakpoints
**Describe the bug**
When the column arrangement changes in mobile view (second column placed above the first), the accordion expands upwards.
- Happens on mobile breakpoints

**To Reproduce**
1. Add a new page
2. Add a two two-column block
3. Add an accordion on the first column, add an image on the second column on each column block
4. On the second column block, in mobile view, make the second column on top of the first column

**Expected behavior**
The accordion should expand downward regardless of column reordering.

**Screenshots**

Image 

Image

Arukuen:
fixes #3398
fixes #3406 

Tested the previous issue on Local with multisite (subdirectories), with non-admin account (editor). No block error so far.

### DIFF
--- a/src/block-components/columns/style.js
+++ b/src/block-components/columns/style.js
@@ -133,12 +133,14 @@ export const addStyles = ( blockStyleGenerator, props = {} ) => {
 			} )
 
 			addBlockCssFunc( {
-				// In the frontend, use the css custom property to specify the order.
+				// Old Issue: In the frontend, use the css custom property to specify the order.
 				// See style.scss We do this because we cannot use the direct descendent
 				// selector ">" in the saved CSS or else non-admin users in multisite
 				// will encounter block errors (this is a WP issue)
+				// Tested on Local with multisite (subdirectories) with non-admin account (editor).
 				renderIn: 'save',
-				styleRule: `--stk-col-order-${ i }`,
+				selector: `.%s-column > :nth-child(${ i })`,
+				styleRule: `order`,
 				responsive: 'all',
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
Changing the way column ordering works by changing the main selector will cause unforeseen problems. Since the underlying issue is that the reordering styles are also affecting nested columns, the better solution would be to introduce something that limits the styles from only being applied to the current columns (and not to their children)

bfintal:
Closed in favor of #3423

PR #3438, opened 2025-02-13
Fix: apply responsive breakpoints filters correctly

### LINKED ISSUE #3434: Blocksy: responsive breakpoints not followed
The filter for adjusting the responsive breakpoints via `stackable_responsive_breakpoints` seems to be not working anymore.

Might be related to #3394 

Steps to replicate:
1. have this filter in place:

Image

2. Make sure there are no Stackable responsive settings set
3. Check in the frontend, the blocks use `max-width: 1023px` for tablet and `max-width: 767px` for mobile

https://github.com/user-attachments/assets/c8d14d47-0c73-4913-b1ba-63cb2715422d

kaeizen:
fixes #3434

### DIFF
--- a/src/dynamic-breakpoints.php
+++ b/src/dynamic-breakpoints.php
@@ -35,17 +35,23 @@ if ( ! class_exists( 'Stackable_Dynamic_Breakpoints' ) ) {
 				// Add a filter for replacing shortcut media queries before the breakpoint adjustment.
 				add_filter( 'stackable_frontend_css', array( $this, 'replace_shortcut_media_queries' ), 9 );
 
-				if ( $this->has_custom_breakpoints() ) {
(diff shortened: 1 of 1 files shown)


src/dynamic-breakpoints.php
```diff
@@ -23,6 +23,8 @@ function stackable_get_responsive_css() {
 if ( ! class_exists( 'Stackable_Dynamic_Breakpoints' ) ) {
 	class Stackable_Dynamic_Breakpoints {
 
+		private $dynamic_breakpoints = false;
```
bfintal, line 27:
```suggestion
		// Holds the value of the saved or default breakpoints
		private $dynamic_breakpoints = false;
```

PR #3447, opened 2025-02-25
fix(icon-label): migration condition, explicit zero should be true

### LINKED ISSUE #3426: Icon Label Block - Icon Gap Resets to a Different Value When an Icon Size Is Set
**Describe the bug**
If an Icon Size has a set value, the Icon Gap resets to a different value instead of remaining at 0 when it is set to 0 in all views.
- Happens in editor
- Happens in frontend

**To Reproduce**
1. Add a page
2. Add an Icon Label Block
3. On the Icon Label Block, set the Icon Gap value to 0 in Desktop, Tablet, and Mobile views
4. On the Icon Block, increase the icon size value to 30
5. Save the page and check on frontend (no icon gap)
6. Refresh the page
7. See the error on the editor, there is an icon gap visible and the value is 34
8. Add the word "block" on the Icon Label Heading and Save page
9. See the error on frontend (Icon Gap is visible)

**Expected behavior**
- When an Icon Size is set to any value, the Icon Gap should remain at the specified value across all views.
- If the Icon Gap value is set to 0 for Desktop, Tablet, and Mobile views, it should stay at 0 and not reset to another value.
- Any content updates should not cause the Icon Gap to change unexpectedly

**Screenshots**

https://github.com/user-attachments/assets/a9a24a4f-2a28-4622-883a-de24ac638c95

Arukuen:
fixes #3426

### DIFF
--- a/src/block/icon-label/deprecated.js
+++ b/src/block/icon-label/deprecated.js
@@ -16,10 +16,13 @@ const deprecated = [
 				return false
 			}
 
+			// Explicit zero should be true
+			const normalizeZero = value => value || value === 0
+
 			const iconBlockAttributes = innerBlocks[ 0 ].attributes
 			const hasIconSize = iconBlockAttributes.iconSize || iconBlockAttributes.iconSizeTablet || iconBlockAttributes.iconSizeMobile ? true : false
-			const hasIconGap = attributes.iconGap || attributes.iconGapTablet || attributes.iconGapMobile ? true : false
-			const hasIconGap2 = attributes.iconGap2 || attributes.iconGap2Tablet || attributes.iconGap2Mobile ? true : false
+			const hasIconGap = normalizeZero( attributes.iconGap ) || attributes.iconGapTablet || attributes.iconGapMobile ? true : false
+			const hasIconGap2 = normalizeZero( attributes.iconGap2 ) || attributes.iconGap2Tablet || attributes.iconGap2Mobile ? true : false
 			return hasIconGap || ( hasIconSize && ! hasIconGap2 )
 		},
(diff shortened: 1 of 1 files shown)


src/block/icon-label/deprecated.js
```diff
@@ -16,10 +16,13 @@ const deprecated = [
...
 
+			// Explicit zero should be true
+			const normalizeZero = value => value || value === 0
+
 			const iconBlockAttributes = innerBlocks[ 0 ].attributes
 			const hasIconSize = iconBlockAttributes.iconSize || iconBlockAttributes.iconSizeTablet || iconBlockAttributes.iconSizeMobile ? true : false
-			const hasIconGap = attributes.iconGap || attributes.iconGapTablet || attributes.iconGapMobile ? true : false
-			const hasIconGap2 = attributes.iconGap2 || attributes.iconGap2Tablet || attributes.iconGap2Mobile ? true : false
+			const hasIconGap = normalizeZero( attributes.iconGap ) || attributes.iconGapTablet || attributes.iconGapMobile ? true : false
+			const hasIconGap2 = normalizeZero( attributes.iconGap2 ) || attributes.iconGap2Tablet || attributes.iconGap2Mobile ? true : false
```
bfintal: For readability and simplicity, I would like to suggest just directly adding the condition.
```suggestion
			const hasIconGap = attributes.iconGap || attributes.iconGap === 0 || attributes.iconGapTablet || attributes.iconGapMobile ? true : false
			const hasIconGap2 = attributes.iconGap2 || attributes.iconGap2 === 0 || attributes.iconGap2Tablet || attributes.iconGap2Mobile ? true : false
```

PR #3449, opened 2025-02-27
feat/3448 global font pairs

### LINKED ISSUE #3448: Global Font Pairs

Arukuen:
fixes #3448


src/global-settings.php
```diff
@@ -282,6 +282,53 @@ public function register_global_settings() {
 				)
 			);
 
+			register_setting(
+				'stackable_global_settings',
+				'stackable_font_pair_name',
```
bfintal: I suggest we change this name to something more descriptive to what it is for:
```suggestion
				'stackable_selected_font_pair',
```

src/plugins/global-settings/typography/index.js
```diff
@@ -69,14 +72,22 @@ const TYPOGRAPHY_TAGS = [
...
+	const [ editingFontPairName, setEditingFontPairName ] = useState( '' )
+
+	const fontPairContainerRef = useRef( null )
 
 	useEffect( () => {
 		fetchSettings().then( response => {
 			// Get settings.
 			setTypographySettings( ( head( response.stackable_global_typography ) ) || {} )
 			setApplySettingsTo( response.stackable_global_typography_apply_to || 'blocks-stackable-native' )
+			setCustomFontPairs( response.stackable_custom_font_pairs || [] )
```
bfintal, line 93: We'll need to move this out of free

src/plugins/global-settings/typography/index.js
```diff
@@ -139,31 +205,130 @@ addFilter( 'stackable.global-settings.inspector', 'stackable/global-typography',
...
+										<span
+											style={ omit( { ...paragraphStyles }, [ 'fontSize', 'lineHeight' ] ) }
+											className="ugb-global-settings-font-pair__sub-label"
+										>
+											{ paragraphStyles?.fontFamily ?? 'Theme Body Default' }
+										</span>
+									</div>
+								)
+
+								const className = classNames( { 'ugb-global-settings-font-pair__selected': selectedFontPairName === fontPair.name } )
```
bfintal: I recommend moving this all inside the FontPairPicker component, then adding the necessary props to it:
```js
<FontPairPicker
  // ...
  isSelected={ selectedFontPairName === fontPair.name }
  headingFontFamily={ headingStyles?.fontFamily }
  bodyFontFamily={ paragraphStyles?.fontFamily }
/>
```

src/plugins/global-settings/typography/index.js
```diff
@@ -139,31 +205,130 @@ addFilter( 'stackable.global-settings.inspector', 'stackable/global-typography',
...
+							<h3>Preset Font Pairs</h3>
+							<Button
+								className="ugb-global-settings-color-picker__add-button"
+								onClick={ addFontPair }
+								icon="plus-alt2"
+							/>
+						</div>
+
+						<div className="ugb-global-settings-font-pair__container" ref={ fontPairContainerRef }>
+							{ allFontPairs.map( fontPair => {
```
bfintal: Instead of mapping through `allFontPairs` and updating the list, you can just output 3 times:
1. For the default font pair
2. ~Custom font pairs - this can be an `applyFilters( 'stackable.global-settings.typography.font-pairs.custom', null, selectedFontPairName )`~ use alternative method below
3. All other preset font pairs

bfintal: Alternative
```js
// Near the top
const CustomPicker = applyFilters( 'stackable.global-settings.typography.font-pairs.customPicker', Fragment )

// Render the custom font pair pickers
<CustomPicker
  isSelected={ selectedFontPairName === fontPair.name }
  headingFontFamily={ headingStyles?.fontFamily }
  bodyFontFamily={ paragraphStyles?.fontFamily }
/>
```

In premium:
```js
const CustomFontPairsPicker = props => {
  const [ customFontPairs, setCustomFontPairs ] = setState( [] )

  useEffect( () => {
    // Load all custom font pairs
    const customFontPairs = []
  }, [] )

  return customFontPairs.map( fontPair => {
    return <FontPairPicker
      // ...pass stuff
    />
  }
}

addFilter( 'stackable.global-settings.typography.font-pairs.customPicker', 'stackable/global-settings/font-pairs', CustomFontPairsPicker )
```

bfintal (changes requested):

src/plugins/global-settings/typography/index.js
```diff
@@ -139,31 +197,114 @@ addFilter( 'stackable.global-settings.inspector', 'stackable/global-typography',
...
+									fontPair={ fontPair }
+									isSelected={ selectedFontPairName === fontPair.name }
+									onClick={ () => {
+										updateSelectedFontPair( fontPair.name )
+										changeStyles( fontPair.typography )
+									} }
+								/>
+							} ) }
+						</div>
+						{ ! isPro && <ProControl type="global-font-pairs" title="Creating and editing Custom Font Pairs are premium features" /> }
```
bfintal: Use this instead:

```js
{ showProNotice && <ProControlButton type="global-font-pairs" /> }
```

Then add the proper text for this new `type`

src/components/font-pair-picker/index.js
```diff
@@ -0,0 +1,90 @@
...
+		loadGoogleFont( paragraphStyles.fontFamily )
+	}
+
+	const label = (
+		<div>
+			<span
+				style={ omit( { ...headingStyles }, [ 'fontSize', 'lineHeight' ] ) }
+				className="ugb-global-settings-font-pair__label"
+			>
+				{ headingStyles.fontFamily ? headingStyles.fontFamily : 'Theme Heading Default' }
```
bfintal: This is too long, please use "Default Heading" and "Default Body" instead.

Also, please go through all the changes, all text strings should be translatable:

```suggestion
				{ headingStyles.fontFamily ? headingStyles.fontFamily : __( 'Default Heading', i18n ) }
```

bfintal (changes requested):

src/plugins/global-settings/typography/index.js
```diff
@@ -139,31 +209,162 @@ addFilter( 'stackable.global-settings.inspector', 'stackable/global-typography',
 		model.save()
 	}
 
+	const getIsAllowReset = selector => {
+		const currentFontPair = [ ...FONT_PAIRS, ...customFontPairs ].find( fontPair => fontPair.name === selectedFontPairName )
+		const fontPairStyle = currentFontPair.typography[ selector ]
```
bfintal: I encountered an error here where the `currentFontPair` was undefined and I couldn't anymore access the global settings.

Feels like this is prone now to errors. since I've encountered a few times where the editor stalled and I couldn't access the settings again. Looks like there is an assumption that font pairs are always objects, and if something happened (e.g. user didn't select a font pair, or deleted a font pair), things produce an error. Please adjust so that this case is not assumed

bfintal (changes requested):

.vscode/settings.json
```diff
@@ -11,4 +11,4 @@
 		"css",
 		"scss"
 	]
-}
+}
```
bfintal:
```suggestion
}

```

src/components/typography-control/index.js
```diff
@@ -35,14 +35,7 @@ const TypographyControl = props => {
...
 				onReset={ props.onReset }
-				allowReset={
-					props.fontFamily ||
-					props.fontSize || props.tabletFontSize || props.mobileFontSize ||
-					props.fontWeight ||
-					props.textTransform ||
-					props.lineHeight || props.tabletLineHeight || props.mobileLineHeight ||
-					props.letterSpacing || props.tabletLetterSpacing || props.mobileLetterSpacing
-				}
+				allowReset={ props?.isAllowReset }
```
bfintal: I would suggest not removing existing functionality here. If the intention is to provide a new prop that can override the `allowReset` then I would prefer that we keep the old functionality back, but only when the new prop `isAllowReset` is passed, then that new one should be used instead.

```js
const {
    isAllowReset = undefined, // Optional prop
} = props

// Allow overriding, or use our original condition
const _allowReset = typeof props.isAllowReset !== undefined ? props.isAllowReset : (
	props.fontFamily ||
	props.fontSize || props.tabletFontSize || props.mobileFontSize ||
	props.fontWeight ||
	props.textTransform ||
	props.lineHeight || props.tabletLineHeight || props.mobileLineHeight ||
	props.letterSpacing || props.tabletLetterSpacing || props.mobileLetterSpacing
)

// In the component props:
allowReset={ _allowReset }
```

bfintal (changes requested):

Issue #3241, opened 2024-07-11
Posts block > Meta - Date language is in English even if using a different language


andeng1106:
**Describe the bug**
- When I switch site language to a different one (e.g. Spanish), the Date language from Posts block > Meta - Date is still in English
- Issue started to happen on **3.13.0 (performance update)** and newer versions
- Issue only on frontend

**To Reproduce**
1. If site language is in English, switch site language to a different language (e.g. Spanish)
    - If site language is in another language, skip this step
2. Add a new page
3. Add a Posts block
4. Check Post Meta - Date if date language is in the **chosen language**
5. Save page and view frontend
6. Check Post Meta - Date if data language is still in **English** (issue)

**Expected behavior**
- Date language in Post Meta - Date should follow the chosen language

**Screenshots**

https://github.com/gambitph/Stackable/assets/103395655/890bc8e8-2207-4260-96f2-5833f5910f84

PR #3459, opened 2025-03-14
fix(post): use date_i18n since the date is already offset with timezone

### LINKED ISSUE #3458: Post - Setting timezone displays incorrect date
**Describe the bug**
When the timezone is set, it erroneously subtracts hours instead of maintaining the appropriate offset based on the timezone and the frontend displays an incorrect date
- Happens on frontend

**To Reproduce**
1. Navigate to Settings > General and set the timezone to UTC-5.
2. Add a new Post.
3. Set the date and time of the post to March 13, 2025, 12:00 AM.
4. Add a new Page.
5. Add a Posts block.
6. See error on frontend (date is March 12, 2025 - 2025-03-12T19:00:00-05:00)

**Expected behavior**
The timezone offset should be correctly applied and the frontend should display the correct date and time

**Screenshots**

https://github.com/user-attachments/assets/df6b6b1e-3a20-46cb-96af-f97a41b7bf66

Arukuen:
fixes #3458

### DIFF
--- a/src/block/posts/index.php
+++ b/src/block/posts/index.php
@@ -91,12 +91,13 @@ if ( ! function_exists( 'generate_render_item_from_stackable_posts_block' ) ) {
 
 		// Date.
 		if ( strpos( $new_template, '!#dateTime!#' ) !== false || strpos( $new_template, '!#date!#' ) !== false ) {
-			$datetime = wp_date( 'c', strtotime( $post['post_date'] ) );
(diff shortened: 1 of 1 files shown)


Arukuen:
Related to previous [issue](https://github.com/gambitph/Stackable/issues/3241).

bfintal:
I do not think this is the right fix for this since the docs https://developer.wordpress.org/reference/functions/wp_date/ say that `wp_date` replaces `date_i18n`

PR #3465, opened 2025-03-25
Feat (Video Popup): add video schema

### LINKED ISSUE #3402: Video Popup Block: Add Video Schema output
When using a Video popup block, the video doesn't get index by search engines

Automatically include a video schema in the output of the Video popup block.
- No need for an option, this will always be included in the output.

- Reference: https://www.semrush.com/blog/video-schema/
- We can output the same way the Accordion block outputs it's schema (but without the option to enable/disable it, this one is always enabled)

Add a new panel in the Style tab > General for schema (at the bottom, add a ControlSeparator then the schema fields:
- Video name (will be used for the schema `name` field, default to blank)
- Video upload date (will be used for the `uploadDate` field, default to blank)

The schema output should be generated PHP-side in the `render_block` function just before outputting and should have the following data:
- `name` - from the `name` field, or default to the `get_the_title()` if available, blank otherwise
- `thumbnailUrl` - should be acquired from the image src in the child image block if available, blank otherwise
- `uploadDate` - from the `uploadDate` field, or default to the publish date of the current post if available, blank otherwise
- `contentUrl` - the video URL

kaeizen:
fixes #3402 

Also added `description` in the schema to prevent "non-critical issues" flag in the [validation](https://search.google.com/test/rich-results).


src/block/video-popup/index.php
```diff
@@ -19,3 +19,60 @@ function stackable_load_videopopup_frontend_script() {
...
+			$video_schema[ '@context' ] = 'https://schema.org';
+			$video_schema[ '@type' ] = 'VideoObject';
+
+			// Get video schema properties from block attributes
+			$attributes = $block[ 'attrs' ];
+
+			// Get video name from the title of the post if not set
+			$name = isset( $attributes[ 'videoName' ] ) ? $attributes[ 'videoName' ] : ( get_the_title() ?? '');
+			// Get video upload date from the date of the post if not set
+			$upload_date = isset( $attributes[ 'videoUploadDate' ] ) ? $attributes[ 'videoUploadDate' ] : ( get_the_date( 'c' ) || '');
```
bfintal: Some errors on the date:
1. This is giving `true` as the value if no upload date was given.
2. If the date is given, it's showing an error:
image

Note: A passing value is of the format: `2025-04-10T03:24:47+00:00`

src/block/video-popup/index.php
```diff
@@ -19,3 +19,60 @@ function stackable_load_videopopup_frontend_script() {
...
+			if ( isset( $block[ 'innerBlocks' ] )
+				&& count( $block[ 'innerBlocks' ] ) === 2
+				&& $block[ 'innerBlocks' ][ 1 ][ 'blockName' ] === 'stackable/image'
+			) {
+				$image_attributes = $block[ 'innerBlocks' ][ 1 ][ 'attrs' ];
+				$thumbnail_url = isset( $image_attributes[ 'imageUrl' ] ) ? $image_attributes[ 'imageUrl' ] : '';
+				$video_schema[ 'thumbnailUrl' ] = $thumbnail_url;
+			}
+
+			$video_schema_json = json_encode( $video_schema, JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES );
```
bfintal: Use `wp_json_encode`
```suggestion
			$video_schema_json = wp_json_encode( $video_schema, JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES );
```

src/block/video-popup/index.php
```diff
@@ -19,3 +19,60 @@ function stackable_load_videopopup_frontend_script() {
...
+			$name = isset( $attributes[ 'videoName' ] ) ? $attributes[ 'videoName' ] : ( get_the_title() ?? '');
+			// Get video upload date from the date of the post if not set
+			$upload_date = isset( $attributes[ 'videoUploadDate' ] ) ? $attributes[ 'videoUploadDate' ] : ( get_the_date( 'c' ) || '');
+			$description = isset( $attributes[ 'videoDescription' ] ) ? $attributes[ 'videoDescription' ] : '';
+			$content_url = isset( $attributes[ 'videoLink' ] ) ? $attributes[ 'videoLink' ] : '';
+
+			$video_schema[ 'name' ] = $name;
+			$video_schema[ 'description' ] = $description;
+			$video_schema[ 'uploadDate' ] = $upload_date;
+			$video_schema[ 'contentUrl' ] = $content_url;
```
bfintal: Escape the values here for security

src/block/video-popup/index.php
```diff
@@ -19,3 +19,60 @@ function stackable_load_videopopup_frontend_script() {
...
+
+			// Get thumbnail URL from the image block if it exists
+			if ( isset( $block[ 'innerBlocks' ] )
+				&& count( $block[ 'innerBlocks' ] ) === 2
+				&& $block[ 'innerBlocks' ][ 1 ][ 'blockName' ] === 'stackable/image'
+			) {
+				$image_attributes = $block[ 'innerBlocks' ][ 1 ][ 'attrs' ];
+				$thumbnail_url = isset( $image_attributes[ 'imageUrl' ] ) ? $image_attributes[ 'imageUrl' ] : '';
+				$video_schema[ 'thumbnailUrl' ] = $thumbnail_url;
```
bfintal: Escape this for security

src/block/video-popup/index.php
```diff
@@ -19,3 +19,60 @@ function stackable_load_videopopup_frontend_script() {
...
+				// Compile all video schema entities into a single script
+				echo '<script type="application/ld+json"> [ ' . implode( ', ', $this->video_entities ) . ' ] </script>';
+			}
+		}
+
+		public function render_block_video_popup_schema( $block_content, $block ) {
+			// Initialize video schema
+			$video_schema = array();
+			$video_schema[ '@context' ] = 'https://schema.org';
+			$video_schema[ '@type' ] = 'VideoObject';
```
bfintal: You can initialize `$video_schema` with the default array values here right away instead of doing 2 additional assignments.

bfintal (changes requested):

src/block/countdown/editor.scss
```diff
@@ -1,4 +1,4 @@
-.ugb-panel--countdown .components-datetime__time-wrapper {
+:is(.ugb-panel--countdown, .ugb-panel--video-popup) .components-datetime__time-wrapper {
```
bfintal: Looks like this is adding extra styles tot he datetime picker. Instead of explicitly adding the class names for the countdown and video popup blocks, can we style the datetime component when it's used in the Stackable block inspector in general? Please move these styles for the datetime outside the countdown editor.scss into somewhere more proper.

bfintal (changes requested):

PR #3475, opened 2025-03-31
fix (color schemes): fix native button text turning black

kaeizen:


kaeizen:
close in favor of #3476

PR #3478, opened 2025-04-02
Feat (Color Schemes): Add help description to manage color schemes

### LINKED ISSUE #3471: Add help description to manage color schemes
Add a description below the color scheme pickers under **background** and **container**:

> Change the color scheme. Manage your color schemes

The manage should trigger the color schemes area to open:

```js
wp.data.dispatch('core/edit-post').openGeneralSidebar('stackable-global-settings/sidebar');
```

kaeizen:
fixes #3471

### DIFF
--- a/src/components/color-schemes-help/index.js
+++ b/src/components/color-schemes-help/index.js
@@ -13,7 +13,7 @@ export const ColorSchemesHelp = () => {
 
 	return <p>
 		{ __( 'Change the color scheme.', i18n ) }
-		<Button variant="link" onClick={ onClick }> { __( 'Manage', i18n ) } </Button>
+		<Button variant="link" onClick={ onClick }> { __( ' Manage', i18n ) } </Button>
 		{ __( ' your color schemes.', i18n ) }
 	</p>
 }
--- a/src/components/color-schemes-help/index.js
+++ b/src/components/color-schemes-help/index.js
@@ -1,0 +1,20 @@
+import { i18n } from 'stackable'
+
+import { __ } from '@wordpress/i18n'
+import { Button } from '@wordpress/components'
+import { dispatch } from '@wordpress/data'
+
+export const ColorSchemesHelp = () => {
+	const onClick = () => {
+		// Open the global settings sidebar and the color schemes panel.
+		dispatch( 'core/edit-post' ).openGeneralSidebar( 'stackable-global-settings/sidebar' )
+		dispatch( 'stackable/global-color-schemes' ).setIsOpen( true )
+	}
+
+	return <p>
+		{ __( 'Change the color scheme.', i18n ) }
+		<Button variant="link" onClick={ onClick }> { __( 'Manage', i18n ) } </Button>
+		{ __( ' your color schemes.', i18n ) }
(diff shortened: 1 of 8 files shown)


src/components/color-schemes-help/index.js
```diff
@@ -0,0 +1,20 @@
...
+	const onClick = () => {
+		// Open the global settings sidebar and the color schemes panel.
+		dispatch( 'core/edit-post' ).openGeneralSidebar( 'stackable-global-settings/sidebar' )
+		dispatch( 'stackable/global-color-schemes' ).setIsOpen( true )
+	}
+
+	return <p>
+		{ __( 'Change the color scheme. ', i18n ) }
+		<Button variant="link" onClick={ onClick }> { __( 'Manage', i18n ) } </Button>
+		{ __( ' your color schemes.', i18n ) }
```
bfintal: Use the entire `Manage color schemes` as the entire button text. 

Just a heads up, the text translations should make sense always for translators. These currently will read to translators as:
1. "Manage"
2. " your color schemes"
In this case it's better to merge the 2

Also, can you make this inline, so that "Change the color scheme. Manage your color schemes"

src/plugins/global-settings/color-schemes/index.js
bfintal, line 1: I feel that adding a new state just to make this work may be a bit over-engineering it. I would recommend to instead just do it the plain JS way:

When the manage button is clicked, the sidebar will open via `dispatch`, then check via plain JS if any other panels that should be closed are open, then do a click on them to close them. Then if the color scheme panel is closed, do a click on it.

src/components/color-schemes-help/index.js
```diff
@@ -0,0 +1,20 @@
+import { i18n } from 'stackable'
+
+import { __ } from '@wordpress/i18n'
+import { Button } from '@wordpress/components'
+import { dispatch } from '@wordpress/data'
+
+export const ColorSchemesHelp = () => {
+	const onClick = () => {
+		// Open the global settings sidebar and the color schemes panel.
+		dispatch( 'core/edit-post' ).openGeneralSidebar( 'stackable-global-settings/sidebar' )
```
bfintal: This might error in the site editor since I think `core/edit-post` is not available there.

bfintal (changes requested):

src/components/color-schemes-help/index.js
```diff
@@ -0,0 +1,38 @@
...
+			const colorSchemeToggle = document.querySelector( '.ugb-global-color-schemes__panel .components-panel__body-title > .components-panel__body-toggle' )
+			// Opens the color scheme panel
+			if ( colorSchemeToggle.getAttribute( 'aria-expanded' ) === 'false' ) {
+				colorSchemeToggle?.click()
+			}
+		}, 10 )
+	}
+
+	return <>
+		<span>{ __( 'Change the color scheme. ', i18n ) }</span>
```
bfintal:
```suggestion
		<span>{ __( 'Change the color scheme.', i18n ) }</span>
		&nbsp;
```

PR #3479, opened 2025-04-02
Fix (Performance): Improve performance for getting custom breakpoints

### LINKED ISSUE #3477: Improve performance of has_custom_breakpoints()
Running the a code profiler shows the the function `has_custom_breakpoints()` runs slowly. It might be because this function is called every block render, and checks for the presence of dynamic breakpoints every time. This function can cause slowdowns and should be optimized so that the dynamic breakpoints and its presence should be cached.

kaeizen:
fixes #3477

### DIFF
--- a/src/dynamic-breakpoints.php
+++ b/src/dynamic-breakpoints.php
@@ -36,9 +36,6 @@ if ( ! class_exists( 'Stackable_Dynamic_Breakpoints' ) ) {
 			add_action( 'rest_api_init', array( $this, 'register_settings' ) );
 
 			if ( is_frontend() ) {
-				// Get the dynamic breakpoints after the theme has loaded.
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
I feel that this implementation has a chance to break themes and other integrations that already successfully change the breakpoints, mainly because you are changing when the values are computed/loaded. Currently, the breakpoints are calculated when the css is about to be parsed, the new implementation does it early in the lifecycle, so existing hooks might not run correctly

Instead, I would recommend caching the values from within `get_dynamic_breakpoints()`.

Use code profiler to check this, maybe this line is also slow:
```php
return ! empty( $breakpoints['tablet'] ) || ! empty( $breakpoints['mobile'] );
```

PR #3482, opened 2025-04-04
feat/typography typescale

Arukuen:


src/block-components/columns/edit.js
```diff
@@ -284,6 +287,8 @@ export const Controls = props => {
 							video: 'column-gap',
 							description: __( 'Sets the distance between two or more columns', i18n ),
 						} }
+						marks={ presetMarks }
+						hasCSSVariableValue={ true }
```
bfintal, line 291: I think we had a discussion before to remove this prop `hasCSSVariableValue` since it can just be detected from the value

src/global-settings.php
```diff
@@ -638,11 +662,19 @@ public function form_tag_selector( $tag ) {
 		}
 
 		public function form_paragraph_selector() {
-			return array_merge(
-				$this->form_tag_selector( 'p' ), // Core text.
-				$this->form_tag_selector( 'li' ), // Core lists.
-				$this->form_tag_selector( 'td' ) // Core table cells.
+			$is_apply_body_to_html = get_option( 'stackable_is_apply_body_to_html' ) ?? false;
```
bfintal, line 665: I believe `get_option()` is never `null` since it returns `false` if the value is not yet set. So we can do just:
```suggestion
			$is_apply_body_to_html = get_option( 'stackable_is_apply_body_to_html' );
```

Can you also move this code to the `if` statement where it is used to make it closer and easier to read.

src/plugins/global-settings/preset-controls/index.php
```diff
@@ -0,0 +1,172 @@
...
+			add_filter( 'stackable_inline_editor_styles', array( $this, 'add_preset_controls_styles' ) );
+		}
+
+		public static function sanitize_array_setting( $input ) {
+			return ! is_array( $input ) ? array( array() ) : $input;
+		}
+
+		private function load_presets( $json_path ) {
+			if ( file_exists( $json_path ) ) {
+				$json_data = file_get_contents( $json_path );
```
bfintal, line 60: Use `wp_json_file_decode` instead

src/plugins/global-settings/preset-controls/editor-loader.js
```diff
@@ -0,0 +1,57 @@
...
+		const _customPresetControls = select( 'stackable/global-preset-controls.custom' )?.getCustomPresetControls()
+		return { customPresets: { ..._customPresetControls } ?? [] }
+	}, [] )
+	const [ styles, setStyles ] = useState( '' )
+
+	useEffect( () => {
+		if ( customPresets && typeof customPresets === 'object' ) {
+			renderGlobalStyles( customPresets, setStyles )
+		}
+	}, [ JSON.stringify( customPresets ) ] )
```
bfintal, line 54: This might work without needing to do `JSON.stringify()`
```suggestion
	}, [ customPresets ] )
```

src/plugins/global-settings/preset-controls/index.php
```diff
@@ -0,0 +1,172 @@
...
+		 * Initialize
+		 */
+  		function __construct() {
+			$this->custom_presets = get_option( 'stackable_global_custom_preset_controls' );
+			$this->theme_presets = WP_Theme_JSON_Resolver::get_theme_data()->get_settings();
+			$this->default_presets = WP_Theme_JSON_Resolver::get_core_data()->get_settings();
+			$this->stackable_presets = $this->load_presets( __DIR__ . '/presets.json');
+
+			add_filter( 'stackable_inline_styles_nodep', array( $this, 'add_preset_controls_styles' ) );
+			add_filter( 'stackable_inline_editor_styles', array( $this, 'add_preset_controls_styles' ) );
```
bfintal, line 51: These will be called even in the front end where I don't think the presets are needed.

I recommend to place the code the populates `custom_presets`, `theme_presets` etc into a new function:

```php
public function load_presets() {
    $this->custom_presets = get_option( 'stackable_global_custom_preset_controls' );
    // ...
}
```

Then calling `$this->load_presets()` when the values are about to be used.

This is so that the loading process will only be called when needed (currently this performed for every page load)

src/plugins/global-settings/preset-controls/index.php
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		public static function sanitize_array_setting( $input ) {
+			return ! is_array( $input ) ? array( array() ) : $input;
+		}
+
+		private function load_presets( $json_path ) {
```
bfintal, line 58: Maybe change this function name since it's not actually loading presets, it's just loading the provided json file

src/plugins/global-settings/preset-controls/index.php
```diff
@@ -0,0 +1,172 @@
...
+			// If using theme presets, use WP generated --wp-preset to support theme.json specific
+			// configuration (fluid, clamping, etc.)
+			foreach ( $presets_by_slug as $slug => $preset ) {
+				$is_custom = $preset[ '__is_custom' ] ?? false;
+		
+				$value = $is_custom || ! $isTheme
+					? $preset['size']
+					: "var(--wp--preset--$prefix--$slug)";
+		
+				$css .= "--stk--preset--$prefix--$slug: $value;\n";
```
bfintal, line 108: Instead of building the css from strings, you can return an object here instead, then later on use `wp_style_engine_get_stylesheet_from_css_rules` to build the css so we get the benefit of sanitation.

src/plugins/global-settings/preset-controls/index.php
```diff
@@ -0,0 +1,172 @@
...
+		/**
+		 * Get the value from an array with an array of keys
+		 *
+		 * @param array $array 
+		 * @param array $keys 
+		 * @return mixed
+		 */
+		public function deepGet( $array, $keys ) {
+			return array_reduce( $keys, fn( $value, $key ) => $value[ $key ] ?? null, $array );
+		}
```
bfintal, line 123: If you use `wp_style_engine_get_stylesheet_from_css_rules` from the previous comment, you might not need to use this anymore

src/plugins/global-settings/typography/index.js
```diff
@@ -75,11 +95,62 @@ const TYPOGRAPHY_TAGS = [
...
+		label: __( 'None', i18n ),
+		value: 'none',
+	},
+	{
+		label: __( 'Custom', i18n ),
+		value: 'custom',
+		disabled: true,
+	},
+	{
+		label: __( '1.067 - Minor Second', i18n ),
```
bfintal, line 109: Make the English portion only as translatable:

```suggestion
		// Translators: First %s is a unit, second %s is a name of the unit
		label: sprintf( __( '%s - %s', i18n ), '1.067', __( 'Minor Second', i18n ) ),
```

src/plugins/global-settings/typography/index.js
```diff
@@ -88,23 +159,90 @@ addFilter( 'stackable.global-settings.inspector', 'stackable/global-typography',
...
+
+			const newSettings = { ...allCustomPresets, fontSizes: fontSizePresets }
+
+			clearTimeout( saveTypographyAsPresetsThrottle )
+			saveTypographyAsPresetsThrottle = setTimeout( () => {
+				const settings = new models.Settings( { stackable_global_custom_preset_controls: newSettings } ) // eslint-disable-line camelcase
+				settings.save()
+			}, 300 )
+
+			dispatch( 'stackable/global-preset-controls.custom' ).updateCustomPresetControls( newSettings )
```
bfintal, line 243: Looks like this code updates the preset controls and belongs to the preset controls area. In the preset controls area, you can just listen to calls to `stackable.global-settings.typography.update-trigger` and perform preset changes accordingly

bfintal (changes requested):

PR #3490, opened 2025-04-21
Feat: theme.json block style inheritance

kaeizen:


src/welcome/admin.js
```diff
@@ -1418,6 +1419,15 @@ const AdditionalOptions = props => {
 									handleSettingsChange( { stackable_global_colors_native_compatibility: checked } ) // eslint-disable-line camelcase
 								} }
 							/>
+							<CheckboxControl
+								label={ __( 'Inherit Block Styles from theme.json', i18n ) }
+								className={ searchClassname( __( 'Inherit Block Styles from theme.json', i18n ), miscellaneous ) }
+								help={ __( `When enabled, Stackable blocks will inherit styles from the theme.json file.`, i18n ) }
```
bfintal, line 1425:
```suggestion
								help={ __( `When enabled, Stackable blocks will inherit styles from the theme.json file of your Block Theme.`, i18n ) }
```

bfintal (changes requested):
1. When I activated this version / PR from another version, the "Inherit Block Styles from theme.json" option was checked by default (my plugin version was `3.15.3`). The expected behavior should be when you update, the option should be **disabled** by default. New installs will have the option **enabled** by default.

2. In 2025 theme (default styles), the button border radius is not inherited. I checked other themes and they look okay:

image

3. In the Block Theme Tove, the button styles are properly inherited, but in the editor it's missing the outline (works in the frontend), can you investigate this?

image

kaeizen:
> 2. In 2025 theme (default styles), the button border radius is not inherited. I checked other themes and they look okay:
> 
> image

The button border radius is not defined in the Twenty Twenty-Five `theme.json`. Instead, it is declared in `wp-includes/blocks/button/style-rtl.css`, which is why it is not inherited—since we're only inheriting styles from `theme.json`.

PR #3414, opened 2025-02-03
Fix: String Translation in admin.js

DAnn2012:
Fix: String Translation in admin.js

Fix #3431

PR #3501, opened 2025-05-28
Feat: Design Library Revamp

kaeizen:


src/welcome/admin.js
DAnn2012, line 1: Hello @mxkae 

Could you include in this PR of yours these changes of mine to this file (#3414 ) from about 5 months ago?

Thanks.

PR #3528, opened 2025-06-13
fix: insert the global settings style at the start instead of appendi…

### LINKED ISSUE #3520: When you click on the top portion of the editor, it scrolls down
If you click on the top portion of the editor (outside the title and above the first block). the editor scrolls down to the bottom.

https://github.com/user-attachments/assets/8a439df0-d10a-4e0e-8fca-b464ca85d9bb

### LINKED ISSUE #3529: Page auto scrolls to bottom when scrolling up and down
The page auto-scrolls to the bottom when trying to scroll up or down and clicking the scrollbar area. Scrolling unexpectedly adds Paragraph blocks to the page. 

https://github.com/user-attachments/assets/fd6b5b83-4e5d-49d8-a25d-faf63a8b5e54

Arukuen:
…ng after the main editor div to avoid unexpected behavior

fixes #3520
fixes #3529

### DIFF
--- a/src/plugins/global-settings/editor-loader.js
+++ b/src/plugins/global-settings/editor-loader.js
@@ -45,16 +45,15 @@ const GlobalSettingsLoader = () => {
 	 */
 	useEffect( () => {
 		const editorBody = editorDom?.closest( 'body' )
-		if ( editorBody ) {
-			// No need to check for firstChild, since there will always be at least one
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
As an additional improvement, how about we add these style tags in the editor's head tag?

PR #3534, opened 2025-06-23
Fix: Add default colors for container color scheme fallback

kaeizen:


bfintal (changes requested):
I'm concerned that with this change there will be people who used color schemes this early will have their websites broken.

Can you add a new option in Stackable > Miscellaneous for "Use v3.16.0 color scheme inheritance". Description: "A bug in color scheme inheritance was present in v3.16.0-v3.16.1. This has been fixed, but in order to preserve the state of affected sites, you may find this option turned on." (feel free to improve this)
- If the option is **on** use the old/broken color scheme inheritance.
- If the option is **off** (default) use the fixed color scheme inheritance.

Upgrading path:
- If the user has at least one color scheme save at the time of updating, make the option **on** (this handles people who are updating from a version before color schemes or haven't used color schemes yet.)
- Otherwise, **off**
- Fresh installs will have this option **off**

Make sure the code that makes the color scheme inheritance "broken" again is split in a deprecated file.

PR #3473, opened 2025-03-28
Fix: Two settings strings are not displayed as translated

DAnn2012:
Hello @bfintal 

Fix: Two settings strings are not displayed as translated

Update index.js

Fix #3472

PR #3321, opened 2024-09-23
Fix String Translation in index.js

DAnn2012:
Hello @bfintal 

Fix String Translation in index.js

Added the text domain to this string because it is no longer present among the [default ones in WordPress 6.6.x](https://translate.wordpress.org/consistency/?search=Select+a+variation+to+start+with.&set=it%2Fdefault&project=).

Fix #3433

PR #3541, opened 2025-07-04
Feat: global block styles

kaeizen:


src/plugins/global-settings/typography/index.js
DAnn2012, line 1: Hello @mxkae 

Could you include in this PR of yours these changes of mine to this file (#3473 ) from over 3 months ago?

Thanks.

DAnn2012:
Hello @mxkae 

Could you please include in this PR these changes I made to file `src/components/variation-picker/index.js` (#3321 ) about 9 months ago?

Thanks.

bfintal (changes requested):
Bug encountered:
- Add Card 3 design from the design library, then pick the first inner column, then create a new block style. The block would disappear and then the other columns would grow taller:

https://github.com/user-attachments/assets/22046e6a-9961-4512-894f-f8b5f9d94c6e


Changes:
1. For the free, simplify the things present in the UI:

What we have now:
Screenshot 2025-07-15 at 6 41 36 PM

Change to:
Screenshot 2025-07-15 at 6 42 05 PM
- Remove "(Modified)"
- Remove reset button

2. I added a button (using blocksy with some blocksy styles), then updated to the PR. My style says "Default (Modified)" with a reset button even though nothing was modified in the block (clicking on the reset button does nothing):

image

3. Change it so that if it's in "Default" block style, it shouldn't ever say "(Modified)" and there shouldn't be any reset button.
4. When clicking the plus button, the name field should be auto-focused so you can type right away.
5. Match the normal style of the normal select drop down. Right now it looks different from the normal drop down. Any reason for this?

What we have:
image

What we normally use:
image

6. Change the alignment of the elements like this. If no reset button, occupy the entire width:
image

if there is a reset button, just add margin right to the select drop down:
image

7. The update button feels weird but I can't put my finger on it. it's like it's too wide or has too much gaps

bfintal (changes requested):
A few more changes:
- When you open the block style popover by clicking on the button, you should be able to press the up and down arrows to switch between the different block styles (focus moves on the buttons), then when you press enter, the block style selected will be selected.
- Add a focus on the "save" and "update" buttons when tabbing through them.
- Clicking on the default block style when the default block style is currently selected should not apply it. Currently, if you add the design "Call to action 1" from the design library, then select the heading block, you can click on the "default" block style and the text would turn black. Expected behavior is it will not do anything.
- When you open the block styles popover, the save style takes a few moments to show up, it should not do this and should just be there right away

https://github.com/user-attachments/assets/415f252a-e324-45a3-9327-562cf3efdc30

src/components/block-css/use-block-style-generator.js
```diff
@@ -30,6 +30,7 @@ export const useBlockCssGenerator = props => {
 			oldText.current = attributes.text
 			return oldCss.current
 		}
+
```
bfintal:
```suggestion
```

PR #3544, opened 2025-07-07
fix: merge but ensure unique slug

### LINKED ISSUE #3543: Presets are added twice in Blocksy theme
**Describe the bug**
The presets in the Inspector are duplicated. This occurs in inner column spacing, row gap, column gap, paddings, margins, inner block gap.

**To Reproduce**
1. Use Blocksy theme
2. Add a Page
3. Add a Column Block
4. In the Column Block, change the preset value of Inner Spacing
5. See error (the presets are duplicated)

**Expected behavior**
The presets should display each size option only once.

**Screenshots**

https://github.com/user-attachments/assets/7236800a-0384-46cb-9ee1-ed01d498ee05

Arukuen:
fixes #3543


bfintal (changes requested):
Please add some error handling just in case the presets do not have slugs and if other variables are not sure of the type

PR #3570, opened 2025-07-31
Optimizations: Generate CSS Files, disable deprecated code, other speed stuff

bfintal:
# This Pull Request does the following optimizations

- [x] Global design system is now a generated CSS file
- [x] Remove need to use hook `typography_detect_native_blocks` if not using global typography
- [x] Do not register Design library as patterns anymore since these produce overhead
- [x] You can define `STACKABLE_DISABLE_DEPRECATED_CODE` to disable all backward compatibility code in-lieu for speed
- [x] Added `did_action` to `load_frontend_scripts_conditionally` to lessen action calls
- [x] Moved filter `safecss_filter_attr_allow_css` only during style generation CC @mxkae 
- [x] Moved `stackable_block_style_inheritance_inline_styles_nodep` generation of block style inheritance only when needed
- [x] Added CSS file generation for block theme style inheritance 
- [x] Add setting for `stackable_use_css_files` defaults to checked (`yes`)
- [x] Add button to invalidate generated CSS files

# Remaining work to be done

- [ ] ~Change `load_frontend_scripts_conditionally`, remove old checker, and instead of hooking to `render_block`, hook to all `render_block_{name}` and then remove hook when completed each.~ this was slow!
- [ ] ~See if `wp_get_global_styles` can be used instead of `WP_Theme_JSON_Resolver` as recommended by WP docs CC @mxkae~

## Summary by CodeRabbit

- New Features
  - Added an option to generate and cache CSS files for improved performance, with automatic inline fallback.
  - Introduced an admin toggle to enable/disable CSS file generation and a button to invalidate cached CSS.

- Performance
  - Optimized frontend and editor asset loading, including footer-loaded scripts and smarter style enqueueing.
  - Improved global typography handling and block style inheritance ordering.

- Bug Fixes
  - More reliable behavior when custom breakpoints are not set.
  - Prevents deprecated components from loading when explicitly disabled.


src/design-library/init.php
```diff
@@ -30,6 +30,10 @@ public function __construct() {
 			add_action( 'rest_api_init', array( $this, 'register_route' ) );
 
 			add_action( 'stackable_delete_design_library_cache', array( $this, 'delete_cache_v3' ) );
+
+			// DEV NOTE: Commenting this out for now since it is producing server load in the frontend.
+			// We can re-enable this when we have a better way to register block patterns.
+			// add_action( 'init', array( $this, 'register_design_pattern' ) );
```
bfintal:
```suggestion
```

malalacharm:
Block errors in editor:
image

PR #3587, opened 2025-08-27
fix (design library): remove from inserter

### LINKED ISSUE #3581: Remove design library block
Remove the Design Library Block from the block inserter

kaeizen:
fixes #3581 

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve existing block visibility settings when applying metadata, avoiding accidental overrides of inserter visibility.
* **Chores**
  * Hide the Design Library block from the inserter by default so it cannot be added directly.

### DIFF
--- a/src/block/design-library/index.js
+++ b/src/block/design-library/index.js
@@ -13,8 +13,6 @@ import edit from './edit'
 import save from './save'
 import metadata from './block.json'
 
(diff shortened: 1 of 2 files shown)


bfintal (changes requested):
We can just set inserter to false from the json

PR #3591, opened 2025-09-15
Fix (design library): render designs if within view

kaeizen:
## Summary by CodeRabbit

- Refactor
  - Improved list rendering performance with debounced visibility checks to reduce scrolling jank.
  - Tuned lazy-load thresholds so items begin loading earlier/later around the viewport for smoother appearance.
  - Maintained consistent card heights while offscreen to prevent layout shifts.
  - Increased stability of preview/background updates by refining update triggers.
  - Removed non-user-facing attributes from placeholders and buttons for cleaner DOM.


bfintal (changes requested):
Please do the change request from coderabbit

PR #3593, opened 2025-09-16
Fix: accessibility of blocks

kaeizen:
## Summary by CodeRabbit

- New Features
  - Accessibility improvements across multiple blocks: carousel and horizontal scroller support keyboard focus; notification close button has a translatable aria-label; progress bar and progress circle include improved ARIA labeling; posts block meta uses a more consistent semantic container.
- Chores
  - Version bumped to 3.19.0.

### DIFF
--- a/src/block/horizontal-scroller/save.js
+++ b/src/block/horizontal-scroller/save.js
@@ -55,8 +55,10 @@ export const Save = props => {
 			 { attributes.generatedCss && <style>{ attributes.generatedCss }</style> }
 			 <CustomCSS.Content attributes={ attributes } />
 			<div
-				className={ contentClassNames }
-				{ ...applyFilters( 'stackable.horizontal-scroller.save.scroller', props.version ) }
+				{ ...applyFilters( 'stackable.horizontal-scroller.save.scroller-props', {
+					className: contentClassNames,
+					tabIndex: 0,
+				}, props.version ) }
 			>
 				<InnerBlocks.Content />
 			</div>
--- a/src/block/horizontal-scroller/save.js
+++ b/src/block/horizontal-scroller/save.js
@@ -56,9 +56,8 @@ export const Save = props => {
 			 <CustomCSS.Content attributes={ attributes } />
 			<div
 				className={ contentClassNames }
-				{ ...applyFilters( 'stackable.horizontal-scroller.save.scroller-props', {
-					tabIndex: 0,
-				}, props.version ) }
+				tabIndex={ 0 }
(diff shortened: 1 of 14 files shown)


src/block/progress-circle/save.js
```diff
@@ -71,6 +72,7 @@ export const Save = props => {
 					aria-valuemax="100"
 					aria-valuenow={ progressValue }
 					aria-valuetext={ attributes.progressAriaValueText ? striptags( attributes.progressAriaValueText ) : undefined }
+					{ ...applyFilters( 'stackable.progress-circle.accessibility', attributes.progressAriaValueText, props.version ) }
```
bfintal: Instead of the filter adding the `aria-label`, the `aria-label` should be added normally, and then the filter would remove it for deprecation. this would also need to be done in the save portion

```suggestion
					aria-label={ attributes.progressAriaValueText ? striptags( attributes.progressAriaValueText ) : undefined }
					{ ...applyFilters( 'stackable.progress-circle.div-props', {}, props.version ) }
```

Then in the deprecation
```
const addAriaLabel = ( props, version ) => {
	// If from an old version, remove the aria-label
	if ( oldVersion ) {
		return { 
			...props,
			'aria-label': undefined,
		}
	}
	return props
}
addFilter( 'stackable.progress-circle.div-props', 'stackable/3.19.0', props, removeAriaLabel )
```

src/block/notification/save.js
```diff
@@ -78,6 +78,7 @@ export const Save = props => {
 				{ attributes.isDismissible &&
 					<button
 						className="stk-block-notification__close-button"
+						{ ...applyFilters( 'stackable.notification.save.close-button', props.version ) }
```
bfintal: Check my suggestion for the progress circle, this prop would need to be added in by default, and then the deprecation filter would remove it. This is so that our save.js always contains our latest version, and the filters are there to modify it for older versions

src/block/horizontal-scroller/save.js
```diff
@@ -54,7 +54,10 @@ export const Save = props => {
 		 >
 			 { attributes.generatedCss && <style>{ attributes.generatedCss }</style> }
 			 <CustomCSS.Content attributes={ attributes } />
-			<div className={ contentClassNames }>
+			<div
```
bfintal, line 57: Check my suggestion for the progress circle, this prop would need to be added in by default, and then the deprecation filter would remove it. This is so that our save.js always contains our latest version, and the filters are there to modify it for older versions

bfintal (changes requested):

PR #3621, opened 2025-10-09
fix: resetting on global typography should only affect the current de…

### LINKED ISSUE #3550: Global mobile typography setting displays reset button on desktop
**Describe the bug**
When a global typography style is configured specifically for mobile viewports, a reset button appears in the desktop view. When you click reset in desktop, it also resets the mobile values.

**To Reproduce**
1. Go to Stackable Settings > Global Typography Typescale
2. Change into Mobile view and select 1.25 Major Third typescale
3. Switch into desktop view
4. See error (reset button on the 

**Expected behavior**
When a global typography style is configured specifically for mobile view, the styling should apply only to mobile view. When the reset button is clicked within the desktop view, it should only reset the typography settings specific to the desktop view.

**Screenshots**

https://github.com/user-attachments/assets/f9446246-82fd-406f-9326-d78d0e992e9e

Arukuen:
…vice type

fixes #3550 

## Summary by CodeRabbit

* Improvements
  * Typography resets are now device-aware, preserving selector-specific styles while clearing size and spacing only for the current device.
  * Default font family is applied based on the active font pair during reset for more consistent results.
  * Detection of when a reset is needed now respects per-device values, reducing unnecessary changes.

* Bug Fixes
  * Resolves inconsistencies where typography resets could unintentionally affect other devices or overlook device-specific settings.

### DIFF
--- a/src/plugins/global-settings/typography/index.js
+++ b/src/plugins/global-settings/typography/index.js
@@ -350,10 +350,29 @@ addFilter( 'stackable.global-settings.inspector', 'stackable/global-typography',
 
 	const resetStyles = selector => {
 		let newSettings = {}
-		const currentFontPair = getCurrentFontPair()
-		if ( ! isEditingFontPair && currentFontPair ) {
-			newSettings = { ...typographySettings, [ selector ]: currentFontPair.typography[ selector ] }
+
+		const deviceFontSizeKey = getDevicePropertyKey( 'fontSize', deviceType )
+		const deviceFontSizeUnitKey = getDevicePropertyKey( 'fontSizeUnit', deviceType )
+		const deviceLineHeightKey = getDevicePropertyKey( 'lineHeight', deviceType )
+		const deviceLetterSpacingKey = getDevicePropertyKey( 'letterSpacing', deviceType )
+
+		newSettings = {
+			...typographySettings,
+			[ selector ]: {
+				...typographySettings[ selector ],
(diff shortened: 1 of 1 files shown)


src/plugins/global-settings/typography/index.js
```diff
@@ -350,11 +350,30 @@ addFilter( 'stackable.global-settings.inspector', 'stackable/global-typography',
...
+		const deviceFontSizeKey = getDevicePropertyKey( 'fontSize', deviceType )
+		const deviceFontSizeUnitKey = getDevicePropertyKey( 'fontSizeUnit', deviceType )
+		const deviceLineHeightKey = getDevicePropertyKey( 'lineHeight', deviceType )
+		const deviceLetterSpacingKey = getDevicePropertyKey( 'letterSpacing', deviceType )
+
+		newSettings = {
+			...typographySettings,
+			[ selector ]: {
+				...typographySettings[ selector ],
+				fontFamily: getDefaultFontFamily( selector ),
```
bfintal, line 363: Since this is a reset, shouldn't the fontFamily be blank?

bfintal (changes requested):

src/plugins/global-settings/typography/index.js
Arukuen: The default font family depends on the currently selected font pair, which will be provided by `getDefaultFontFamily`.

PR #3624, opened 2025-10-09
Fix (color schemes): change selectors for color schemes

kaeizen:
## Summary by CodeRabbit

- New Features
  - Better Blocksy theme defaults for container heading/link colors.
  - Hover effects now respect container/background schemes in editor and frontend.
  - Frontend/editor logic optimized to handle empty color-scheme sets more efficiently.

- Bug Fixes
  - More consistent application of color schemes across blocks.
  - Editor body-class synchronization improved to keep preview and frontend aligned.

- Refactor
  - Unified color-scheme targeting to a configurable selector and grouped scheme-aware hover/background rules.

### DIFF
--- a/src/compatibility/blocksy/style.scss
+++ b/src/compatibility/blocksy/style.scss
@@ -1,0 +1,6 @@
+// Use Blocksy theme colors as fallback defaults for container color schemes.
+// These defaults apply when the default container color scheme is unset, but background or custom schemes exist.
+:where(.stk--is-blocksy-theme.stk--has-container-scheme) {
+	--stk-default-heading-color: var(--theme-heading-color, var(--theme-heading-2-color, var(--theme-headings-color)));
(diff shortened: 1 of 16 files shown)


src/compatibility/blocksy/style.scss
bfintal, line 1: Do we still need this since the color scheme selectors won't anymore add styles when color schemes are not used?

kaeizen: This fix is necessary because the `--stk-default-heading-color` and `--stk-default-link-color` variables become undefined when using Blocksy.

Image

The variables above define default colors for containers when background or custom color schemes exist, but no container (or base) scheme is set. They prevent containers from unintentionally inheriting colors from the background scheme, as shown in the design library.

Image

Here’s how a heading block inside a container looks without these styles:

Image

And here’s the expected appearance once they're added:

Image

PR #3648, opened 2025-11-26
fix (icon list): don't override custom icon colors

### LINKED ISSUE #3512: Icon List Block - Multi-colored SVGs appear as solid black due to CSS fill override
**Describe the bug**
Multi-colored SVG icons (such as Noto Color Emoji) appear as solid black when used in the Icon List block. The `fill ` rule overrides the internal colors of the SVG and forcing a single-color fill.

**To Reproduce**
1. Add a page
2. Add an Icon List Block
3. Change the icon into any Noto Color Emoji
4. See error (icon turns into black)

**Expected behavior**
Multi-colored SVG icons should retain their original colors when displayed in the Icon List block

**Screenshots**

https://github.com/user-attachments/assets/7822be6e-4201-4235-a5b5-5b73a33b01ef

kaeizen:
fixes #3512 

## Summary by CodeRabbit

* **New Features**
  * Added a "Use Custom Icon Color" toggle in the icon list block settings to let editors opt icons out of the block's default color styling.

* **Style**
  * Improved icon styling so custom-colored icons retain their chosen colors while standard icons continue to receive the block's color treatment.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/block/icon-list-item/style.scss
+++ b/src/block/icon-list-item/style.scss
@@ -16,12 +16,6 @@
 		border-width: 1px;
 		border-color: rgba(0, 0, 0, 0.4);
 	}
-
-	// Apply the color to the SVG path on the defintion
-	svg.ugb-custom-icon :is(g,path,rect,polygon,ellipse) {
-		fill: var(--stk-icon-list-marker-color) !important;
-		color: var(--stk-icon-list-marker-color) !important;
-	}
 }
 
 .stk-block-icon-list-item__content {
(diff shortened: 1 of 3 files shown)


bfintal (changes requested):
This change will affect existing sites and will remove the currently functionality of allowing custom icons to be colorized.

Instead of doing this change for all, maybe we can add an option in the color settings to retain the original colors of custom uploaded icons

PR #3649, opened 2025-11-26
fix (posts block): fix image width in horizontal block styles

### LINKED ISSUE #3554: Posts Block - formatting breaks on Horizontal and Horizontal 2 Layout
**Describe the bug**
The Text content in the Posts block becomes compressed vertically inside the column and the image is stretched (width is not applied)

**To Reproduce**
1. Add a page
2. Add a two-column Columns Block
3. On the other column, add a Posts block
4. Change image width into 300px
5. See error (formatting breaks)

**Expected behavior**
The Horizontal and Horizontal 2 layout should display the posts properly

**Screenshots**

Image

Image

kaeizen:
fixes #3554 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image sizing in horizontal and list post layouts: linked images that use percentage-based widths now preserve their intended display and sizing, avoiding incorrect width reduction and delivering more consistent visuals across layout modes.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/block/posts/style.js
+++ b/src/block/posts/style.js
@@ -281,26 +281,13 @@ Image.addStyles( blockStyles, {
 		const blockStyle = getBlockStyle( variations, className )
 		return ! [ 'list', 'horizontal', 'horizontal-2', 'portfolio', 'portfolio-2', 'vertical-card-2' ].includes( blockStyle?.name )
 	},
-	saveEnableWidthCallback: getAttribute => {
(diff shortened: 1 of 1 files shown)


bfintal (changes requested):
Looks like the original behavior is we disallowed the width option for horizontal layouts. Not sure if this is the right fix for this because we disabled it for a reason.

PR #3684, opened 2026-02-19
fix: fix lazy-loaded images in safari

Arukuen:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed rendering issues for lazy‑loaded images in Safari on iOS/macOS to reduce flicker and improve visual stability.
  * Improved perceived performance for images that load lazily in affected Safari versions.

### DIFF
--- a/src/block-components/image/style.scss
+++ b/src/block-components/image/style.scss
@@ -71,7 +71,7 @@
 	}
 }
 
-// This is needed to fix the images rendering as lines when lazy-loaded in Safari.
+// This is needed to fix the images rendering as lines when lazy-loaded in Safari. We target only the known issue here based on lazy loading from another plugin so as not to affect other uses of the image
 // Targets Safari 9+ on iOS and macOS, works as of late 2025.
 @supports (font: -apple-system-body) {
 	img.stk-img.lazyloaded {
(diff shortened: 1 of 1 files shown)


src/block-components/image/style.scss
```diff
@@ -70,3 +70,11 @@
 		border-radius: 9999px !important;
 	}
 }
+
+// This is needed to fix the images rendering as lines when lazy-loaded in Safari.
+// Targets Safari 9+ on iOS and macOS, works as of late 2025.
+@supports (font: -apple-system-body) {
+	img.stk-img {
```
bfintal:
```suggestion
	img.stk-img.lazyloaded {
```

src/block-components/image/style.scss
```diff
@@ -70,3 +70,11 @@
 		border-radius: 9999px !important;
 	}
 }
+
+// This is needed to fix the images rendering as lines when lazy-loaded in Safari.
```
bfintal:
```suggestion
// This is needed to fix the images rendering as lines when lazy-loaded in Safari. We target only the known issue here based on lazy loading from another plugin so as not to affect other uses of the image
```

PR #3655, opened 2025-12-03
Fixed WordPress minimum required version number to 6.8.2

DAnn2012:
Hello @bfintal

Fixed WordPress minimum required version number to 6.8.2

Thanks.

## Summary by CodeRabbit

* **Documentation**
  * Updated installation “Minimum Requirements” to require WordPress **6.8.2+** (was **6.5.5+**).
  * Refreshed the **3.19.7** changelog text to correct spelling in “Posts block” bullet items (keeping existing issue references).

PR #3637, opened 2025-11-02
Fixed Text Domain in attributes.php

DAnn2012:
Fixed Text Domain in attributes.php

## Release Notes

This update contains internal maintenance improvements with no visible changes to end-user functionality. The blog posts component has been updated to use standardized configuration for localization handling, ensuring consistent translation practices across the application.

* **Chores**
  * Updated internal localization configuration for blog posts component

PR #3636, opened 2025-11-02
Fix: Update translation function in map example block

DAnn2012:
Fixed Function in example.js

## Summary by CodeRabbit

* **Chores**
  * Updated translation handling in the Map block example for improved consistency.

PR #3690, opened 2026-03-11
fix: prevent shifting in safari

### LINKED ISSUE #3687: Icon Block - icon shifts position on hover in Safari when gradient background is applied to both normal and hover states
**Describe the bug**
When a gradient background is applied to an icon in both the normal and hover states, the icon shifts or “moves” slightly when hovered in Safari.

**To Reproduce**
1. Add a Button block
2. Add an Icon within the button block
3. For the Button color, add a gradient color for Normal State and Hovered State
4. See error in Safari (icon is shifting)

**Expected behavior**
When a gradient background is applied to both the normal and hover states of a button, the icon should stay in place on hover.

**Screenshots**

https://github.com/user-attachments/assets/8dd5830f-16f9-47b2-aee1-7e070e1caea2

Arukuen:
fixes #3687 

## Summary by CodeRabbit

* **Style**
  * Optimized button rendering to reduce visual layout shifts during interactions, with enhanced stability for Safari browsers.

### DIFF
--- a/src/block/button/style.scss
+++ b/src/block/button/style.scss
@@ -28,3 +28,12 @@
 		box-sizing: border-box;
 	}
 }
+
+// Force Safari to render the SVG on its own layer to prevent shifting
+// during parent transition.
+// @see https://github.com/gambitph/Stackable/issues/3687
+@supports (font: -apple-system-body) {
+	.stk-button svg {
+		will-change: transform;
+	}
+}


bfintal (changes requested):
Since the issue only lies when the button has a **gradient background**, this fix should only apply during that scenario, you can use a class check to only apply the fix to buttons with gradient backgrounds

Arukuen:
@bfintal 
The issue always happens as long as there is an icon, regardless of whether the button has a gradient background or not.

https://github.com/user-attachments/assets/4f39c0f2-9367-4351-bd17-861d6d6c38d3

DAnn2012:
Hello @bfintal 

I'd like to ask you, even if it's off-topic, to also take a look at this my PRs #3636 #3637 #3655 

Thanks.

PR #3691, opened 2026-03-19
fix: transform on <g> instead of <svg> as a safari fix

### LINKED ISSUE #3650: Image Block - vertical/horizontal flip shape not working properly on mobile view
**Describe the bug**
The Image Block Vertical Flip and Horizontal Shape Flip options are not behaving correctly on mobile devices. The image does not flip at all compared to desktop to view.

**To Reproduce**
1. Add a page
2. Add Image block and add image
3. Select Image shape Blob 1
4. Enable Flip Shape Horizontally/Vertically
5. See error (on desktop view its correct, on mobile view its incorrect)

**Expected behavior**
When applying Vertical Flip or Horizontal Shape Flip to the Image Block, the image should consistently display the flipped orientation across all devices. 

**Screenshots**
Desktop view:
Image

Mobile view:

Image

Arukuen:
fixes #3650

Safari doesn't support transform on the root <svg> element when used as a CSS mask. The fix is to wrap the content in a <g> and apply the transform there instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed flipped image-shape rendering so horizontal and vertical flips now display consistently across contexts by embedding the flip into the image data. This resolves cases where flipped shapes appeared incorrect or misaligned, preserving intended appearance and layout.

### DIFF
--- a/src/block-components/image/get-shape-css.js
+++ b/src/block-components/image/get-shape-css.js
@@ -30,9 +30,32 @@ export const getShapeCSS = ( shape, shapeFlipX, shapeFlipY, shapeStretch ) => {
 	const MaskImage = getShapeSVG( shape )
 	const MaskComp = <MaskImage
 		preserveAspectRatio={ ! [ '', 'square' ].includes( shape ) && shapeStretch ? 'none' : undefined }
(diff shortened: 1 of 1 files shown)


bfintal:
@malalacharm will have to check whether this PR will produce errors in existing blocks that 1) do not encounter the problem, and 2) encounter the problem

PR #3745, opened 2026-08-18
fix: add custom attributes for generated links

### LINKED ISSUE #3743: Image Block - add an accessibility label input to the link settings
**Describe the bug**
The Image/Image Box block fails accessibility audits when a link is applied to the block.

Image Box block: Has an aria-label, but it is incorrectly placed on the outer `<div>`
Image block: The linked `<a>` has no aria-label at all

**To Reproduce**
1. Add an Image Block
2. Add a Link to the block
3. Check on frontend
4. See error in Lighthouse ""Links must have discernible text"

**Expected behavior**
When a value is provided, the accessibility label should be applied directly to the generated `<a>` element, ensuring that the link has a discernible and accessible name.

For the Image Box block, the label should no longer be applied only to the outer `<div>` container. For the Image block, the label should be added to the linked `<a>` element.

Arukuen:
fixes #3743

## Summary
- Replaces the dedicated link accessibility label field with a second Custom Attributes input that writes to the generated `<a>`, not the outer block wrapper.
- Users can add `aria-label="Learn more"` (or any other HTML attribute) under **Advanced > Custom Attributes > Link Custom Attributes**.
- Image and Icon apply `linkCustomAttributes` through `Link.Content`. Image Box and other block-link blocks apply `blockLinkCustomAttributes` through `BlockLink.Content`.

Existing block Custom Attributes stay on the wrapper. Move `aria-label` into the new link field when the label needs to sit on the anchor for accessibility audits.

## Test plan
- [ ] Image: add a link, set `aria-label="..."` in Link Custom Attributes, confirm it is on the image `<a>` on the frontend (not the wrapper `div`).
- [ ] Image Box: add a block link, set `aria-label="..."` in Link Custom Attributes, confirm it is on the overlay `<a>` (not the outer `div`).
- [ ] Block Custom Attributes still apply only to the block wrapper.
- [ ] Other block-link blocks (Card, Column, Call to Action, etc.) get the same link field and output.
- [ ] Blocks without a generated link still show a single Custom Attributes field.

### DIFF
--- a/src/block-components/block-link/index.js
+++ b/src/block-components/block-link/index.js
@@ -29,6 +29,7 @@ BlockLink.Content = props => {
 			target={ attributes.blockLinkNewTab ? '_blank' : '' }
 			rel={ attributes.blockLinkRel || undefined }
 			title={ attributes.blockLinkTitle || undefined }
+			aria-label={ attributes.blockLinkAriaLabel || undefined }
 			aria-hidden={ isHidden ? 'true' : undefined }
 			tabindex={ isHidden ? '-1' : undefined }
 		/>
--- a/src/block-components/helpers/link/attributes.js
+++ b/src/block-components/helpers/link/attributes.js
@@ -50,4 +50,19 @@ export const addLinkAttributes = ( attrObject, attrNameTemplate = '%s', selector
 		versionAdded: '3.0.0',
 		versionDeprecated: '',
 	} )
+
+	attrObject.add( {
+		attributes: {
+			ariaLabel: {
+				type: 'string',
+				source: 'attribute',
+				selector,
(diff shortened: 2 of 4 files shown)


bfintal:
@Arukuen doesn't custom attributes already handle this? can't we specify `aria-label="value"` already in there?

Arukuen:
@bfintal

Custom Attributes can accept aria-label="value", but applies it to the block’s outer BlockDiv, not to a nested link. This PR adds an option to apply the label directly to the link, which is why it appears under Advanced > Link.

bfintal:
@Arukuen I think better if we add another custom attributes field (can be below the existing one), but this time this will be for the actual link html

